PR #5, opened 2014-06-12
Add preset for Libav

lu-zero:


saudet:
I am not seeing an advantage in maintaining additional presets for Libav in addition to FFmpeg, so I will drop this pull request, but if you would still like to spend some time on this, please let me know! Thanks

PR #41, opened 2015-03-18
make cmake configurable

javelinjs:
There may exists several different version of cmake. For example, RHEL 6 contains both cmake and cmake28 in its yum repo. This PR allows user to specify the exact version, using 'CMAKE=cmake28 bash cppbuild.sh ...'


saudet:
Thanks!!

Issue #56, opened 2015-06-15
compiling opencv gives error


jduke32:
hi , i when i tried to compile opencv from javacpp , i got error as shown below:
```
C:\Users\Kadir\Desktop\javacpp-presets\opencv\target\classes\org\bytedeco\javacp
p\jniopencv_objdetect.cpp(8865) : warning C4244: 'initializing' : conversion fro
m 'jlong' to 'jint', possible loss of data
C:\Users\Kadir\Desktop\javacpp-presets\opencv\target\classes\org\bytedeco\javacp
p\jniopencv_objdetect.cpp(9036) : warning C4244: 'initializing' : conversion fro
m 'jlong' to 'jint', possible loss of data
Microsoft (R) Incremental Linker Version 12.00.31101.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:jniopencv_objdetect.dll
/ltcg
/dll
/implib:jniopencv_objdetect.lib
/OUT:C:\Users\Kadir\Desktop\javacpp-presets\opencv\target\classes\org\bytedeco\j
avacpp\windows-x86\jniopencv_objdetect.dll
/LIBPATH:C:\Users\Kadir\Desktop\javacpp-presets\opencv/cppbuild/windows-x86/lib/

opencv_world300.lib
jniopencv_objdetect.obj
   Creating library jniopencv_objdetect.lib and object jniopencv_objdetect.exp
jniopencv_objdetect.obj : error LNK2001: unresolved external symbol "public: __t
hiscall cv::DetectionBasedTracker::Parameters::Parameters(void)" (??0Parameters@
DetectionBasedTracker@cv@@QAE@XZ)
C:\Users\Kadir\Desktop\javacpp-presets\opencv\target\classes\org\bytedeco\javacp
p\windows-x86\jniopencv_objdetect.dll : fatal error LNK1120: 1 unresolved extern
als

C:\Users\Kadir\Desktop\javacpp-presets\opencv>
```

saudet:
Right, that seems to be a bug in OpenCV... I guess we'll have to create a
patch for that. Would you be able to look into this? Thanks!

jduke32:
why not.Yes i can test

jpsacha:
@saudet Why do you think it is a bug in OpenCV?

I did a little experiment. The linker complains about 
```
cv::DetectionBasedTracker::Parameters::Parameters(void)
```
[...]

PR #59, opened 2015-06-22
Fix creation of OpenCV 3 wrapper on Windows (issue #56).

jpsacha:
OpenCV is build from source on Windows. Patches are applied to correct issues in OpenCV.

Requires installation of Visual Studio 2013 and Windows version of CMake. The Visual Studio 2013 Community edition works fine too.

### DIFF
--- a/opencv/cppbuild.sh
+++ b/opencv/cppbuild.sh
@@ -7,26 +7,15 @@ if [[ -z "$PLATFORM" ]]; then
     exit
 fi
 
-if [[ $PLATFORM == windows* ]]; then
-    OPENCV_VERSION=3.0.0
-    [[ $PLATFORM == *64 ]] && BITS=x64 || BITS=x86
-    download http://downloads.sourceforge.net/project/opencvlibrary/opencv-win/$OPENCV_VERSION/opencv-$OPENCV_VERSION.exe opencv-$OPENCV_VERSION.exe
+OPENCV_VERSION=3.0.0
+download https://github.com/Itseez/opencv/archive/$OPENCV_VERSION.tar.gz opencv-$OPENCV_VERSION.tar.gz
+# download https://github.com/Itseez/opencv_contrib/archive/$OPENCV_VERSION.tar.gz opencv_contrib-$OPENCV_VERSION.tar.gz
 
-    mkdir -p $PLATFORM
-    cd $PLATFORM
-    7za x -y ../opencv-$OPENCV_VERSION.exe opencv/build/OpenCV* opencv/build/include opencv/build/$BITS/vc12/lib opencv/build/$BITS/vc12/bin
-    cd opencv
-else
-    OPENCV_VERSION=3.0.0
-    download https://github.com/Itseez/opencv/archive/$OPENCV_VERSION.tar.gz opencv-$OPENCV_VERSION.tar.gz
-#    download https://github.com/Itseez/opencv_contrib/archive/$OPENCV_VERSION.tar.gz opencv_contrib-$OPENCV_VERSION.tar.gz
-
-    mkdir -p $PLATFORM
-    cd $PLATFORM
-    tar -xzvf ../opencv-$OPENCV_VERSION.tar.gz
-#    tar -xzvf ../opencv_contrib-$OPENCV_VERSION.tar.gz
-    cd opencv-$OPENCV_VERSION
-fi
+mkdir -p $PLATFORM
+cd $PLATFORM
+tar -xzvf ../opencv-$OPENCV_VERSION.tar.gz
+# tar -xzvf ../opencv_contrib-$OPENCV_VERSION.tar.gz
+cd opencv-$OPENCV_VERSION
 
 case $PLATFORM in
     android-arm)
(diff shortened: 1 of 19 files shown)


saudet:
Good, thanks for the work on this! There's a few things that we're going to need to fix before this goes in though:
- We don't need to patch files for issues related to text encoding. We can tell Java to use something else than UTF-8 by putting a few lines in the `pom.xml` file, as shown in https://github.com/bytedeco/javacpp-presets/blob/master/flycapture/pom.xml#L37
- Visual Studio comes with a script file to set all the environment paths for us so we don't have to hard code them in `cppbuild.sh`.
- Since we set CMAKE_INSTALL_PREFIX, might as well leave the installation up to CMake instead of copying files manually with `cp` or anything else.

Thanks!!

jpsacha:
I will tray to take care of the UFT-8. 

The rest is not such a clear cut. While you can set paths for Visual Studio tools, CMake is not part of those tools and you have to use Windows version of CMake, you have to hard code that path. There is a lot of places in presets where windows paths are hard coded, almost every opencv java file annotation has `preloadpath = "C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/redist/x86/Microsoft.VC120.CRT/"`.

CMake install things differently on Windows than other platforms. While "include" is in the same location, the binaries are under `x64\vc12\bin` not `bin`, similarly `x64\vc12\lib`. So either you copy then to `bin` and `lib` or modify each preset to let it know that `bin` and `lib` is somewhere else.  It is really easier to copy them than fiddle with setting up paths. 

Actually I got stack when I removed copying on compilation of `cvkernels`. It cannot find libraries when linking. Not clear what is causing compilation of `cvkernels`. I was trying to adjust preset for `opencv_core` and add `linkpath` but looks that it is not passing library paths when linking `jnicvkernels`.

I am running out of time on this, may need to stop for a couple of weeks.

jpsacha:
All the changes are now rebased into a single commit. Ready to go.

saudet:
Thanks! Sorry for the delay. Let's see, `opencv_core.java` still gets modified although we don't need to modify it.

As for the hard coded paths in the presets, sure that's something I want to get rid of. If you have any good ideas, let me know! But for CMake at least, we don't need to hard code them. It can pick up the paths from the environment variables, even in the case of MSVC. Check inside `artoolkitplus/cppbuild.sh` for example. It works just fine, exactly the same way as on Linux, and doesn't install files in weird locations.

BTW, does `msbuild` support the install target or not? If it does, please try to use it! It might work better than you think. If it doesn't support it, we might as well stick with `nmake`, I suppose...

saudet:
Thinking about it a bit now, I know how this could work. In `cppbuild.sh` we could pick up the directory of MSVC from the `VSINSTALLDIR` environment variable, and copy the runtime libraries in the `target`, so we would not need to hard code any directories in the presets. Well, that's something I can do myself after, so don't worry about that.

jpsacha:
Please pay attention to changes in `opencv_core.java`. The Windows part is now more like Linux, linking to `"opencv_core300", "opencv_imgproc300"` instead of `"opencv_world300"`. That is also required when creating wrappers for `xfeatures2d` (but that will be a separate PR, once we are done with this one)

I do not see anything in `artoolkitplus/cppbuild.sh` related to CMake except simple invocation `$CMAKE -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=..`

MSBuild does invoke the `install` target:
```
MSBuild.exe INSTALL.vcxproj /p:Configuration=Release
```
`bin` and `lib` are copied from where it installs then, not from where it builds them. I had some trouble setting options for the old `nmake`, so I went for simpler `MSBuild`. With MSBuild you also have a project (solution) that you can load into an IDE, in case things go wrong with the build.

jpsacha:
I tried `nmake`. I cannot build OpenCV with it (outside of presets) it crashes while building `zlib`. I did not have luck building `artoolkitplus` preset neither. `cppbuild` just quits after downloading and un-tarring. No errors. Does not attempt to execute CMake.

I will not be able to work on it for a good couple of week. Bottom line: the setup with MSBuild in the current PR works fine. Maybe somebody can continue and make it into `nmake` or `msys` instead.

saudet:
The changes for the library names, that's good! I was referring to the corrupted double quote.

If the install target of `msbuild` works and puts things in `include`, `lib`, and `bin`, then that's fine too.

Are you sure you executed `vcvars32.bat` or `vcvars64.bat` or something like that before trying to build `artoolkitplus`? What to you get when you type `cl` in such an environment?

saudet:
Also, could you explain why you had to hard code the path for CMake? The installer should offer to put it in the system PATH...

saudet:
Anyway, please remove the corrupted double quote here: https://github.com/jpsacha/javacpp-presets/commit/302eda60b6034854b5fddc6f93272c0331167006#diff-098f93602679a6676e7669064dd1a9ebR16800 and I'll merge what you have! Thanks

saudet:
I've tested this and it's working well overall. Thanks a lot! I made couple of small changes/fixes, but that's it. It looks like the double quotes corruption is just a glitch in GitHub: it doesn't show up in git at all.

Issue #54, opened 2015-06-15
Add support for xfeatures2d module of OpenCV 3


jpsacha:
Some of the algorithms previously available in javacpp-presets  moved in OpenCV 3 to xfeatures2d module. Those  include SIFT, SURF, and related.There are also some other interesting algorithms in xfeatures2d module, like DAISY.

saudet:
The problem with `opencv_contrib` is that they don't provide binaries for Windows. Would you be able to adjust the `cppbuild.sh` file to build OpenCV with `opencv_contrib` from source for Windows as well? In addition, it would be a good idea to build it with MinGW-w64. This way, we can build under Linux as well, and consequently on Travis CI for free.

BTW, if the only thing that you are worried about are descriptors, AKAZE usually provides better results than SIFT, SURF, DAISY or anything else, and doesn't have any of their legal issues either.

jpsacha:
I was looking at updating the cookbook examples, I have most under control (as far I can tell so far), except examples that use SURF (1) and SIFT (6). Some of the most interesting/elaborated examples use SIFT.

I can compile OpenCV on Window with Visual Studio. I was not able to do that with MinGW, getting syntax errors, not clear how to it should really be setup on Windows. I tried using MinGW tool chain under Cygwin using cmake:

```
cmake -DCMAKE_CXX_COMPILER:FILEPATH=/usr/bin/i686-w64-mingw32-g++ -DCMAKE_C_COMPILER:FILEPATH=/usr/bin/i686-w64-mingw32-gcc -DWITH_OPENEXR=NO ../opencv.git
```
 that can create cmake setup. It compiles part but eventually ends in errors like:

```
usr/i686-w64-mingw32/sys-root/mingw/include/float.h:278:57: error: expected ‘,’ or ‘...’ before numeric constant
```

If this is not possible to add xfeatures2d, I will probably drop I will drop SURF and SIFT from the cookbook examples.

saudet:
Well, if you get it working with MSVC, that'll be alright for now too. Please send your modifications to `cppbuild.sh`!

As for SURF or SIFT in the cookbook examples, those could be converted to AKAZE. Check the changes to https://github.com/bytedeco/javacv/blob/master/src/main/java/org/bytedeco/javacv/ObjectFinder.java for an example.

jpsacha:
OK. I will try to setup presets to build with VS2012/VS2013 and xfeatures2d.

[...]

PR #75, opened 2015-07-25
Add wrapper for `xfeatures2d` module (issue #54).

jpsacha:


saudet:
Thanks! Though the `opencv_core.java` files appears to contain corrupted characters. Could you fix that? I wonder what causes that.

Also, the build commands for Android were not updated, but this is something I can do, no problem!

jpsacha:
I fixed them before committing, at least `diff` was indicated that they were fixed (that there were no changes). I will try to just commit old version of the `opencv_core.java`. This PR does not need to modify anything in `opencv_core.java`.

jpsacha:
PR is updated. `opencv_core.java` is not shown as changed by the PR.

Issue #65, opened 2015-06-30
Huge size of the java bindings for Leptonica


maximumspatium:
Hello Samuel,

**Issue description:**

Tesseract has the binary dependency on Leptonica that currently counts about 2500 functions. Tesseract utilizes only 118 Leptonica functions.
Unfortunately, Leptonica has been designed as a huge monolithic library incorporating several independent image processing technologies. When Tesseract's dependencies are being resolved, the whole Leptonica library will be loaded into RAM but only 4,8% of its code will be actually used!

Moreover, the way Tesseract external API is designed, we have to go through Leptonica in order to be able to get Tesseract data in and out. Therefore, we need to have a java bindings for Leptonica which adds another 5,3 MB unused code laid out in RAM.

**Proposed solution:**

Add another JNI wrapper called "leptonica-min" that provides java bindings for the Leptonica functions used by the Tesseract external API. This (tiny) wrapper should be shipped together with the Leptonica preset.

Projects relying on Tesseract only could link to "leptonica-min" and thus save a significant amount of memory (5MB!!!). Projects relying on further Leptonica functionality could link to the full Leptonica wrapper instead provided in the same JAR.

**Question:**
What do you think of the above mentioned solution?

Best regards
Maxim

saudet:
Sounds good! And if it's small enough, it could even be bundled with Tesseract, and we could save some more hassle that way. ;)

Is there already a "min" config readily available for Leptonica?

maximumspatium:
> Is there already a "min" config readily available for Leptonica?

No, but I'm currently working on such a config. One thing I ain't quite sure about is which file to strip in order to achieve this goal. Leptonica presets contains two files: "presets/lept.java" and "lept.java". The former seems to be the main config as described in the Wiki's "Create new presets". The latter seems to be machine-generated. Its purpose is opaque to me...

Will it be enough to strip Java files? Or should I strip Leptonica's main header - "allheaders.h" - instead?

Thank you in advance!
Best regards
Maxim

saudet:
You're going to need to modify Leptonica itself. The Java files are auto-generated yes, so we can do that after.

saudet:
Hi there, any progress?

saudet:
In my opinion, the effort required to save a few megs here is not worth it, so I'm closing this, but if you would still like to spend some time on this, please let me know! Thanks

PR #80, opened 2015-08-24
First attempt to reduce Leptonica's size

maximumspatium:
Hi,

I've managed to create the "minified" configuration for Leptonica as suggested in the [Issue #65](https://github.com/bytedeco/javacpp-presets/issues/65). It greatly reduces the binary size of the Leptonica bindings from >5Mb to 407Kb!

I wrote a small Python script that searches Tesseract's source for Leptonica functions. This pull request doesn't include the script but I'm going to share it with you.

Tesseract currently utilizes 160 from 2481 functions available in Leptonica. This number counts for methods used for training as well but because we don't build training tools we could remove these extra methods.

The following issues need to be resolved:
- package this minified version with Tesseract somehow so the Leptonica preset won't be affected
- try to reduce code size even more. At the time being, it seems that deleting the ".put" statements from InfoMap doesn't stop Javacpp from generating classes for objects extracted from C headers. So we end up having Java classes for unused Leptonica structs. Maybe something like "InfoMap.ignore()" would solve the puzzle...

### DIFF
--- a/tesseract/allheaders_min.h
+++ b/tesseract/allheaders_min.h
@@ -1,0 +1,244 @@
+/*====================================================================*
+ -  Copyright (C) 2001 Leptonica.  All rights reserved.
+ -
+ -  Redistribution and use in source and binary forms, with or without
+ -  modification, are permitted provided that the following conditions
+ -  are met:
+ -  1. Redistributions of source code must retain the above copyright
+ -     notice, this list of conditions and the following disclaimer.
+ -  2. Redistributions in binary form must reproduce the above
+ -     copyright notice, this list of conditions and the following
+ -     disclaimer in the documentation and/or other materials
+ -     provided with the distribution.
+ -
+ -  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ -  ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ -  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ -  A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL ANY
+ -  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ -  EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ -  PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ -  PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ -  OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ -  NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ -  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *====================================================================*/
+
+#ifndef  LEPTONICA_ALLHEADERS_H
+#define  LEPTONICA_ALLHEADERS_H
+
+
+#define LIBLEPT_MAJOR_VERSION   1
+#define LIBLEPT_MINOR_VERSION   72
+
+#include "alltypes.h"
+
+#ifndef NO_PROTOS
+/*
+ *  These prototypes were autogen'd by xtractprotos, v. 1.5
+ */
+#ifdef __cplusplus
+extern "C" {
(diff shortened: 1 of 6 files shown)


saudet:
Looking good! Yes, that should become part of the Tesseract module somehow...

As for ignoring classes and what not, there's `InfoMap.skip()` for that...

maximumspatium:
I've just updated the Java configuration to skip unused classes. This has caused a further size reduction to 297Kb on x86. Please review.

What's the next step to do?

saudet:
Include it as part of Tesseract I guess. It sounds reasonable...

saudet:
Bundling it as part of Tesseract would look something like what is done for FFmpeg, so be sure to check that one out for an example. Thanks!

maximumspatium:
Can you be a bit more specific? I need to build two different versions of the same library based on the same source and dependencies. Therefore, only the configuration need to be changed. Unfortunately, two different configs aren't allowed in the same preset. Could you provide me with a hint of how to solve this dilemma?
Thank you in advance!

saudet:
I see your changes as becoming part of the Tesseract presets. I don't really see a better way of doing this, but I'm open to suggestions.

saudet:
Basically, Leptonica would get built twice: Once for the Leptonica presets, and again once, independently for the Tesseract presets. That's what you want isn't?

maximumspatium:
Hello Samuel,

I've just updated my pull request offering a minified Leptonica build bundled together with the Tesseract preset. The trick was to "borrow" Leptonica's cppbuild.sh script using "source" and place all required binaries in the Tesseract's install directory.

The modified preset builds successfully as far. I couldn't test whether it actually works with Java because there is no any unit test. I think it would be great to have unit tests for our presets. It sounds like a next major enhancement...

Please review.
Best regards
Max

saudet:
Looks good, yes! Could you clean up spurious unrelated changes and squash all the commits into one? (See http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit) Thanks!

maximumspatium:
> Could you clean up spurious unrelated changes and squash all the commits into one? 
Done! All related changes has been squashed. Unrelated ones has been deleted. I've rebased all my changes ontop of your last commit.

I had to do a force push because two branches have been diverged too much.

I hope that everything is OK now and that I didn't introduce any regressions...

saudet:
Thanks for the squash! Force push is fine for temporary branches and forks. I've tried your new build, and although it works fine, we still get a library file for Leptonica that is about 4 megs large, so what is the advantage over the current standalone presets for Leptonica?

Also the pull request still contains a couple of unrelated changes. We should also
- Remove the execute permission from the `presets/lept.java` file
- Move `allheaders_min.h` under `src/main/resources/org/bytedeco/javacpp/`

Anyway, let's get Leptonica sized down first before worrying about other things, thanks!

maximumspatium:
Hello Samuel,

> I've tried your new build, and although it works fine, we still get a library file for Leptonica that is about 4 megs large, so what is the advantage over the current standalone presets for Leptonica?

Well, the goal of my changes is to reduce the size of the JNI bindings for Leptonica, that's ```libjnilept.so```. Therefore, the title of this pull request might be a bit misleading. Reducing the size of Leptonica ITSELF is IMHO beyond the scope of this project. This is something that should be directly negotiated with the upstream.

Currently, we have two libraries - ```liblept.so.4``` and ```libjnilept.so```, both approx. 3,5 MB large (this is true for x86. Other architectures will have different sizes). My configuration reduces the size of the latter down to 300 KB making another 3 MB available for Java applications. Well, not really a big hit, but still too costly for storing dead code, especially on mobile devices with limited memory.

Could you kindly check whether the size of libjnilept.so is getting smaller with my patch?

If you think the outcome doesn't worth the price, we could look for a better solution...

saudet:
I see. Well, if we link Leptonica statically, we might save a lot of space, and that would make more sense. Have you tried that? What kind of results would that give?

maximumspatium:
> I see. Well, if we link Leptonica statically, we might save a lot of space, and that would make more sense. Have you tried that?

Well, I haven't tried that but that sounds like a good idea to me. I'll give a try...

maximumspatium:
> if we link Leptonica statically, we might save a lot of space, and that would make more sense. Have you tried that?

I've just tried that way. It looks like it doesn't work out of the box. Disabling generation of shared libraries for Leptonica immediately breaks Tesseract configuration. Manipulating linker flags via LDFLAGS has the same effect.
I'm afraid I have to patch Tesseract's (bloated) build scripts. Ufff...

saudet:
Let's see, we could link with the minimal set of the `.o` found in the `leptonica-1.72/src/.libs/` subdirectory to cover the functionality exposed by `allheaders_min.h`. To do that, we could either modify the build of Leptonica, or the one of Tesseract. I guess it would make most sense to make a patch for Tesseract to link with a list of `.o` files instead of the whole `lept` library.

saudet:
In my opinion, the effort required to save a few megs here is not worth it, so I'm closing this, but if you would still like to spend some time on this, please let me know! Thanks

PR #121, opened 2015-12-02
Correct hdf5 download link.

jpsacha:
The link for downloading hdf5 was no longer valid.


saudet:
Thanks! I guess we should update the version too while we're at it.

PR #123, opened 2015-12-05
Incremented openssl dependency version for ffmpeg

paulheideman:
openssl-1.0.2d appears to be MIA...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,31 @@
 language: cpp
-dist: trusty
+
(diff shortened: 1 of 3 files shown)


saudet:
Sure, thanks. There appears a lot of unrelated commits in this PR though. Could you clean this up? Thanks!

paulheideman:
Hmmm... I'll need to work that out. It looks like it has pulled in my playing with travis too.

PR #153, opened 2016-02-18
remove dep on xz

ashleyj:
The shell script llvm/cppbuild.sh does not run on OS X 10.11.x as it no longer ships with the 'xz' binary. This PR shifts all the handling to 'tar' -- all modern implementations will support the 'z' switch and 'xz' compression.

### DIFF
--- a/llvm/cppbuild.sh
+++ b/llvm/cppbuild.sh
@@ -14,11 +14,11 @@ download http://llvm.org/releases/$LLVM_VERSION/cfe-$LLVM_VERSION.src.tar.xz cfe
 mkdir -p $PLATFORM
 cd $PLATFORM
 INSTALL_PATH=`pwd`
-xz -dc ../llvm-$LLVM_VERSION.src.tar.xz | tar -xvf -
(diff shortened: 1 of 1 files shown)


saudet:
Are you sure? Have you tested that?

saudet:
I get this on Fedora 23:
```
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

ashleyj:
Ahh yes, you're right - 'z' was the incorrect switch. This switch is actually ignored in bsdtar when used with 'x'.  Have committed the change with the correct switches 'xvf' -- tested on centos 6.5 with gnu tar 1.23 and OS X 10.11 with bsd tar 2.8.3

saudet:
Actually, we could use the `J` flag to make this explicit, but yes it works without installing xz! Good to know.

I'll merge this in, but could you squash this into one commit? The easiest way is to `git reset master` your branch, recommit, and `git push --force`. Thanks!

ashleyj:
done

saudet:
Cool, thanks!

PR #158, opened 2016-02-21
Added files via upload

lobsteroh:
added my initial work on javacpp-presets in case someone has the ability to finish this one.


saudet:
Thanks! But It's only showing up the API for Leptonica... ?

PR #161, opened 2016-02-22
opencv with working test and travis integration / javacpp-presets needs reorganization

jjYBdx4IL:
There are two or three things one should absolutely not do in maven:

1.) define actual dependencies in parent descriptors.
2.) define plugin executions in parent descriptors.
3.) constructing package version according to tools that have been used to build them.

What I would recommend:

regard javacv/javacpp tooling as tooling only, give every library its own project, and call artifacts like "opencv-3.0.0.jar" etc. If a release needs fixing, call it "opencv-3.0.0.1.jar" or "opencv-3.0.0-1.1" if maven allows that.

Can we agree on that? I wouldnt want to invest more work before we can agree on a proper solution.

Technically, it seems to really work well what you have done there. Really great job. But we need to get the organization under control. :-)


saudet:
We've discussed some of that before. For the version number, there's like `opencv-3.0.0-1.0` that works with JavaCPP 1.0 and `opencv-3.0.0-1.1` that works with JavaCPP 1.1. How would you differentiate that in your scheme?

And what are you recommendations for the other problems you point out? How do `open-arch` and `open-it` submodules address these issues? It's not clear to me, so please elaborate! Thanks

jjYBdx4IL:
what do you mean by saying "works with JavaCPP 1.1"? The opencv module descriptor should already contain all necessary dependency information, doesn't it?

saudet:
Yes, it does, but say we had used simply version `3.0.0` for the OpenCV module we had created with JavaCPP 1.0. Then when we would have updated it with JavaCPP 1.1, because there was no new version of OpenCV, it's still version 3.0.0. But what version would you give to this updated OpenCV module? We need to give it another version number, or Maven will complain. What would it be?

jjYBdx4IL:
It's more about the warnings maven is printing out because of how you define the version in there. It even gets uploaded to maven central without being replaced first:

https://repo1.maven.org/maven2/org/bytedeco/javacpp-presets/opencv/3.0.0-1.1/opencv-3.0.0-1.1.pom

The -arch module basically replaces the construct where you try to add dependencies to sub-modules in the parent/aggregator module javacpp-presets. Trying to propagate plugin executions and explicit dependencies from parent/aggregator modules is not good. One example: it prevented me from creating a test module there. And if you pull the dependency declaration for the binary jar down to opencv, it even gets circular because the opencv module now depends on its own output ie on itself. That's why I moved the binary dependency to the -arch pkg which really is -autoarch :-)

saudet:
Ok, sounds like a reasonable solution -- for one platform. What would you suggest for users that want to slurp in all dependencies for all platforms?

jjYBdx4IL:
Hmm. We could create -allarch meta modules for that.

How different are the binary projects? Are there manual, library-specific adjustments beyond building the binaries (cppbuild.sh) required somewhere? In other words: can't we just drop the opencv modules etc from the repository and create them on-the-fly?

saudet:
You're right, we could have a module for all platforms, and it could derive from a parent where we keep a list of all platforms, and I think a JavaCV user (for example) would also be able to select which one (single platform or all platforms) module with a system property. And I think that would work with SBT and Gradle too. Interesting. If we can get everything working for JavaCV with Maven, Gradle, and SBT, I'd seriously consider switching! Could you try it out? It would help a lot :)

Maven artifacts are supposed to be binaries, by definition. That's what they are. People are free to build from source, though. That's what is on the repo here on GitHub. :) So, I'm not sure I understand the question... ?

jjYBdx4IL:
I was talking about the native binary modules only.

saudet:
Well, just think of it that way: Are you running Gentoo? Why not?

saudet:
Ah, I think I misunderstood, you weren't talking about the binary files themselves, just the project files. Yes, those are adjustments too. Check for example Caffe, that depends on OpenCV. But in any case, making archtypes is a good idea! Could you create a demo of that? Would be awesome

jjYBdx4IL:
sry, too much effort right now just to get a few tests going.

saudet:
Yes, it takes time, so that's why I need help... What would you like to do that wouldn't take too much of your time? We don't need to do everything right away. Just a small thing here and there, that will eventually result in big changes, that's fine.

jjYBdx4IL:
well, the easiest thing to do would be to pull down all native libs by one directory and inject a native libs parent module. Into that module (libs-parent) we could move all the executions and dependency declarations from the javacpp-presets module. The new structure would be sth like:

```
javacpp-presets/pom.xml
  |
  +--libs/pom.xml
  |   +--libs/opencv/pom.xml
  |   +--libs/ffmpeg/pom.xml
  |   + ...
  +--opencv-it/pom.xml
  + potentially more test modules
```

This would be a first step that at least allows non-native-libs modules to be added to the project and which does not require any real structural changes.

saudet:
javacpp-presets isn't meant to contain non-native artifacts, so what
non-native artifacts would you like to add? just the tests or something
else?

if you mean that the platform specific artifacts should be built via
another pom.xml file, that sounds reasonable. but in that case, I don't
think Maven would allow us to use the same artifact names for the artifacts
like we have now. does it? if not, what naming scheme would you propose?

jjYBdx4IL:
yeah, just tests.

I don't see why one would need to rename them. Just switch out the parent so they aren't directly childs to javacpp-presets parent.

saudet:
Let's see, what I'm thinking of doing is to add something like a "sample" directory under each current submodule (like "opencv/sample", "ffmpeg/sample", etc), which would basically contain the "pom.xml" file and the source file currently found in each "README.md" file. And then as test for each submodule, we could simply run the corresponding sample, and make sure it runs without errors. What do you think? I'm pretty sure that works fine on my machine anyway.

PR #164, opened 2016-02-24
add definition for Map<string, AttrValue>

### LINKED ISSUE #162: [Tensorflow] Reading out attributes from NodeDef
I am trying to read out shape information from TensorFlow placeholders at the moment (these are the input nodes of a Tensorflow graph). For this, access to the attr field of NodeDef is required. The definition is as follows (see `tensorflow/core/framework/graph.proto`):

```proto
message NodeDef {
  // stuff
  map<string, AttrValue> attr = 5;
}
```
and the generated code I am interested in wrapping looks like this (see `graph.pb.h` generated by BAZEL):
```cpp
  ::google::protobuf::Map< ::std::string, ::tensorflow::AttrValue >*
      mutable_attr();
```
With current master, the parser doesn't seem to be able to handle this. I tried to add
```java
.put(new Info("std::map<std::string,tensorflow::AttrValue>").pointerTypes("StringAttributes").define())
```
in the tensorflow.java configuration file and this leads to StringAttribute being defined which looks reasonable, but the field is still not wrapped.

Any help on this is appreciated!

pcmoritz:
this fixes #162

### DIFF
--- a/tensorflow/src/main/java/org/bytedeco/javacpp/presets/tensorflow.java
+++ b/tensorflow/src/main/java/org/bytedeco/javacpp/presets/tensorflow.java
@@ -101,6 +101,7 @@ public class tensorflow implements InfoMapper {
                .put(new Info("tensorflow::Allocator::is_simple<bfloat16>").skip())
 
                .put(new Info("basic/containers").cppTypes("tensorflow::gtl::InlinedVector"))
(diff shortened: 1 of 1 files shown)


saudet:
Thanks! BTW, we can merge the `new Info("basic/containers")` statements, and for consistency I think we should name the type `StringAttrValueMap`?

pcmoritz:
You are right, that's better. I changed it.

PR #169, opened 2016-03-01
fix openblas build, add sample

jjYBdx4IL:
caffe build is failing here because no TARGET is given to openblas make.


jjYBdx4IL:
ah, let me check how to clean up the commit history....

jjYBdx4IL:
That's better.

saudet:
That's a good start, thanks, but I still would prefer if we could find a way to use the sample as is. Are there any reasons why we couldn't use this pom.xml file exactly as it is inside, for example, opencv/sample:
```
<project>
    <modelVersion>4.0.0</modelVersion>
    <groupId>org.bytedeco.javacpp-presets.opencv</groupId>
    <artifactId>stitching</artifactId>
    <version>1.1</version>
    <properties>
        <exec.mainClass>Stitching</exec.mainClass>
    </properties>
    <dependencies>
        <dependency>
            <groupId>org.bytedeco.javacpp-presets</groupId>
            <artifactId>opencv</artifactId>
            <version>3.0.0-1.1</version>
        </dependency>
    </dependencies>
</project>
```
And have the test plugin as part of the opencv/pom.xml module execute that?

saudet:
I've included your fix for OpenBLAS together with another fix I had in the pipeline ;) Thanks

As for sample data, we'll need to include something, I think. It doesn't need to be large, so something like the images used here: https://github.com/bytedeco/sample-projects/tree/master/opencv-stitching  I'm not sure what the copyright is on that though...

saudet:
It's looking good, but is it possible to make this work with `maven-surefire-plugin` somehow? That's what I'd really like to know. If it's possible to use a sample like that as test, it would make it a lot easier to create samples and tests for all the other presets.

jjYBdx4IL:
What do you want to test? Stitching.java only? The whole sample module?

saudet:
I'm thinking we could use the current samples (the stitching one in the case of opencv, but other presets for ffmpeg, caffe, etc have their own samples, check them out) with some sample data, of course, as integration tests themselves. It won't test everything, but I think it's good enough. And it would be ideal if these same tests could serve a double purpose as sample project files. 

So, that's why I was wondering if we could use maven-surefire-plugin to execute a sample project like that. If the project fails to execute, we make the test fail. That would be it. Do you know if this is easily doable? I realize you don't have a lot of time, and I don't either, so that's why I want to find the simplest and most effective way of adding both tests and sample project files.

PR #173, opened 2016-03-04
Opencv sample

jjYBdx4IL:
how about this? this is your sample very slightly modified so it can be tested using surefire. and the sample pom has the testing stuff in a clearly separated testing profile.

I could add a similarity test for input and output pictures. Where in opencv is that again? XD

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@
 # IntelliJ
 *.iml
 .idea
+/opencv/sample/target/
--- a/nb-configuration.xml
+++ b/nb-configuration.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project-shared-configuration>
+    <!--
+This file contains additional configuration written by modules in the NetBeans IDE.
+The configuration is intended to be shared among all the users of project and
+therefore it is assumed to be part of version control checkout.
+Without this configuration present, some functionality in the IDE may be limited or fail altogether.
+-->
+    <properties xmlns="http://www.netbeans.org/ns/maven-properties-data/1">
+        <!--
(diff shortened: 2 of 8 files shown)


jjYBdx4IL:
we could move the test to the opencv module and run the sample via "mvn exec:java -f sample" from there inside the surefire test.....

saudet:
Yes, that sounds good. I see you've started using `maven-failsafe-plugin`, which seems better than `maven-surefire-plugin` yes, and OpenIMAJ looks interesting too: Good idea! Let me know when you get this working with a sample through `mvn exec:java -f sample` command, or something. Thanks!

jjYBdx4IL:
should work as it is

saudet:
Why is there both `Stitching.java` and  `StitchingIT.java`. It's not possible to use `Stitching.java` for the maven-failesafe-plugin as well?

jjYBdx4IL:
StitchingIT *is* the maven failsafe test to run Stitching.java.

saudet:
Sorry for the late reply. I have a lot to deal with these days, but your proposal is starting to sound really interesting. Ok, so we need a file like `StitchingIT.java` for JUnit. Now, I've been thinking, would it be possible to add a generic `SampleIT.java` to the parent project and have that file included in all submodules? The parent does not and should not compile the file. It should merely be included as a source file for the submodules. My hope is that we wouldn't need to hard code the arguments. I think it's possible, and that's the kind of thing I was saying would be really nice to have. It would make it a lot easier for people to create **and test** new presets!

saudet:
I don't want you to close this :( I'd like to enhance this to make it easier to test the other presets as well. I think you have an awesome idea here. It just needs to be generalized a little bit!

PR #174, opened 2016-03-04
add netbeans editor settings to pom.xml (tab size 2)

jjYBdx4IL:
is tab size 2 valid for all file types or just pom.xml?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,6 +84,21 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <javacpp.parser.skip>false</javacpp.parser.skip>     <!-- To skip header file parsing phase: -Djavacpp.parser.skip=true  -->
     <javacpp.compiler.skip>false</javacpp.compiler.skip> <!-- To skip native compilation phase: -Djavacpp.compiler.skip=true -->
+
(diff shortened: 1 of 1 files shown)


saudet:
I try to follow Maven's convention here:
https://maven.apache.org/developers/conventions/code.html
So, 2-space indents for the XML files, but 4 spaces for Java source files...

(Hum, they seem to have changed the recommended ordering of the elements inside pom.xml files.)

jjYBdx4IL:
I don't see any possibility to set indentation for XML in Netbeans, so I set default to 2, and change java to 4.

saudet:
There are settings for XML, for example:
https://github.com/mojohaus/nbm-maven-plugin/blob/nbm-maven-plugin-3.1/pom.xml
However, almost no one uses those settings in the pom.xml file. Even the `nbm-maven-plugin` have since removed them from their pom.xml:
https://github.com/mojohaus/nbm-maven-plugin/blob/master/pom.xml
Looks like there might be something bad about using those properties...

saudet:
http://editorconfig.org/ appears to be a much better alternative. If you create a `.editorconfig` file, please send a pull request for that! Thanks

PR #175, opened 2016-03-04
use antrun maven plugin to run cppbuild scripts within the maven build

jjYBdx4IL:


saudet:
Looks good, thanks! But the maven-antrun-plugin screws up the output of the script. exec-maven-plugin doesn't appear to have that problem, so could we use that instead? 

And we need to pass on the user-specified `javacpp.platform` property to the `-platform` option of the script. And I'll merge that in! thanks :)

jjYBdx4IL:
yeah that's better

saudet:
Looks good! So, could you squash this into one commit so it will merge cleanly?

BTW, the environment variables set aren't used for anything yet, so we could remove them, but if you'd like to keep them, let's make the names consistent. Something like the following maybe?
```xml
            <PROJECT_VERSION>${project.version}</PROJECT_VERSION>
            <PROJECT_BASEDIR>${project.basedir}</PROJECT_BASEDIR>
```
Thanks!

jjYBdx4IL:
done

saudet:
Thanks!

PR #177, opened 2016-03-06
add additional linux-armhf builds

vb216:
I've tuned the arm parameters a bit more, so they may well not run on the pi1 but should do better on pi2 + pi3. 

Added them into artoolkit, flandmark, libfreenect and libdc1394 so they seem to build OK for arm too, and modified the pom.xml for flycapture so it will skip the cppbuild but produce a jar.

This means that:

mvn install -Pffmpeg -Djavacpp.platform=linux-armhf -Djavacpp.platform.compiler=arm-linux-gnueabihf-g++ -Dplatform=linux-armhf -Dplatform.compiler=arm-linux-gnueabihf-g++ -Dmaven.test.skip=true

Builds fine. So that should be I'll be able to add into Jenkins and get the produced jars automatically into Maven (either automatically on new commits, or just for tags). 

(something seems up with openssl-1.0.2f - I think they've moved to version g, but I get something about needing to re run make depends with that and then build failures)


saudet:
Looks good, thanks! Actually, flycapture has a version for linux-arm as well: https://www.ptgrey.com/flycapture-sdk   We need register to download, but it's free. Anyway, it should work just like linux-x86, so no need to modify that pom.xml file or anything related to the flycapture module.

And could you squash everything into one commit? The easiest way is to `git reset master` your branch, recommit, and `git push --force`, and I'll merge this in. Thanks!

vb216:
Hm are you sure about the flycapture? I think the linux* was matching, but I got errors that it couldn't find header files, that's why I added in the skip cpp part just for linux-armhf (didn't want to mess up any existing linux stuff that worked)

+sure will sort out the squash this evening

saudet:
Have you installed the FlyCapture SDK? If so, where did the header files end up being copied?

vb216:
closing to redo a few bits

saudet:
You don't need to close this. Just keep on modifying your branch and the changes will automatically appear in this pull request.

vb216:
Finally got there I think

PR #179, opened 2016-03-12
fix arm setting for opencv

vb216:
Correction to armhf settings for opencv


saudet:
It's better without compiler flags?

vb216:
Actually... needs more thought. At the moment it wont compile with those flags at all, but I'd like to try and get a build that works across more platforms. No joy yet..

saudet:
Anyway, I'll merge in what you have here and in time others will hopefully help us figure out a better set of flags! Thanks!

PR #190, opened 2016-04-04
[WIP] Feature/clandmark preset

lloydmeta:
Would like some feedback on this PR. In short: this is me hacking things together until:

- `./cppbuild.sh install clandmark`
- `mvn clean install --projects clandmark -X`

both ran without errors.

### DIFF
--- a/clandmark/cppbuild.sh
+++ b/clandmark/cppbuild.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+# This file is meant to be included by the parent cppbuild.sh script
+if [[ -z "$PLATFORM" ]]; then
+    pushd ..
+    bash cppbuild.sh "$@" clandmark
+    popd
+    exit
+fi
+
+CLANDMARK_VERSION=master
+download https://github.com/uricamic/clandmark/archive/$CLANDMARK_VERSION.zip clandmark-$CLANDMARK_VERSION.zip
+
+mkdir -p $PLATFORM
+cd $PLATFORM
+INSTALL_PATH=`pwd`
+mkdir -p include lib bin
+unzip -o ../clandmark-$CLANDMARK_VERSION.zip
+cd clandmark-$CLANDMARK_VERSION
+
+OPENCV_PATH=$INSTALL_PATH/../../../opencv/cppbuild/$PLATFORM/
+
+case $PLATFORM in
+    android-arm)
+        $CMAKE -DCMAKE_TOOLCHAIN_FILE=$INSTALL_PATH/../../android-arm.cmake -DCMAKE_BUILD_TYPE=Release -DOpenCV_DIR=$OPENCV_PATH/sdk/native/jni/ -DANDROID_NDK_ABI_NAME=armeabi_v7a -DCMAKE_INSTALL_PREFIX=..
+        make -j4
+        make install clandmark
+        ;;
+    android-x86)
+        $CMAKE -DCMAKE_TOOLCHAIN_FILE=$INSTALL_PATH/../../android-x86.cmake -DCMAKE_BUILD_TYPE=Release -DOpenCV_DIR=$OPENCV_PATH/sdk/native/jni/ -DANDROID_NDK_ABI_NAME=x86 -DCMAKE_INSTALL_PREFIX=..
+        make -j4
+        make install clandmark
+        ;;
+    linux-x86)
+        CC="$OLDCC -m32" CXX="$OLDCXX -m32" $CMAKE -DCMAKE_BUILD_TYPE=Release -DOpenCV_DIR=$OPENCV_PATH/share/OpenCV/ -DCMAKE_INSTALL_PREFIX=..
+        make -j4
+        make install clandmark
+        ;;
+    linux-x86_64)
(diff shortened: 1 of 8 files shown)


clandmark/cppbuild.sh
```diff
@@ -0,0 +1,69 @@
...
+        ;;
+    linux-armhf)
+        CC=arm-linux-gnueabihf-gcc CXX=arm-linux-gnueabihf-g++ $CMAKE -DCMAKE_BUILD_TYPE=Release -DOpenCV_DIR=$OPENCV_PATH/share/OpenCV/ -DCMAKE_INSTALL_PREFIX=..
+        make -j4
+        make install clandmark
+        ;;
+    macosx-*)
+        CXX="g++ -fpermissive" $CMAKE -DCMAKE_BUILD_TYPE=Release -DOpenCV_DIR=$OPENCV_PATH/share/OpenCV/ -DCMAKE_INSTALL_PREFIX=..
+        make -j4
+        make install/strip
```
lloydmeta: I've only tested this one, since I only have 1 OS

clandmark/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+    <parent>
+        <groupId>org.bytedeco</groupId>
+        <artifactId>javacpp-presets</artifactId>
+        <version>1.2-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.bytedeco.javacpp-presets</groupId>
+    <artifactId>clandmark</artifactId>
+    <!-- TODO figure out what version CLandmark is actually at?!?!? -->
+    <version>0.01-${project.parent.version}</version>
```
lloydmeta: Could not figure out what version CLandmark is at, so left this at 0.01..

clandmark/src/main/java/org/bytedeco/javacpp/presets/clandmark.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * @author Lloyd @ github.com/lloydmeta
+ */
+@Properties(value = @Platform(
+        define = "_MSVC_COMPAT__",
+        include = {"CLandmark.h", "Flandmark.h", "CFeaturePool.h", "CFeatures.h"},
+        link = { "clandmark@.1.5", "flandmark@.1.5"}),
+        inherit = opencv_imgproc.class, target = "org.bytedeco.javacpp.clandmark")
+public class clandmark implements InfoMapper {
```
lloydmeta: Please take a look at the mappings and annotations here...

lloydmeta:
CC @saudet  and @jpsacha

clandmark/src/main/java/org/bytedeco/javacpp/presets/clandmark.java
```diff
@@ -0,0 +1,44 @@
...
+    public void map(InfoMap infoMap) {
+        infoMap.put(new Info("fl_double_t", "clandmark::fl_double_t").cast().valueTypes("float").pointerTypes("FloatPointer").define());
+        infoMap.put(new Info("cimg_library::CImg<unsigned char>").cast().pointerTypes("BytePointer"));
```
lloydmeta: This one in particular is dubious; "unsigned char" itself needs translation, and `CImg` is a generic type..

lloydmeta:
After taking another look, I realised that CImg is not in fact what I wanted to cast it as; looks like I need to include `CImg.h` in order to have that class available, but then once I include it, the resulting file fails to compile with a lot of errors. Definitely need some more experienced eyes looking at this.

saudet:
Right, the problem is, it's hard to figure out automatically which header files contain the actual interface we're interested in, and which ones contain implementation details and what not. To compensate, users need to include all the files they want to process, manually, in the correct order. So, check inside `CImg.h` to see which other files get included, and list those up as well, a potential recursive process... :)

lloydmeta:
No, `CImg.h` is the only file in that library, albeit a 2.59MB one; see [the repo](https://github.com/dtschump/CImg). The project prides itself on being one single header file.

So, yeah, look for `struct CImg {` (line 9307) and you will see it's in that file, which subsequently includes a *lot* of stuff (the entire file is 46180 lines long), some of which the parser can't handle, and likely a lot that the parser isn't handling properly. Some more specific details on what to do in this situation would be helpful.

saudet:
Ah, I see. The CImg library is separate from CLandmark though, but it's a requirement? Hum... It's a template library, so we'll need to create instances manually, as shown for Caffe here for example:
https://github.com/bytedeco/javacpp-presets/blob/master/caffe/src/main/java/org/bytedeco/javacpp/presets/caffe.java#L95
So, for example, we'd have `ByteCImg` for `CImg<unsigned char>` I guess.

lloydmeta:
Ah ok, let me give that a try ! Thanks.

saudet:
Yeah, it looks like it's going to be a hard one if we need to parse CImg as well :(

saudet:
Maybe we could provide a small amount of accessor methods in a custom header file, and map those instead of all of CImg? Sounds a lot easier to do! Want to try this out?

lloydmeta:
@saudet I've hard-pressed for time and so didn't really progress with this PR. I'll close it now to give someone else a try. Your idea sounds great: maybe you should try it out and see how far you get with it?

saudet:
Or @jpsacha?

saudet:
Actually, a better idea would be to provide a couple of functions to convert between CImg and OpenCV's Mat, so it would work pretty much like the current wrappers for flandmark from JavaCV's point of view. I think that'd be great and should be easy to do. Let me know when you guys have some time to work on that!

Issue #191, opened 2016-04-04
running caffe with cudnn 


mmanco:
I applied a [patch ](https://github.com/bytedeco/javacpp-presets/issues/187)which will allow building caffe with USE_CUDNN. The patch resolved the compilation errors but now when running I get an exception ```java.lang.UnsatisfiedLinkError: /tmp/javacpp681349592249244/libjnicaffe.so: /tmp/javacpp681349592249244/libcaffe.so.1.0.0-rc3: undefined symbol: _ZN5caffe21CuDNNConvolutionLayerIfED1Ev```

Any idea what can go wrong here ?

saudet:
```
$ c++filt _ZN5caffe21CuDNNConvolutionLayerIfED1Ev
caffe::CuDNNConvolutionLayer<float>::~CuDNNConvolutionLayer()
```
Looks like Caffe doesn't export the destructor for that... Sounds like a bug in Caffe?

mmanco:
We are running equivalent code in cpp against same version of caffe which was compiled aside of javacpp-presets and we don't see any issue. We will try to debug caffe compilation process in javacpp-presets to see maybe some parts are missed during that phase.

saudet:
You've made sure to apply explicitly "delete" on one of those objects in your C++ code and it works just fine? Strange... Maybe it's another problem that might arise if we have mismatched defines in the presets..

mmanco:
The issue resolved by deleting cppbuild dir under caffe preset. For some reason previous builds affected the results of new ones. Just wondering if mvn clean should delete the build working dirs ?

saudet:
The previous build affecting the current build, I'd file that as a bug upstream.

As for "mvn clean" calling "bash cppbuild.sh clean", sounds like a good idea! Could you send a PR for that? :)

mmanco:
@saudet Will do that with pleasure. Actually moving cppbuild under target will do that automatically without the need to call "bas cppbuild.sh clean"

PR #197, opened 2016-04-07
Hook maven clean phase with javacpp.sh clean

mmanco:
As agreed [here ](https://github.com/bytedeco/javacpp-presets/issues/191) maven clean phase should trigger cppbuild.sh clean

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -340,24 +340,42 @@
         <version>1.4.0</version>
         <executions>
           <execution>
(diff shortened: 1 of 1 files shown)


saudet:
Works well, thanks! I like the way we can use `-Djavacpp.cppbuild.skip` to skip the cleaning phase too. We can easily duplicate the old behavior that way. :)

Issue #225, opened 2016-05-20
ClassNotFoundException: org.bytedeco.javacpp.avutil$Pool_free_Pointer on 3.0.2-1.2 version


mtadmk:
When trying to run code from tutorial on line
`final AVFormatContext pFormatCtx = new AVFormatContext(null);`
class not found exception is being thrown:
```
 Exception in thread "Thread-5" java.lang.ExceptionInInitializerError
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at org.bytedeco.javacpp.Loader.load(Loader.java:464)
	at org.bytedeco.javacpp.Loader.load(Loader.java:409)
	at org.bytedeco.javacpp.avformat$AVFormatContext.<clinit>(avformat.java:2719)
	at Tutorial01.realStart(Tutorial01.java:126)
	at Tutorial01.lambda$startCamera$1(Tutorial01.java:321)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.ClassNotFoundException: org.bytedeco.javacpp.avutil$Pool_free_Pointer
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at org.bytedeco.javacpp.Loader.putMemberOffset(Loader.java:749)
	at java.lang.ClassLoader$NativeLibrary.load(Native Method)
	at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1938)
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1821)
	at java.lang.Runtime.load0(Runtime.java:809)
	at java.lang.System.load(System.java:1086)
	at org.bytedeco.javacpp.Loader.loadLibrary(Loader.java:637)
	at org.bytedeco.javacpp.Loader.load(Loader.java:492)
	at org.bytedeco.javacpp.Loader.load(Loader.java:409)
	at org.bytedeco.javacpp.avutil.<clinit>(avutil.java:10)
	... 8 more

Error putting member offsets for class org/bytedeco/javacpp/avutil$Pool_free_Pointer. 
```

[...]

Issue #234, opened 2016-06-03
Missing jnimxnet.dll in windows-x86_64.jar coming from the maven repository


Neiko2002:
Congratulation to the 1.2 release. I've read the changelog and was happy to find the new MXNet preset. Running the ImageClassificationPredict example on OSX was straightforward as usual, but when I tried it on windows I got the following exception:
UnsatisfiedLinkError: no jnimxnet in java.library.path

The windows-x86_64.jar coming from Maven seems to be empty:
[http://search.maven.org/remotecontent?filepath=org/bytedeco/javacpp-presets/mxnet/master-1.2/mxnet-master-1.2-windows-x86_64.jar](url)

saudet:
Yep, it's missing. Would you like to contribute the bit that is missing from the cppbuild.sh script?

Neiko2002:
I'm working on it right now. The first strange thing I found was after opening the "VS2013 x86 Native Tools Command Prompt" and calling "mvn clean install" in the javacpp directory, I got two Java Out-of-Memory-Exceptions for the testLongIndexer and the testDoubleIndexer test. Removing them solved the problem and resulted in a successful build so far.

Neiko2002:
Next thing, in the "Build Environments" Wiki you might want to add "tar", "make" and "patch" to the tools needed to build some of the presets:
pacman -S base-devel tar patch make ...

saudet:
Testing JavaCPP on 32-bit platforms will results in errors like that, but very few developers don't have access to 64-bit machines anyway, so I'm not sure it's something worth worrying about.

Neiko2002:
This also happens on a 64-bit windows platform using the "VS2013 x64 Native Tools Command Prompt" &#10132; "c:\msys64\mingw64_shell.bat" command line. But like I said the problem only occurs with javacpp 1.2.1, the current 1.2 is fine.

Neiko2002:
The commit c731d1d worked. I do get the OutOfMemoryErrors but the build does not crash because of it.  But there is a potential other problem. For all test*Pointer tests a get the same error:

```
testIntPointer(org.bytedeco.javacpp.PointerTest)  Time elapsed: 0.006 sec  <<< FAILURE!
java.lang.AssertionError: expected:<7622623231> but was:<7622623232>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
[...]

Issue #112, opened 2015-11-12
Feature Request: MKL Presets


bashimao:
I would like to see presets for Intel MKL. Specifically parts that are not covered by netlib-java. For example the Vector Statistical Library (VSL) (chapter 9 in MKL documentation, https://software.intel.com/en-us/articles/intel-math-kernel-library-documentation). I know MKL is vast. But it would definitely help a lot of people to have Intel's arsenal available in Java.

saudet:
I'm adding presets for OpenBLAS that indirectly link with MKL when available. Please check it out and let me know what you think! Thanks

saudet:
I've finally added presets for the whole of MKL: https://github.com/bytedeco/javacpp-presets/commit/190fcfc6a728874a423161b87619e40df9418282  Enjoy!

saudet:
And now it has been released!
http://search.maven.org/#artifactdetails%7Corg.bytedeco.javacpp-presets%7Cmkl%7C2017.3-1.3%7Cjar

PR #236, opened 2016-06-04
Windows support for the mxnet preset (issue #234)

Neiko2002:


saudet:
Could you make it so that it builds from source? 
Using existing binaries causes problems we can't fix: https://github.com/bytedeco/javacpp-presets/issues/225

Neiko2002:
It is possible, but on a normal windows system you don't have OpenBLAS installed and added to your PATH. Which means we would need to include it (plus its dependencies libgcc_s_seh-1.dll, libgfortran-3.dll, libquadmath-0.dll) to the resulting .jar file. Increasing the size by about 40MB. 

For windows it would be good to use the existing/installed libraries and warn the user it they are not configured correctly.

saudet:
OpenBLAS comes with MSYS2. Just call 
```
pacman -S mingw-w64-x86_64-openblas mingw-w64-i686-openblas`
```

saudet:
And if we them to the `link` or `preload` value of the `@Platform` annotation, the build process of JavaCPP will bundle them in the JAR file, so no library needs to be available at runtime on the machine.

saudet:
But it shouldn't increase the size by more than 10 megs though. Where do you get 40 megs from?

Neiko2002:
I have installed the OpenBLAS binary and the libopenblas.dll inside is 38MB.
https://sourceforge.net/projects/openblas/files/v0.2.14/

Getting OpenBLAS via MSYS2 downloads a 33MB libopenblas.dll. If we bundle all the necessary binaries together it will result in a >40MB MXNet jar file. We will need OpenBLAS for other libraries as well, so I was thinking that there might be another way to distribute the 3rd party libraries.

saudet:
Right, but compressed it's more like 5 megs:
http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-openblas-0.2.17-1-any.pkg.tar.xz

saudet:
Sure, we could make presets for OpenBLAS. After all, it gets used by Caffe, ND4J, and probably others as well. BLAS is an API implemented by other libraries though, so we might want to figure out how to make binaries compatible with MKL as well: https://github.com/bytedeco/javacpp-presets/issues/112

saudet:
Build instructions for Windows are available here:
https://github.com/dmlc/mxnet/blob/master/docs/how_to/build.md#building-on-windows
Could you translate that into a bash script? That's what I'd like to add to the cppbuild.sh script.

Neiko2002:
The next 2-3 weeks are full or work, but I think within the next month it should be ready.

saudet:
Duplicate of #309. Please don't create duplicate pull requests. Simply update the pull request next time!! Thank you

PR #247, opened 2016-07-01
Removing unnecessary OpenCV modules

Neiko2002:
The build process of OpenCV created too many modules. These interfere in other builds (like mxnet).


saudet:
Thanks! Please do a `git rebase` instead of a `git merge` to update your branch next time though

Issue #250, opened 2016-07-11
cudnnGetPoolingNdForwardOutputDim missing from cudnn bindings


cnuernber:
We are getting issues where our convolution,pooling layers have different calculated sizes from what cudnn calculates.  I would like to have runtime checks to make sure things line up but this function appears to be missing from the cudnn bindings.

saudet:
You mean the strides? There are functions to specify custom strides. In any
case, could you point out what is missing exactly?

cnuernber:
The specific function named in the title and the fix is in the linked pull request.

saudet:
Yes, saw that after. Sorry :) Thanks a lot for fixing this, will merge

cnuernber:
Totally understand.  I spoke to the cudnn guys and this fix isn't absolutely necessary:

https://devtalk.nvidia.com/default/topic/949999/cuda-programming-and-performance/cudnn-calculates-layer-sizes-different-than-caffe/

Thus you *have* to check convolution sizes but you do *not* have to check pooling sizes.

saudet:
Thanks for the fix! It is now included in version 1.3.

PR #251, opened 2016-07-11
Removing functions skipping the output dim queries

cnuernber:
https://github.com/bytedeco/javacpp-presets/issues/250


saudet:
Those functions used to be undefined, so that's why we had to skip them, but now they seem available.

PR #270, opened 2016-08-14
Add armhf to builds and fix for building on sdt4docker image

vb216:


saudet:
What happened?

vb216:
Ah I guess it closed as I rebaselined my fork.. it was way out of sync. It's actually pretty close, but is there something up with openblas? I've added linux-armhf build in there and cppbuild is fine, but mvn step complains that linePatterns can't be found in javacpp..? I've tried a fresh checkout of javacpp+presets on a linux machine too, same problem. Strange as I can see the function there in the latest version

Apart from that, should be ready to go soon

saudet:
The versions in the pom.xml files don't match... I plan to rectify this on the next release of JavaCPP next week I hope :) And then start using SNAPSHOT more properly

PR #272, opened 2016-08-18
tensorflow: bind DotGraph

Mistobaan:
This is a first attempt at getting more familiar with javacpp and the tensorflow bindings. Not sure if is the best way to map this simple func + struct, but it compiles and works .

### DIFF
--- a/tensorflow/src/main/java/org/bytedeco/javacpp/presets/tensorflow.java
+++ b/tensorflow/src/main/java/org/bytedeco/javacpp/presets/tensorflow.java
@@ -27,6 +27,8 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+
+import org.bytedeco.javacpp.BytePointer;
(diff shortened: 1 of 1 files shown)


saudet:
Yes, I think that's pretty much the best we can do. Good job!

The ordering and formatting of the new code could be a bit more consistent with the existing code, but it's not a big deal.

Let me know if you intend to work more on this or not! Thanks

Mistobaan:
sure I can fix the formatting. Is there a defined format for the project that I can just run and auto formats it ? (like `go fmt`).

I am not able to "subclass" those Function Pointers to actually implement a Java-struct to pass to inside the C++ Dot function. Even adding `@Virtual` in the code. Maybe I am doing it wrong. How would you subclass it properly ?

saudet:
No, don't have a formatter for that. Like I said, it's still alright the way it is now :)

When you say you can't subclass them, what problems are you encountering? Simply overriding a `call()` method and letting it use the protected constructor should work. What is the expected behavior vs the observed one?

saudet:
In any case, it builds fine for me, so let's merge this in. If you still have problems getting callbacks to work, let me know! Thanks

PR #273, opened 2016-08-22
New preset for Intel's librealsense version 1.9.6 .

poqudrof:
Hello, 
I created a preset for librealsense and it was not so painful (much easier than libfreenect2...) , even better it works. 

I also created a camera for JavaCV, which will have its own pull request soon.

It is a little bit Hacky for the patch, but it works nicely for the few features I tried:
- get the color image, 
- get the depth image
- get the extrinsics. 

Cheers, 
Jeremy.

### DIFF
--- a/cppbuild.sh
+++ b/cppbuild.sh
@@ -96,7 +96,7 @@ function download {
 }
 
 if [[ -z ${PROJECTS:-} ]]; then
-    PROJECTS=(opencv ffmpeg flycapture libdc1394 libfreenect videoinput artoolkitplus chilitags flandmark hdf5 openblas fftw gsl llvm leptonica tesseract caffe cuda mxnet tensorflow)
+    PROJECTS=(opencv ffmpeg flycapture libdc1394 libfreenect librealsense videoinput artoolkitplus chilitags flandmark hdf5 openblas fftw gsl llvm leptonica tesseract caffe cuda mxnet tensorflow)
 fi
 
 for PROJECT in ${PROJECTS[@]}; do
--- a/librealsense/README.md
+++ b/librealsense/README.md
@@ -1,0 +1,67 @@
+JavaCPP Presets for librealsense
+================================
+
+Introduction
+------------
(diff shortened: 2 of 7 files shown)


saudet:
Nice, thanks! Since only linux-x86_64 is supported though, could you add `value = "linux-x86_64"` to the `@Platform` annotation at https://github.com/poqudrof/javacpp-presets/blob/realsense-pull/librealsense/src/main/java/org/bytedeco/javacpp/presets/RealSense.java#L24 so it creates empty JARs on other platform instead of failing?

saudet:
Of course, we'll get it running on other platforms with time, but for now just having the build not fail on other platforms would be a good start. Sounds good?

poqudrof:
Yes, I will update this soon. I am between jobs now, I am rebuilding my development environment so I will add this as soon as possible. I guess it would work on any supported platform by librealsense (OSX, Windows, Linux).

saudet:
Hey, don't worry if it's not ready on all platforms. As long as you specify in `@Platform(value=...` the platforms that you do support, that's fine. I'll merge that. Thanks!

poqudrof:
It this version is OK, I can create a new pull request with a clean history.

saudet:
GitHub has a little "squash and merge" feature now, so we won't need to make clean pull requests anymore :) If you're satisfied with what you see in "Files changed" tab, and it builds fine on my end (will check a bit later) I'll merge it! Thanks

saudet:
First thing: Please switch the parent pom version from "1.2" to "1.2.5-SNAPSHOT". Also, there are unnecessary spaces added to the parent pom.xml... Thanks!

poqudrof:
Ok, its done.

saudet:
Looks good. I've tested it here and everything builds fine! I'd merge it like that, but since I need to go to sleep here now, if you have the time to modify the README.md file with the SNAPSHOT version as well, please do. I'll merge it tomorrow either way. Thanks!

PR #279, opened 2016-09-04
linux-armhf additions

vb216:
Hopefully these fix the lot for arm-hf, seems to build OK with:
mvn install -Djavacpp.platform=linux-armhf -Dplatform=linux-armhf -Dplatform.compiler=arm-linux-gnueabihf-g++ -Djavacpp.platform.compiler=arm-linux-gnueabihf-g++

to build all the presets. Only bit to check, I modified cuda and took out the linux-arm - I'm not sure that would be used? Otherwise needs some way to not try and build it for linux-armhf?

### DIFF
--- a/caffe/src/main/java/org/bytedeco/javacpp/presets/caffe.java
+++ b/caffe/src/main/java/org/bytedeco/javacpp/presets/caffe.java
@@ -37,7 +37,7 @@ import org.bytedeco.javacpp.tools.InfoMapper;
  * @author Samuel Audet
  */
 @Properties(inherit = opencv_highgui.class, target = "org.bytedeco.javacpp.caffe", value = {
-    @Platform(value = {"linux", "macosx"}, define = {"NDEBUG", "CPU_ONLY", "SHARED_PTR_NAMESPACE boost", "USE_LEVELDB", "USE_LMDB", "USE_OPENCV"}, include = {"caffe/caffe.hpp",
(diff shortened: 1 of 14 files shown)


saudet:
You were not able to build for CUDA with their binaries for ARM? We might want to let them know...

vb216:
I thought since 6.5 it was only 64 bit?

saudet:
Oh yeah, looks like they dropped ARM. Ok, that's fine then.

saudet:
Specifying "linux_x86" will also disable "linux-x86_64" and "linux-x86_64". It should be "linux-x86", with a hyphen, not an underscore.

The idea with FFTW is to build both fftw3 and fftw3f, so if we build both in the cppbuild.sh script (using the --enable-float flag to get fftw3f), we shouldn't need to modify the presets.

Otherwise looks good! When you get sometime to fix the above, I'll be merging this. Thanks

vb216:
Ah sorry.. looks like a typo, should have spotted that one. Hopefully a CI system checking pull requests will pick that up ;) 

Fixed the FFTW one too, looks better with the enable float flag. 
Hope thats all OK now - I rebuilt and goes through fine, apart from OpenBLAS that we know about.

saudet:
Thanks! AFAIK, we need to call the build for FFTW twice to get both libraries. Just do like linux-x86, it should work :) And then we don't need to modify the presets

vb216:
Done! Rechecked it out + built fftw to check, looks OK. I guess it would have shown up if I did a clean rebuild on the last attempt, hopefully all good now.

saudet:
Yes, looks good now. Thanks!

PR #281, opened 2016-09-09
Added OpenCV contirbution Bioinspired package

Maurice-Betzel:
Added OpenCV contribution Bioinspired package, including a patch for CMake 3.5 crashing in cl2cpp.cmake #6370 . Tested on Windows x86_64 and RHEL x_86_64.
The patch was not needed for the Linux build.


saudet:
We're not building with OpenCL, so we shouldn't need to patch anything there... ?

In any case, please squash your commits into one! The easiest way is to `git reset` your fork to the last commit that isn't yours, and `git commit` again all your changes. Thanks a lot for the contribution!

Maurice-Betzel:
Hello Samuel, if i do not patch the compile process finishes with reporting failure for bioinspired.
Concerning git, will do.

Maurice-Betzel:
It on #282 now

saudet:
We don't need to create a new pull request to update it. Please refrain from duplicating pull requests like that!

Maurice-Betzel:
Thanks for your correction. I have put 'read a good book about git' on my agenda.

PR #308, opened 2016-10-24
remove hdf5 compile step for linux-armhf

vb216:
Same cross compile problem with hdf5 for linux-armhf in docker.. hope these changes take that out of the build but leave the rest of the linux* in there.

### DIFF
--- a/cppbuild.sh
+++ b/cppbuild.sh
@@ -97,7 +97,7 @@ function download {
 }
 
 if [[ -z ${PROJECTS:-} ]]; then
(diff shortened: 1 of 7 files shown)


saudet:
Yes, hard to support, that's alright, we can take it out, but please leave  `hdf5/src/main/java/org/bytedeco/javacpp/hdf5.java` as is!

saudet:
Please don't close and reopen duplicate pull requests. Simply update the pull request!!

saudet:
Duplicate of #310

Issue #298, opened 2016-10-03
Update Wiki Build Environments (VS 2015, MSYS2 changes)


wongjoel:
Following the wiki instructions on "Build Environments" (Windows section) no longer appears to work.

I attempted to build javacpp-presets according to the instructions at [https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments](https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments), but it failed to compile successfully, and many of the steps did not work as written (at least for me).

I had hoped to submit an update to the wiki myself, but I have so far failed to get a working build environment.

For the Windows section:
1. Link to Visual Studio Community 2013 now takes you to the download for Visual Studio Community 2015 (not sure if that causes any issues)
2. running "pacman -S base-devel tar patch make git unzip zip nasm yasm pkg-config mingw-w64-x86_64-cmake mingw-w64-x86_64-gcc mingw-w64-i686" in a MSYS2 shell will fail, claiming conflicting files (can be forced with --force flag, but not sure if that's the correct thing to do?)
3. c:\msys64\mingw64_shell.bat no longer exists with the current installer of msys2 (I believe the equivalent command for a new install is "C:\msys64\msys2_shell.cmd -mingw64")

currently I can run the bash cppbuild.sh script, but I get a cmake issue, and can't work out if the issue is with my Visual Studio install, or my MSYS2 install:
-- The CXX compiler identification is unknown
-- The C compiler identification is unknown

Sorry if that's too many things put into one issue, the main concern I have here is that there doesn't seem to be a working set of instructions on how configure a clean install of windows to be able to build javacpp-presets.

wongjoel:
Okay, so I found the cause of the cmake issue - when using Visual Studio Community 2015, the cppbuild.sh scripts need to be updated to say <"$CMAKE" -G "Visual Studio 14 2015 Win64"> instead of <"$CMAKE" -G "Visual Studio 12 2013">, otherwise cmake looks for the visual studio 2013 compiler, which isn't installed.

Still having issues running "mvn clean install -Djavacpp.platform=windows-x86_64", but hopefully if I get that working too, I can write up instructions on setting up a build environment on Windows 10, with Visual Studio 2015 and MSYS2

saudet:
Thanks for testing this out! If you could check that everything works with Visual Studio 2015, that would be great. If there is anything that doesn't work, we can still download old versions here:
https://www.visualstudio.com/vs/older-downloads/

[...]

PR #309, opened 2016-10-25
Build MXNet under windows 

Neiko2002:
Several fixes for the new javacpp-preset version 1.2.5. Many of them have been discussed here:
https://github.com/bytedeco/javacpp-presets/issues/298

Additionally there is Windows support for MXNet.


saudet:
A few other things:
1. Please get rid of the tabs and use spaces instead
2. There is a `return 0` that prevents the build from actually proceeding in `cppbuild.sh`
3. Could you use the files as downloaded in the `downloads` directory, the same as for the other platforms?

Then the native library builds, but "Nothing generated for C:\msys64\tmp\javacpp-presets\mxnet\target\classes\org\bytedeco\javacpp\jnimxnet.cpp". If I use the original presets then "LINK : fatal error LNK1181: cannot open input file 'mxnet.lib'", so there's still a bit to clean up, but once that's done, it looks like it's going to work, so it should be good to merge! Thanks

Neiko2002:
I will have a look at it tomorrow. Downloading archives from the master branch might not be a good idea. Sometime MXNet is not stable, whats why I added the the stable commit part. They plan to release a new version 0.9 which we could use later on. Until then it might be better to git clone a stable commit for all operating systems.

saudet:
Sure, please download those for other platforms as well, but do not use git: It doesn't always work well on Windows.

saudet:
FYI, it's possible to download a specific commit if that was your question, for example:
https://github.com/dmlc/mxnet/archive/a5aeb0c43028f41863a8148eb3ecf30e90caad1e.tar.gz

saudet:
FYI, I made some changes on the master branch that conflicts with your fork, so make sure to update it and resolve them!

Neiko2002:
There should not be any conflicts. Github is blaming openblas.java because I've changed line 40 by adding "link = "libopenblas"," while you have changed line 39 to include the new blas_extrax.h

saudet:
You could just reset and rebase your fork, that'll clear out everything for sure :)

Neiko2002:
How do I reset a pull request? If I delete my remote fork Neiko2002/javacpp-presets nothing will happen to this pull request. 

I could not find any help here either:
https://help.github.com/categories/collaborating-with-issues-and-pull-requests/

saudet:
Something like this: https://help.github.com/articles/syncing-a-fork/
But after fetching the upstream repo, we can also reset and rebase instead of merge...

Neiko2002:
When I do all of this, I have a local branch with your changes first, followed by the changes of this PR. But my remote master [javacpp-presets](https://github.com/Neiko2002/javacpp-presets) is still the same. I can not reverse the changes of the remote master and therefore not the changes in this PR.

I've pushed the rebased and updated local branch to another remote branch:
https://github.com/Neiko2002/javacpp-presets/tree/mxnet

saudet:
It's always possible to do `git reset` to whatever commit we wish.

saudet:
`git merge` should work fine as well. Anyway, feel free to create another pull request if you prefer.

saudet:
Hi, any news on this? The presets are now officially upgraded to Visual Studio 2015 as per #298, so this should go a lot more smoothly now. It would be great to have a build of MXNet for Windows!

In any case, I think it would be easiest to start a new PR from scratch. GitHub will probably do that anyway if we try to reset the branch.

Thanks for everything and let me know if you run into issues again. Thanks!

PR #313, opened 2016-11-07
Update patch and scripts to TensorFlow 0.11.0rc2

andreas-eberle:
I updated the Android patch and added some more environment variables to the cppbuild.sh script to make sure TensorFlow's configure runs without a user prompt. 

Unfortunately the TensorFlow guys decided that they won't include the improved `util/python/python_config.sh` in the 0.11.0 build. However, as the released version requires a user input, I created a patch replacing the `python_config.sh` file with the newer one from TensorFlow's master. 

I tested the build for android-arm and macos-x86_64 without CUDA.


saudet:
Awesome! Thanks

Issue #327, opened 2016-11-30
Adding support for opencv text module


beligum:
As mentioned in #319 , I'm working on support for the opencv text module, but I'm looking for some help because I'm new to creating new presets.

After activating the "-DBUILD_opencv_text=ON" flag in javacpp-presets/opencv/cppbuild.sh,
I added the following code under /javacpp-presets/opencv/src/main/java/org/bytedeco/javacpp/presets/opencv_text.java
https://gist.github.com/beligum/2146b7819612e99af4536be032849bc0

Trying to compile on Linux x64 (Ubuntu Trusty) using:
mvn compile --projects opencv -Djavacpp.platform=linux-x86_64 -Djavacpp.platform.dependency=false

Below is the error that's thrown. Can someone with more insight than me have a look at it?
```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/bram/Programs/javacpp/javacpp-presets/opencv/src/main/java/org/bytedeco/javacpp/opencv_text.java:[221,22] <identifier> expected
[ERROR] /home/bram/Programs/javacpp/javacpp-presets/opencv/src/main/java/org/bytedeco/javacpp/opencv_text.java:[221,82] <identifier> expected
[ERROR] /home/bram/Programs/javacpp/javacpp-presets/opencv/src/main/java/org/bytedeco/javacpp/opencv_text.java:[221,84] ';' expected
[ERROR] /home/bram/Programs/javacpp/javacpp-presets/opencv/src/main/java/org/bytedeco/javacpp/opencv_text.java:[221,90] <identifier> expected
[ERROR] /home/bram/Programs/javacpp/javacpp-presets/opencv/src/main/java/org/bytedeco/javacpp/opencv_text.java:[221,93] illegal start of type
[ERROR] /home/bram/Programs/javacpp/javacpp-presets/opencv/src/main/java/org/bytedeco/javacpp/opencv_text.java:[221,104] '(' expected
[ERROR] /home/bram/Programs/javacpp/javacpp-presets/opencv/src/main/java/org/bytedeco/javacpp/opencv_text.java:[465,58] <identifier> expected
[ERROR] /home/bram/Programs/javacpp/javacpp-presets/opencv/src/main/java/org/bytedeco/javacpp/opencv_text.java:[465,60] ';' expected
[...]

Issue #168, opened 2016-03-01
Add support for OpenCV text module 


manuelh9r:
In the OpenCV preset the text module (Scene Text Detection & Scene Text Recognition) is currently disabled.

Would it be possible to enable this module as well? As far as I know this module also has dependencies to Tesseract and Leptonica. How should these dependencies be represented in JavaCpp?

I would be happy to contribute to enable this module. However, I would need some guidance as I am lacking experience with JavaCpp. I already enabled the text module in cppbuild.sh and created a  src/main/java/org/bytedeco/javacpp/presets/opencv_text.java file with the following contents:

``` java
@Properties(inherit = {opencv_highgui.class, opencv_ml.class}, value = {
    @Platform(include = {"<opencv2/text/erfilter.hpp>", "<opencv2/text/ocr.hpp>", "<opencv2/text.hpp>"}, link = "opencv_text@.3.1"),
    @Platform(value = "windows", link = "opencv_text310")},
        target = "org.bytedeco.javacpp.opencv_text")
public class opencv_text implements InfoMapper {
    public void map(InfoMap infoMap) {
        infoMap.put(new Info("std::vector<cv::text::ERStat>").pointerTypes("ERStatVector").define());
        infoMap.put(new Info("std::vector<cv::Vec2i>").pointerTypes("Vec2iVector").define());
        infoMap.put(new Info("std::vector<double>").pointerTypes("DoubleVector").define());
    infoMap.put(new Info("std::deque<int>").pointerTypes("IntDeque").define());
    }
}
```

I get a lot of errors during the build of the generated jniopencv_text.cpp file, like the following:

```
/home/manuel/dev/sources/javacpp-presets/opencv/target/classes/org/bytedeco/javacpp/jniopencv_text.cpp:6509:19: error: expected ';' before 'adapter4'
/home/manuel/dev/sources/javacpp-presets/opencv/target/classes/org/bytedeco/javacpp/jniopencv_text.cpp:6526:76: error: 'adapter4' was not declared in this scope
         ptr->run(*ptr0, *ptr1, (std::string&)adapter2, ptr3, (std::string&)adapter4, adapter5, arg6);
                                                                            ^
/home/manuel/dev/sources/javacpp-presets/opencv/target/classes/org/bytedeco/javacpp/jniopencv_text.cpp:6539:26: error: 'adapter4' was not declared in this scope
     signed char* rptr4 = adapter4;
                          ^
[...]

PR #333, opened 2016-12-12
Adding support for opencv text module

beligum:
Changed the opencv build script to include the text module and added a new opencv_text preset.
Tested (only) on linux-x86_64 (Ubuntu Trusty).


saudet:
Thanks!! Fixes #327 and #168

PR #356, opened 2017-01-12
Add LiquidFun presets

floybix:
Hey,

Thanks for making this amazing tool. I've put together a preset for [LiquidFun](https://github.com/google/liquidfun), a 2D physics library. (incidentally, this is now a more complete replacement for the pure-java porting effort https://github.com/jbox2d/jbox2d )

The latest release is missing some stuff so I actually referred to a github commit... seems pretty stable though; their last commit was nearly 2 years ago.

I only have a mac, so just added support for that platform so far.


saudet:
Thanks for your contribution! Sorry for being a bit slow in checking this out. I've pushed support for Linux x86 and x84_64. Seems to work fine. :) Now, what should we do about the version? 1.1.0 is pretty old... For other libraries like that that don't maintain versions, instead of trying to change our version based on every commit in their master branch, I just use the version of the JavaCPP Presets + "master" as version. What do you think?

floybix:
Great!

On the versioning scheme, I'm not sure how it works if you call it "master":

* Option a. The download reference is a commit hash. When they push a new
commit, and we want to use it, we have to bump the version of
javacpp-presets.

* Option b. The download reference is to "master". When they push a new
commit we therefore automatically use it.

Option a seems bad because you'd have unnecessary new versions of all the
other presets.

Option b seems bad because what you get could depend on **when** you build
it.

On the other hand it wouldn't be great to use the liquidfun commit hash as
a version identifier, because then you can't tell the ordering - can't tell
which commit is newer.

How about this then: we version it with a date string 20160130 and make it
refer to the commit that was master's head at that date.

I might have misunderstood something.

saudet:
It looks like the accepted convention is a mix of both, for example:
https://ffmpeg.zeranoe.com/builds/
https://fedoraproject.org/wiki/Packaging:Versioning#Snapshot_packages

floybix:
Ah! That makes sense. Do you want me to make the change to yyyymmdd-hash or
will you?

btw I'll be travelling the next few days so might not be very responsive.

saudet:
Yes, yyyymmdd-hash sounds like the right thing to do. Please make the changes! Thanks

saudet:
Just add a patch in the directory here:
https://github.com/bytedeco/javacpp-presets/tree/master/ffmpeg
And apply it at build time from inside the `cppbuild.sh` script.

floybix:
Done; let me know if there's anything else to do.

saudet:
I added a couple of things missing, but looks good! Thanks

saudet:
FYI, thanks to @vb216, SNAPSHOT artifacts are now available here:
https://oss.sonatype.org/content/repositories/snapshots/org/bytedeco/javacpp-presets/liquidfun/
Let us know if you have any issues with them and thanks again for your contribution!

PR #379, opened 2017-02-16
Patch tensorflow files to use tensorflow 1.0.0

andreas-eberle:
@saudet: I updated the files and tested the build. On Android and MacOS, it worked. However, when building for Linux x86-64 on my Mac Book, I get the following error:

https://gist.github.com/andreas-eberle/15845e7b5e13158f8db7b2b89f5112ed

Do you have an idea what could be causing this?


andreas-eberle:
I tested it on Ubuntu and there it worked. So I think this can be merged.

saudet:
Great, thanks! Could you update the version in the README.md file and both `platform/pom.xml` files as well?

andreas-eberle:
@saudet: Is that ok?

PR #381, opened 2017-02-20
specify build host for 32 bit

vb216:


saudet:
I haven't experienced any problems without that option, but if that helps, we probably should be adding it for the other platforms as well: x86_64-linux, ppc64le-linux, x86_64-darwin, i686-w64-mingw32, x86_64-w64-mingw32 should work...

vb216:
Seemed to fix the build error here
http://javacv.hopto.org:8085/job/Presets-Linux32/28/console

I guess was pulling in x64 lib for some reason, as that would be present in
the same docker container.

Bit wary of making wholesale changes across the rest if theyre building ok

saudet:
What happened? We don't need this anymore?

vb216:
Ah, think it got lost when I rebased by git.. Will wrap it into the other PR

vb216:
Yeah actually, looks like this isn't needed, was a conflict with having a locally built libusb in the docker container.

Having said that, looks like there's a problem later on with the latest tensorflow, I've updated bazel which cured the first problem, but now seeing errors that look to do with 32 bit compile attempt. I'd ran from cmd line but re-running whole job, in theory it should fail with the same issue here in an hours time.. http://javacv.hopto.org:8085/job/Presets-Linux32/34/console

saudet:
@vb216 FYI, fixed that issue with TensorFlow: https://github.com/bytedeco/javacpp-presets/commit/715bbaec04ddd1fba47d1efe2b24ea753d5c4451

PR #388, opened 2017-03-02
armhf MMAL accelerator

emilio1986:
MMAL video acceleration for Rassbery Pi 3 armhf processors


saudet:
Great, thanks! @vb216 what do you think? Does it still work on old Raspberry Pi with software fallback?

emilio1986:
Not sure. Tested only on RPI 3 :)

vb216:
hopefully should get a chance this weekend to try on a pi zero as a test, tho not sure I call any MMAL stuff normally. is good to get optimisations like this in though!

vb216:
@emilio1986 would you be able to send over a simple test case that uses the mmal functionality? I'd just like to give it a proper test, I dug out a pi1 to add to the test hardware now

saudet:
BTW, does feature show up as an entirely separate codec? If so, then it probably does not interfere with the software codecs.

saudet:
I'm trying to cross compile FFmpeg with MMAL, but it's looking for a file named `mmal.h` and it's not anywhere in https://github.com/raspberrypi/tools. Where can we find that file?

saudet:
Ok, they are found in https://github.com/raspberrypi/userland. After executing `./buildme` and passing those paths to the configure scripts, I am able to build, but we get a dependency on `libmmal_core.so`, etc. Are those files present on almost all Raspberry Pis?

vb216:
Looks like it lives here /opt/vc/lib/libmmal_core.so, that's on a recent raspbian distro - that usually covers alot of the userbase.

PR #399, opened 2017-03-21
Annoy preset

pommedeterresautee:
Compile but issues in retrieving some results

Scala code for the builder
```scala
object AnnoyInstanceJavaCPP {

  def generateJNIClass(): Unit = {
    val c = classOf[annoy_preset]
    val builder = new Builder().classesOrPackages(c.getName)
    builder.outputDirectory("./src/main/java/").build
  }

  private def setFolder(): Unit = {
    System.setProperty("java.library.path", "./src/main/java/")
  }

  def compileJNI(): Unit = {
    setFolder()
    val c = classOf[annoy]
    val builder = new Builder().classesOrPackages(c.getName)
    builder.outputDirectory("./src/main/java/").compilerOptions("-ffast-math").build();
  }

  def bindJNI() {
    setFolder()
    val c = classOf[annoy]
    Loader.load(c)
  }
}
```

Scala code for the tests
```scala
package app.annoy

import load.annoy.AnnoyInstanceJavaCPP
import org.bytedeco.javacpp.annoy.AnnoyIndexAngular

import scala.util.Random

object TestJavaCpp extends App {

  //  AnnoyInstanceJavaCPP.generateJNIClass() // called only the first time to generate the JavaCpp Code
  AnnoyInstanceJavaCPP.compileJNI() // Compilation - not required each time

  AnnoyInstanceJavaCPP.bindJNI() // required at each launch

  val seed = 42
  Random.setSeed(seed)
  val numberOfVectors = 10000
  val dimensions = 30
  val trees = 30

  val repetitionTest = 10

  for (testSeries ← 1 to repetitionTest) {
    println(testSeries)

    val annoy3 = new AnnoyIndexAngular(dimensions)
    annoy3.set_seed(seed)
    val annoy4 = new AnnoyIndexAngular(dimensions)

    annoy4.set_seed(seed)
    val vectors: Array[(Array[Float], Int)] = Array.fill(numberOfVectors)(Array.fill(dimensions)(Random.nextGaussian().toFloat)).zipWithIndex

    vectors.foreach {
      case (vector, index) ⇒
        annoy3.add_item(index, vector)
        annoy4.add_item(index, vector)
    }

    annoy3.build(trees)
    annoy4.build(trees)

    assert(annoy3.get_n_items() == numberOfVectors)
    assert(annoy4.get_n_items() == numberOfVectors)

    //      }

    for (testRetrievalNumber ← 1 to 50) {
      val elementToTest = Random.nextInt(numberOfVectors)
      val r3 = Array.fill(dimensions)(-1f)
      annoy3.get_item(elementToTest, r3)
      val r4 = Array.fill(dimensions)(-1f)
      annoy4.get_item(elementToTest, r4)
      assert(r3.sameElements(vectors(elementToTest)._1))
      assert(r4.sameElements(vectors(elementToTest)._1))

      val r3b = Array.fill(100)(-1)
      val distances3 = Array.fill(100)(-1f)
      annoy3.get_nns_by_item(elementToTest, 100, -1, r3b, distances3)
      val r4b = Array.fill(100)(-1)
      val distances4 = Array.fill(100)(-1f)
      annoy4.get_nns_by_item(elementToTest, 100, -1, r4b, distances4)
      assert(r3b.sameElements(r4b))
      println("r -> " + r3b.toSeq)
      println("d -> " + distances3.toSeq)
      //        }
    }
  }
}

```


saudet:
Looks good, thanks! But there are a few things that need to be done before we can merge this:
1. Format the directory structure like other presets
2. Provide a pom.xml file for the build (instead of the Scala code above)
3. Move the example code into the README.md file and potentially into a subdirectory like this one:
https://github.com/bytedeco/javacpp-presets/tree/master/libfreenect2/example

Thanks!

pommedeterresautee:
@saudet as explained in gitter this code compile but won't work without modifying Annoy C++ code.
You said may be some std::vector thing may help...

saudet:
Well, it looks like we already have perfectly usable mappings for `get_nns_by_item()` and `get_nns_by_vector()` here already:
```java
public native void get_nns_by_item(int item, @Cast("size_t") long n, @Cast("size_t") long search_k, @StdVector IntPointer result, @StdVector FloatPointer distances);

public native void get_nns_by_vector(@Const FloatPointer w, @Cast("size_t") long n, @Cast("size_t") long search_k, @StdVector IntPointer result, @StdVector FloatPointer distances);
```
What doesn't work when you try to use those ones?

pommedeterresautee:
Both methods seem usable, the issue is... the results :-)
Let's say you want to retrieve the first 100 similar vectors to item 4:
```scala
      val r3b = Array.fill(100)(-1)
      val distances3 = Array.fill(100)(-1f)
      annoy3.get_nns_by_item(elementToTest, 100, -1, r3b, distances3)
```
After calling the function, both `r3b` and `distance3` are kept unchanged!
The reason is because of the way Annoy works with them.
If you follow the C++ code, these vectors are supposed to be loaded with the retrieved values at this point(https://github.com/spotify/annoy/blob/master/src/annoylib.h#L592-L596):
```
    for (size_t i = 0; i < p; i++) {
      if (distances)
        distances->push_back(D::normalized_distance(nns_dist[i].first));
      result->push_back(nns_dist[i].second);
    }
```
So the idea is that the method `push_back` is applied to both vectors `result` and `distance` (which were arrays in the Java world).
The values retrieved by Annoy are put **AFTER** the 100 `-1` used to init the Java array (which is transformed in a vector in the C++ code).
Just before finishing the C++ code, the vectors have a size of 200 (the first 100 `-1` values followed by the retrieved values).
When both vectors go back to Java as arrays, the array size is still 100, and their content is only the first 100 values used to init them (so `-1` and not the one retrieved by Annoy).
To work around this behavior, I have modified Annoy code, use empty vector to retrieve the values, then when the work is finished I copy the content of the new vectors in the one representing the Array from Java, it works, but this is not ok if we want to share a preset.

If it is not clear, please don't hesitate to request more information.

saudet:
It should work properly if we use the `Pointer` variants of the methods, for example:
```java
    IntPointer r3b = new IntPointer();
    FloatPointer distance3 = new FloatPointer();
    get_nns_by_item(elementToTest, 100, -1, r3b, distances3);
    r3b.get(...);
    distances3.get(...);
```

PR #400, opened 2017-03-24
Update for FlyCapture v.2.10.3.266.

jpsacha:


saudet:
Thanks! The problem with that though is they dropped support for PGRFlyCapture, which FlyCaptureFrameGrabber depends on:
https://github.com/bytedeco/javacv/blob/master/src/main/java/org/bytedeco/javacv/FlyCaptureFrameGrabber.java
There is also FlyCapture2FrameGrabber, but does it work just as well as FlyCaptureFrameGrabber?
https://github.com/bytedeco/javacv/blob/master/src/main/java/org/bytedeco/javacv/FlyCapture2FrameGrabber.java 
I do not currently have access to a FlyCapture camera, so I cannot test this myself at the moment...

jpsacha:
Do we need two of them? Do you know if there is there something unique in FlyCaptureFrameGrabber that should be ported to FlyCapture2FrameGrabber?

I will try to test FlyCapture2FrameGrabber with 2.10.3.266.

saudet:
We don't need 2 of them, but I'm under the impression that FlyCapture2FrameGrabber is missing some things. @poqudrof ?

jpsacha:
I tested FlyCapture2FrameGrabber it worked fine with 2.10.3.266 (first without modifications).
I also made changes to FlyCapture2FrameGrabber to make it similar to FlyCaptureFrameGrabber, see 
[javacv PR #655](https://github.com/bytedeco/javacv/pull/655)

Issue #402, opened 2017-03-27
OpenBLAS preset not working on Android


jadarve:
Hi, I am trying to run the openblas preset on Android. I asked about this in [StackOverflow][SOlink]. When I run the example code from https://github.com/bytedeco/javacpp-presets/tree/master/openblas in my [MainActivity.onCreate()][MainActivity] I get the following error:

```
java.lang.UnsatisfiedLinkError: No implementation found for int org.bytedeco.javacpp.openblas.LAPACKE_dgels(int, byte, int, int, int, double[], int, double[], int) (tried Java_org_bytedeco_javacpp_openblas_LAPACKE_1dgels and Java_org_bytedeco_javacpp_openblas_LAPACKE_1dgels__IBIII_3DI_3DI)
    at org.bytedeco.javacpp.openblas.LAPACKE_dgels(Native Method)
    at org.androidopenblas.ExampleDGELSrowmajor.runExample(MainActivity.java:80)
    at org.androidopenblas.MainActivity.onCreate(MainActivity.java:19)
    ...
```

My gradle dependencies are as follow:

```
    compile 'org.bytedeco.javacpp-presets:openblas:0.2.19-1.3'
    compile 'org.bytedeco.javacpp-presets:openblas:0.2.19-1.3:android-x86'
    compile 'org.bytedeco.javacpp-presets:openblas:0.2.19-1.3:android-arm'
```

A sample app illustrating the error is at https://github.com/jadarve/AndroidOpenBLAS

Am I missing something?

[SOlink]: http://stackoverflow.com/questions/43034915/using-javacpp-openblas-in-android
[MainActivity]: https://github.com/jadarve/AndroidOpenBLAS/blob/master/app/src/main/java/org/androidopenblas/MainActivity.java

saudet:
We need gfortran to build LAPACK. Unfortunately, the Android NDK does not have gfortran, so we would first need to build gfortran for Android. If you already know of a solution to this problem, please let us know.

jadarve:
Have you tried using the toolchain from https://github.com/buffer51/android-gfortran ? I can build the openblas library for platform android-x86 with fortran support.

I am trying to use the same toolchain to build the java-preset, but so far it does not work.

saudet:
I haven't tried, but do let me know if you get it working! Thanks

jadarve:
Hi, I got it working both for android-arm and android-x86 platforms :). Here are the steps

[...]

PR #405, opened 2017-04-01
Fixed LAPACK support for OpenBLAS in Android.

jadarve:
Patch to support LAPACK in Android as discussed in https://github.com/bytedeco/javacpp-presets/issues/402

These changes require to modify the Android NDK and replace the x86 and arm7 toolchains with those provided in https://github.com/buffer51/android-gfortran/releases


saudet:
Thanks! Could you move the build instructions to the wiki page for that though?
https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments#android-arm-and-x86

jadarve:
I moved the instructions to the wiki page.

saudet:
Thanks! Have you verified that all other presets also build fine with that
installed? We'll need to confirm that...

jadarve:
Oh, actually I have not verified if the other presets build with the modified toolchain. I will try and let you know.

saudet:
We now have Travis CI working thanks to @vb216. Could you make those toolchain adjustments in the build script at https://github.com/bytedeco/javacpp-presets/blob/master/ci/install-travis.sh#L127 ? We'll then be able to see easily that everything builds fine with those changes and finally merge this! Thanks a lot

jadarve:
I am not aware of any alternative.

duanshanchong:
I have also encountered this problem. Have you solved it？

saudet:
I've upgraded the presets for OpenBLAS 0.3.21 (see commit https://github.com/bytedeco/javacpp-presets/commit/7d9084fcb16404c35c0b144b541960eeafa7fcf8), which no longer requires a Fortran compiler to build LAPACK, so it's now available for all platforms, including Android, iOS, and Mac on ARM, so I think we can close this pull request. Please give it a try with the snapshots: http://bytedeco.org/builds/ Let me know if you encounter any issues with them though! Thank you

Issue #305, opened 2016-10-18
Missing binary libraries for LLVM presets on Windows


romildo:
When trying to run the [Fac](https://github.com/bytedeco/javacpp-presets/tree/master/llvm#sample-usage) example application on Windows 10, I am getting the error:

```
C:\bcc328\fac>mvn exec:java
[INFO] Scanning for projects...
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Building fac 1.2
[INFO] ------------------------------------------------------------------------
[INFO]
[INFO] --- exec-maven-plugin:1.5.0:java (default-cli) @ fac ---
[WARNING]
java.lang.reflect.InvocationTargetException
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.codehaus.mojo.exec.ExecJavaMojo$1.run(ExecJavaMojo.java:294)
        at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.UnsatisfiedLinkError: no jniLLVM in java.library.path
        at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
        at java.lang.Runtime.loadLibrary0(Runtime.java:870)
        at java.lang.System.loadLibrary(System.java:1122)
        at org.bytedeco.javacpp.Loader.loadLibrary(Loader.java:654)
        at org.bytedeco.javacpp.Loader.load(Loader.java:492)
        at org.bytedeco.javacpp.Loader.load(Loader.java:409)
        at org.bytedeco.javacpp.LLVM.<clinit>(LLVM.java:10)
        at Fac.main(Fac.java:10)
        ... 6 more
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 0.953 s
[INFO] Finished at: 2016-10-18T07:20:46-02:00
[INFO] Final Memory: 9M/155M
[INFO] ------------------------------------------------------------------------
[...]

PR #407, opened 2017-04-03
Added LLVM for Windows x86 and x86_64

Wolftein:
#305


saudet:
Thanks for the contribution! Unfortunately, I wasn't able to get it to build following the instructions here:
https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments#windows-x86-and-x86_64

First, I get an error when decompressing the archives:
```
tar: cfe-4.0.0.src/test/Driver/Inputs/basic_cross_linux_tree/usr/x86_64-unknown-linux-gnu/bin/ld: Cannot create symlink to ‘ld.gold’: No such file or directory
tar: cfe-4.0.0.src/test/Driver/Inputs/basic_cross_linux_tree/usr/i386-unknown-linux-gnu/bin/ld: Cannot create symlink to ‘ld.gold’: No such file or directory
```

Ignoring that, I get a lot of errors from CMake about missing files... Would you have any additional instructions to get this working?

Wolftein:
It looks like msys2 doesn't support creating symlink from a package file, can't find a solution to this, but a workaround would be copying all files from the windows LLVM folder or providing LLVM already compiled.

saudet:
Actually, MSYS2 and Windows 10 do support symbolic links. We have to uncomment `MSYS=winsymlinks:nativestrict` from `mingw64.ini`, and run it as "Administrator". Is this what you did? Or how did you get it to build?

Wolftein:
Already did that and still i'm unable to retrieve a symbolic link from a package file (Maybe is related to this?
 https://github.com/Alexpux/MSYS2-packages/issues/140).

I copied the include and lib folder (built in windows with the same command) into the llvm folder and then running maven.

saudet:
So, let's forget about MSYS2. How did you build it if not by running
cppbuild.sh in MSYS2?

Wolftein:
I removed the download, unpacking line from cppbuild.sh and copied everything from my llvm's windows folder. Then run cppbuild and maven.

saudet:
And where do these files in your llvm folder come from? Did you extract the archive with something else like 7-Zip?

Wolftein:
Yes, i extract them using 7-Zip

saudet:
Ah, I see. MSYS2 also comes with a version of 7-Zip. I wonder if that one also works for extracting that archive from the command line, from inside the script...

Wolftein:
What about instead of download llvm source code and compile it, provide the .lib files and then run javacpp? llvm takes 30+ minutes to compile (if using all cores) otherwise it takes more than 1hs.

saudet:
Sure, where can we get the precompiled DLL?

Wolftein:
I have both x86 and x86_64 llvm.lib precompiled, i could upload them to github and replace the script or i could send them to you, so you can upload them on your own repository

saudet:
Are you planning on building and maintaining all future versions of LLVM this way?

If not, let's make a script to build those!

saudet:
Marking as duplicate of #530

Issue #364, opened 2017-01-19
OpenCV for Java with GPU (CUDA) Support


archenroot:
Hi,

first of all thanks for this great contribution!

I am going to process 4 camera system with OpenCV on FullHD resolution with 60FPS in realtime and I am looking for CUDA support. I am on Gentoo system, the openCV library is media-libs/opencv-3.1.0-r7 and is build with CUDA and Java flags. I can even find opencv-310.jar file on my file system from OpenCV installation, but cannot find GPU related packages, so I think this is not supported in the moment.

Does this project handle this somehow?

I tried to compile current dev branch, but it is failing on:
[ERROR] Plugin org.bytedeco:javacpp:1.3.2-SNAPSHOT or one of its dependencies could not be resolved: Could not find artifact org.bytedeco:javacpp:jar:1.3.2-SNAPSHOT in prokyon-public (http://blackhole.archenroot.org:50001/nexus/content/groups/public/) -> [Help 1]

[This artefact seems to be missing in maven](https://search.maven.org/#search%7Cga%7C1%7Ca%3A%22javacpp%22).

So I wasn't able to compile the source.

archenroot:
Ok, update, I am build javacpp latest snapshot, so 1.3.2 version is available and I able to continue with javacpp-presets itself.

archenroot:
Update, I played a little with pom.xml, but finally got working opencv module alone from master branch.

I also got working the Stitching.java example and tried with --try_use_gpu yes switch on 2 images and it works, but I cannot see any kernel running on GPU core with nvidia-smi utility:
`watch -n 0.0001 nvidia-smi`

But I don't see anything executed on CUDA, any explanation? Is CUDA really used, or is there any other example which I try and confirm it?

Thanks guys for your hard work!

saudet:
Duplicate of https://github.com/bytedeco/javacv/issues/481

PR #416, opened 2017-05-05
Add initial OpenCV CUDA bindings

SamCarlberg:
Creates bindings for:

| Header | Class |
|---|---|
| `opencv2/core/cuda.hpp` | `opencv_cuda` |
| `opencv2/cudaimgproc.hpp` | `opencv_cudaimgproc` |
| `opencv2/cudafilters.hpp` | `opencv_cudafilters` |

Addresses (but doesn't solve) #364, https://github.com/bytedeco/javacv/issues/481


saudet:
Looking good thanks! There are few things to fix before we can merge this though:
- Since CUDA is available only on some platforms, we need to list them in the annotations. Something like the following should work:
```java
@Platform(value = {"linux-x86_64", "linux-arm64", "linux-ppc64le", "macosx-x86_64", "windows-x86_64"}, ...
```
- And so `-DWITH_CUDA=ON` cannot be used on other platforms in the `cppbuild.sh` script.
- The content of `opencv_photo` and `opencv_shape` changes, I wonder why...?

SamCarlberg:
I think `opencv_photo` and `opencv_shape` were changed because I removed the generated class for the forward declaration of `cv::cuda::Stream` in `opencv_core` I guess I could change those presets to inherit from `opencv_cuda` to get the real class declaration.

saudet:
I don't think that's it, all the other modules don't change. Anyway, I'll
look into that. Make the other changes and it'll be fine!

One other thing though, to reduce build times, we should disable modules we
don't use...

SamCarlberg:
I'm not too happy about stuff like this:

```java
public native void detectMultiScale(@ByVal Mat image,
                                    @ByVal Mat objects);
public native void detectMultiScale(@ByVal UMat image,
                                    @ByVal UMat objects);
public native void detectMultiScale(@ByVal GpuMat image,
                                    @ByVal GpuMat objects);
```

that should only accept `GpuMat` inputs like this:
```java
public native void detectMultiScale(@ByVal GpuMat image,
                                    @ByVal GpuMat objects);
```

But `InputArray` and `OutputArray` need to be aliased as `Mat` or `UMat` for some conversion code, as well as the methods needed for uploading and downloading mats in host memory to device memory (`GpuMat::upload` and `GpuMat::download`). What would be a good solution to this problem?

saudet:
We can remove "Mat" and "UMat" from the redefinition, or do we need them somewhere? If it's just one place or two, we can use Info.javaText() for those.

BTW, the build is failing, on Windows:

> Failed to execute JavaCPP Builder: Could not parse "opencv2\cudafilters.hpp": File does not exist

saudet:
Ouch, it looks like OpenCV builds for all these archs by default:
```
--     NVIDIA GPU arch:             20 30 35 37 50 52 60 61
```
Limiting that to say `30` will reduce compile time by 8. And we could also use CUDA_ARCH_PTX to create binaries that will still work on all architectures.

SamCarlberg:
> Limiting that to say 30 will reduce compile time by 8. And we could also use CUDA_ARCH_PTX to create binaries that will still work on all architectures.

That sounds good. I'll try tackling that later this week.

SamCarlberg:
21c4f4d only compiles the CUDA stuff for 3.0 + PTX. Everything _should_ still work, but I had some weird issues with double frees by the PTX JIT compiler. Full dump [here](https://gist.github.com/SamCarlberg/ae881f987153353571572f82c6e07031). I'm not sure how robust the PTX stuff is.

saudet:
Thanks! Now the build finishes on time for Linux, but still not for Mac :/

@vb216 is looking into stages, yes. It won't work for Windows with AppVeyor though :(

So, recent versions of CUDA have issues compiling older PTX versions? That's unfortunate...

SamCarlberg:
Looks like the mac build was cancelled because the log got too big from warnings about unused parameters and functions in the core and cudev modules (mostly from stubs as far as I can tell)

And it's a problem with the PTX JIT. Caused a double free and probably a bunch of other issues I'm randomly getting when the cuda module gets loaded.

saudet:
@vb216 What do you think? Should we merge this?

vb216:
Yeah fine with me. Nearly got nexus AWS instance working which would let us
rejig the jobs to make up some time too

saudet:
I see why some modules like `opencv_photo` change. It's because we need to update their dependencies. In the case of `opencv_photo`, it's easy. We just need to replace `opencv_imgproc.class` by `opencv_cudaimgproc.class`, but others like `opencv_stitching` depend on modules that are not currently mapped:
```
$ readelf -d libopencv_stitching.so.3.2 | grep libopencv_cuda
 0x0000000000000001 (NEEDED)             Shared library: [libopencv_cudafeatures2d.so.3.2]
 0x0000000000000001 (NEEDED)             Shared library: [libopencv_cudalegacy.so.3.2]
 0x0000000000000001 (NEEDED)             Shared library: [libopencv_cudawarping.so.3.2]
 0x0000000000000001 (NEEDED)             Shared library: [libopencv_cudaimgproc.so.3.2]
 0x0000000000000001 (NEEDED)             Shared library: [libopencv_cudafilters.so.3.2]
 0x0000000000000001 (NEEDED)             Shared library: [libopencv_cudaarithm.so.3.2]
```

We'll need to add those to the `preload` list so they get properly loaded...

saudet:
Not sure how I missed that, but when built with CUDA, it looks like all libraries become dependent on CUDA! Check the output of `ldd libopencv_core.so | grep cu`:
```
libopencv_cudev.so.3.2 => .../javacpp-presets/opencv/cppbuild/linux-x86_64/lib/libopencv_cudev.so.3.2 (0x00007fc1b5277000)
libcudart.so.8.0 => /usr/local/cuda/lib64/libcudart.so.8.0 (0x00007fc1b49bf000)
libnppc.so.8.0 => /usr/local/cuda/lib64/libnppc.so.8.0 (0x00007fc1b474f000)
libnppi.so.8.0 => /usr/local/cuda/lib64/libnppi.so.8.0 (0x00007fc1add2b000)
libnpps.so.8.0 => /usr/local/cuda/lib64/libnpps.so.8.0 (0x00007fc1ad335000)
libcufft.so.8.0 => /usr/local/cuda/lib64/libcufft.so.8.0 (0x00007fc1a44e6000)
```
We can't have that. How should we deal with this?

SamCarlberg:
@saudet From my understanding, most modules will compile with CUDA support if the flag is set. If OpenCV loads the CUDA library only when a CUDA function is called (ie not at startup), like how Java does it, then I think it's okay. But if not, I think there would have to be separate builds for CUDA and non-CUDA versions

saudet:
All the libraries link statically with CUDA, so users would have to install it. It's ok for something like Caffe where most users use a GPU anyway, but that's not the case with OpenCV.

jcfrk101:
Has anyone decided how to proceed? A separate CUDA build or not?

saudet:
We'll need another build, yes. But to make releases, I'd like to be able to
use Travis CI and AppVeyor. Could you help with that?

saudet:
I've added an "extensions" feature to JavaCPP to allow us to have more than one set of binaries per platform: https://github.com/bytedeco/javacpp/commit/35463a65f93b52db5522a8dc8ec281289b6ac4be. We still need to modify the pom.xml files of the presets a bit, but it should work as follows:
1. Execute the build as usual, "mvn install"
2. Reexecute once more, but with the new "extension" plugin property set to say "-cuda". (This will build or pick up libraries in `cppbuild/<platform>-cuda` subdirectories.), say "mvn install -Djavacpp.extension=-cuda".

Furthermore, we should be able to create Maven profiles in the parent pom.xml to override the list of modules to only list the submodules that support say "cuda", and another list for say "avx512", etc.

At runtime, JavaCPP will look at the `@Platform(extensions={"-cuda", "-avx512", ...} ...` annotation value, and try to load them one by one, only trying the default one without extension name at the end if all else fails. The idea is that users wanting to use the "-cuda" one would have to put that artifact in their class path manually, if not only the default one would get added by the "-platform" artifact.

Thoughts?

/cc @akdeoras

SamCarlberg:
That's probably the best solution since OpenCV will break if built with CUDA and is loaded on a system that doesn't have CUDA support.

Side note, Travis has beta support for build stages now so it should be possible to make the mac builds not time out any more. https://docs.travis-ci.com/user/build-stages/

saudet:
I seem to recall there was still an issue with using stages and that's why we're not using them, @vb216 ? Also, it's not available on AppVeyor...

vb216:
Yeah problem is that there's no means of sharing output between build stages, there's issues open with Travis and Appveyor on this as alot of people are keen on that functionality.

Thinking of using S3 to upload build output to, making it available to dependencies and then splitting out the long opencv chain.

saudet:
Thanks for the contribution @SamCarlberg! Let me know if you encounter any problems using those functions.

PR #426, opened 2017-05-25
update travis and appveyor

vb216:
- split out install steps from travis.yaml file to separate script
- add android and linux32 projects
- bring build order inline with pom.xml (opencv and dependencies done in one go though)
- switch to google drive for download dependencies

Think that's most of it. There's still a few projects failing though, if you've got any views on those would be good to add them into the PR too


saudet:
Looking nice, thanks! Looks like MXNet broke again, TensorFlow is known not to build on Android for now (/cc @andreas-eberle), Skia on linux-x86 (/cc @nornagon), and I'm not sure why MKL doesn't link on Mac, but those are not blockers for this PR.

Some 32-bit libraries are missing from the image for libfreenect2 though:
https://travis-ci.org/bytedeco/javacpp-presets/jobs/236044004#L1181
Could you check into just that one? I think we're good to merge after we figure this out. Thanks!

vb216:
I think the libfreenect2 issue, is that there doesn't seem to be a glfw .i686 available? I've double checked and can't see any sign of it. Wondered how you've got around this for previous builds?

saudet:
Hum, you're right. I guess I just built it from source. It's not a big complicated library so maybe we should just add it to the cppbuild.sh file. Well, something else to put on the to-do list :)

Let me know when this is ready to merge! Thanks

vb216:
Ready from my side.. the to-do list was getting too long so I've just quickly added in the glfw build for linux32, worked OK in centos docker test I just ran and should turn another test green :)

saudet:
Great! Thanks. To be clear, I'm not expecting you to work on that to-do list. Just one thing, could you include in the header of gDownload.py where you got that function? And let me know if you have anything else to modify for this PR. If not, let's merge this!

vb216:
Ah yeah my bad, forgot to add the source for that file. Added now.

Nothing more to modify in the PR for now, I've started looking at adding linux-armhf, and I think the travis file could be simplified alot more by using their matrix setup properly, but I think this is a good point to get a bunch of updates in

PR #433, opened 2017-06-16
Include arm-hf in travis builds and deploy platform artefacts

### LINKED ISSUE #430: Upload *-platform packages to sonartype
At the moment the *-platform artefacts aren't uploaded to sonatype.

@vb216:  @saudet said you would be the right one for this.

vb216:
Various tidy ups (remove use of .m2 cache to reduce cache size, fix project order, remove building of javacpp per project and use snapshot instead), includes arm-hf in Travis build, and should upload platform artefacts (fixes #430 )


vb216:
Switched to use the new stages feature in Travis now, so each OS-platform is built as a stage before moving onto the next. Could potentially be used to to a platform step at the end of each OS target, or add in any pre/post steps easily now too.

vb216:
FYI this isn't all building successfully yet.. need to check how to allow failures but mark overall job as a fail

saudet:
Looks great! Thanks! Let me know when you feel you're done for this PR.

Just one thing, could we put the following in the `pom.xml` instead? I really hate it when Maven suddenly starts overwriting things in my repo cache. Then, we just need to call `mvn` with the `-U` option everywhere to force a check of the SNAPSHOT repositories, when we want.

```xml
  <repositories>
    <repository>
      <id>sonatype-nexus-snapshots</id>
      <name>Sonatype Nexus Snapshots</name>
      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
      <releases>
        <enabled>false</enabled>
      </releases>
      <snapshots>
        <enabled>true</enabled>
        <updatePolicy>never</updatePolicy>
      </snapshots>
    </repository>
  </repositories>
  <pluginRepositories>
    <pluginRepository>
      <id>sonatype-nexus-snapshots</id>
      <name>Sonatype Nexus Snapshots</name>
      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
      <releases>
        <enabled>false</enabled>
      </releases>
      <snapshots>
        <enabled>true</enabled>
        <updatePolicy>never</updatePolicy>
      </snapshots>
    </pluginRepository>
  </pluginRepositories>
```

vb216:
Ah yeah that makes sense. That change is in now. That's all done from my side and ready to go. There's an issue outstanding with flycapture now though.. the armhf is trying to use the latest SDK, and fails on not finding MultiSyncLibraryPlatform.h, it doesn't look like this is included in the new SDK though? The other platforms are probably building OK because they're using an older version.

saudet:
Thanks! And we should really be using `mvn -U` everywhere to make sure we always get the latest versions of stuff. 

As for FlyCapture, thanks for the note! I'll check and fix what's needed.

saudet:
Looks like the "MultiSync" feature of FlyCapture isn't available on ARM. I fixed the build in commit https://github.com/bytedeco/javacpp-presets/commit/cd82725e7c9c3e7d6e89dd7b15b2884d64d55c96 . Thanks for pointing this out!

BTW, I also updated HDF5 in commit https://github.com/bytedeco/javacpp-presets/commit/c8172a4dc969a18d58f5d4ebf40501358e804a81 and the new version spits out a lot less warnings, which should help the long OpenCV->Caffe build finish properly.

saudet:
Maybe this could fix what is happening with AppVeyor? https://github.com/appveyor/ci/issues/1161

vb216:
Hmm.. in theory I think they're started as new sessions, as I fire off a new bash for each. But, it looks like this https://github.com/StephanTLavavej/mingw-distro/issues/20 might mean the previous instance hasn't finished. One to keep an eye on..

Issue #436, opened 2017-06-28
Accelerate BLAS Preset for macOS


ibinti:
I wonder if javacpp can bind macOS Frameworks. If so, please consider adding BLAS in Accelerate Framework.

ibinti:
It should be doable as the macOS frameworks are basically a bundle of header files and respective dynamic libraries. And Accelerate BLAS are written in C, not Objective-C/C++ or swift.
On my macOS 10.12.5, the headers and *.dylib are located in /System/Library/Frameworks/Accelerate.framework/versions/A/Frameworks/vecLib.framework/Versions/A

ibinti:
javacpp jni library should be written as framework in order to be more useful (at least to me).

saudet:
We can use JavaCPP as a code generator as part of Xcode, that's fine.

ibinti:
@saudet I am going to try it myself. Could you give me some more directions if possible?

saudet:
Are you interested only in BLAS? If so, check the presets/openblas.java
file. There are bits about MKL. Try to do the same for Accelerate.

ibinti:
BLAS and LAPACK presets seem working.
I checked BLAS function `cblas_dgemm` and the multiplication result is correct.
There is no issue finding jni library if I put javacpp-presets jars in the lib.
But if I try to use from mavenLocal, then it complains like:
```
java.lang.UnsatisfiedLinkError: no jniblas in java.library.path
        at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
        at java.lang.Runtime.loadLibrary0(Runtime.java:870)
        at java.lang.System.loadLibrary(System.java:1122)
        at org.bytedeco.javacpp.Loader.loadLibrary(Loader.java:976)
        at org.bytedeco.javacpp.Loader.load(Loader.java:777)
        at org.bytedeco.javacpp.Loader.load(Loader.java:684)
        at org.bytedeco.javacpp.accelerate.<clinit>(accelerate.java:10)
        at ExampleDGEMMrowmajor.main(ExampleDGEMMrowmajor.java:118)
```
What could be the reason?

saudet:
Well then it's obviously not in the class path.

ibinti:
[...]

PR #444, opened 2017-07-16
Accelerate BLAS

ibinti:
Address the issue #436


ibinti:
Could you give me a hint how I can clean unrelated files?

saudet:
What do you mean? Just remove them.

saudet:
I see you're only mapping BLAS and LAPACK. Could you make this work with the presets for OpenBLAS instead?

saudet:
BTW, I have a better idea on how to go about this. We should create a new generic presets for "netlib", which would be able to bind not only to Accelerate, OpenBLAS, or MKL but to others like AMD BLAS as well, in a fashion similar to the now defunct https://github.com/fommil/netlib-java project. Would you like to work on that?

saudet:
I've made the necessary modifications in the commit above to be able to use Accelerate, etc with the bindings for OpenBLAS, so no need of separate presets for netlib for now. Instructions for Mac OS X at
https://github.com/bytedeco/javacpp-presets/tree/master/openblas#documentation

And thanks to @johanvos, when pull #515 gets merged, we should be able to do the same with iOS!

PR #456, opened 2017-08-30
Convert all known FFmpeg channel layouts into long

### LINKED ISSUE #455: AV_CH_LAYOUT_ defines are int instead of long
They're weakly typed in C relying on the integer type auto conversion (and this applies to Java as well).

However, inside Kotlin this creates a problem as conversion must be explicit and it gives you a bad feeling to manually cast a constant.

ishitatsuyuki:
Fix #455

It's very sad that I couldn't use glob for this. I ended up using grep and awk creating the list, but I hope this would be improved in future.


saudet:
Adding support for regex patterns there might not be a bad idea indeed.

PR #458, opened 2017-08-30
Min build 1.2 svc2

romaldg:


saudet:
Thanks for the patch! Could you wrap that up together into one cppbuild.sh script file though?

PR #463, opened 2017-09-06
Ffmpeg with rtmpdump

atiqmumtaz507:
We need RTMPS. I tried to compile it on my MAC and it works, but i need files from Gradle,


saudet:
Could we try to get it to work with OpenSSL 1.1.0 as per @pascalbaljet? BTW, is it known to work better with GnuTLS? If so, we could switch to that as well since it would make things clearer license-wise. @mifritscher might be interested in working on that too.

webdeutschland:
@pascalbaljet can you create a pull request ?? I need RTMPS Solution badly. please help !

saudet:
Could you please resolve the conflicts with the master branch? Thanks!

Issue #469, opened 2017-09-29
Trouble using presets/cuda-8.0-6.0-1.3 from clojure


cnuernber:
When using 8.0-1.2 version of presets/cuda and running lein deps I see it downloads the platform-specific bindings from maven-central:

Retrieving org/bytedeco/javacpp-presets/cuda/8.0-1.2/cuda-8.0-1.2-linux-x86_64.jar from central

I cannot seem to find these same bindings on maven central when running using lein repl if I upgrade my dependency to 8.0-6.0-1.3.  I was wondering if I am missing a define on the command line or of the platform-specific-binaries will no longer be uploaded to maven central or if this is some clojure-specific issue that I need to sort through.

cnuernber:
I found the fix is to use the platform dependency instead of the root dependency.  So concretely:

didn't work:
:dependencies [[org.bytedeco.javacpp-presets/cuda "8.0-6.0-1.3"]]

did work:
:dependencies [[org.bytedeco.javacpp-presets/cuda-platform "8.0-6.0-1.3"]]

saudet:
Yes, that should be clear from the README.md file or would you have
something to propose?

saudet:
Maybe we could add the necessary lines for Leiningen in the README.md?

cnuernber:
https://github.com/bytedeco/javacpp-presets/pull/470

saudet:
Closing as duplicate of pull #470. Let's get the pull request in shape for a merge! Thanks

PR #470, opened 2017-09-30
Updating readme to include clojure example.

cnuernber:
https://github.com/bytedeco/javacpp-presets/issues/469#issuecomment-333260746

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,12 @@ We can also have everything downloaded and installed automatically with:
 
 * leiningen project.clj (cuda 8.0, cudnn 6.0, javacpp 1.3)
 ```clojure
+(def module-name "cuda")
+(def module-version "8.0-6.0")
+
 (defproject myproject "0.1.0-SNAPSHOT"
   :dependencies [[org.clojure/clojure "1.9.0-alpha17"]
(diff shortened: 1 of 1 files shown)


saudet:
Thanks! Could you make this more generic like with Gradle or sbt? Or if the idea is to provide a sample for CUDA, maybe https://github.com/bytedeco/sample-projects/ would be a better place for this?

saudet:
Thanks, but I meant, could you use the same variables names as the snippet for the other tools? And if you'd like to have an example that mentions "cuda" explicitly, please send a pull request https://github.com/bytedeco/sample-projects/. Thank you!

saudet:
I've added instructions for Leiningen, including JavaCPP and JavaCV, in the following commits: https://github.com/bytedeco/javacpp/commit/96045c9b86625df009e3986c555e467bd7d58853 https://github.com/bytedeco/javacv/commit/726dc3df26c8ada6e4f0c75e0f4be85ce07c090c https://github.com/bytedeco/javacpp-presets/commit/a3484bce7d01402ed639e5911a8cd0e184dc2d25

Thanks for the contribution!

PR #471, opened 2017-09-30
ppc cross compile

vb216:
Slowly getting there.. ffmpeg, cuda, and leptonica outstanding. 

ffmpeg - not picking up the earlier built libs, i've tried a few different angles but not cracked it yet
cuda - its using x86_64 cuda in the docker image, but when it comes to linking it cant find the ppc64 .so files. might be a case of just bundling them and manually copying in to the docker image
leptonica - a little similar to ffmpeg, not finding a zlib its happy with


saudet:
Thanks! Will look into it as I have time as well. 

For CUDA, we need to install the binaries from the repo of NVIDIA:
https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=ppc64le&target_distro=Ubuntu&target_version=1604&target_type=deblocal
You had that in your script, but have removed it. It doesn't want to install?

vb216:
Lost track a little! Yeah was in there for ppc64, but failed to install and then if I remember correctly was struggling for dependencies.

saudet:
Not sure this might help, but FYI:
https://developer.ibm.com/linuxonpower/advance-toolchain/advtool-compiler/

vb216:
Might give better performance if anyone wants to improve it down the road. In theory, all solveable just need to find a bit of time. I think for CUDA, might just zip up the ppc64 .so libs. Other two problems probably cracked in the same solution - getting config to pick up the built deps properly

vb216:
Hm, with cuda, its now failing due to not finding lcuda below. But, I can only see a libcuda.so in the stubs, has something changed in v9? Or am I missing something else. 

Compiling /home/travis/build/javacpp-presets/cuda/target/classes/org/bytedeco/javacpp/linux-ppc64le/libjnicublas.so
[INFO] powerpc64le-linux-gnu-g++ -I/usr/local/cuda-9.0/include/ -I/usr/lib/jvm/java-8-openjdk-amd64/include -I/usr/lib/jvm/java-8-openjdk-amd64/include/linux /home/travis/build/javacpp-presets/cuda/target/classes/org/bytedeco/javacpp/jnicublas.cpp -mcpu=power8 -m64 -O3 -s -Wl,-rpath,$ORIGIN/ -Wl,-z,noexecstack -Wl,-Bsymbolic -Wall -fPIC -shared -o libjnicublas.so -L/usr/local/cuda-9.0/lib64/ -Wl,-rpath,/usr/local/cuda-9.0/lib64/ -lcublas -lcuda -lcudart 
/usr/lib/gcc-cross/powerpc64le-linux-gnu/5/../../../../powerpc64le-linux-gnu/bin/ld: cannot find -lcuda

saudet:
It usually comes with the driver. It should be somewhere it the .deb, but
the stub might work as well.

vb216:
Hopefully this is it finished.

tesseract/cppbuild.sh
```diff
@@ -79,7 +79,17 @@ case $PLATFORM in
...
     linux-ppc64le)
         #patch -Np1 < ../../../tesseract-linux.patch
-        ./configure --prefix=$INSTALL_PATH CC="$OLDCC -m64" CXX="$OLDCXX -m64" LEPTONICA_CFLAGS="-I$LEPTONICA_PATH/include/leptonica/" LEPTONICA_LIBS="-L$LEPTONICA_PATH/lib/ -llept" CPPFLAGS="-I$LEPTONICA_PATH/include/" LDFLAGS="-L$LEPTONICA_PATH/lib/"
+        MACHINE_TYPE=$( uname -m )
+        if [[ "$MACHINE_TYPE" =~ ppc64 ]]; then
+          ./configure --prefix=$INSTALL_PATH CC="$OLDCC -m64" CXX="$OLDCXX -m64" LEPTONICA_CFLAGS="-I$LEPTONICA_PATH/include/leptonica/" LEPTONICA_LIBS="-L$LEPTONICA_PATH/lib/ -llept" CPPFLAGS="-I$LEPTONICA_PATH/include/" LDFLAGS="-L$LEPTONICA_PATH/lib/"
+        else
+          cd ..
+          rm -rf tesseract-3.05.01
+          git clone -q --branch=3.05 https://github.com/tesseract-ocr/tesseract.git tesseract-3.05.01
```
saudet: There's only been 18 commits between 3.05 and 3.05.01. We can probably just make a patch to revert the offending one...

saudet:
Could we put the content of `ci/hdf5` into a patch and in the main `hdf5` subdirectory itself?

ci/install-ppc.sh
```diff
@@ -0,0 +1,92 @@
...
+docker exec -ti $DOCKER_CONTAINER_ID /bin/bash -xec 'add-apt-repository "deb [arch=ppc64el] http://ports.ubuntu.com/ubuntu-ports xenial-security main restricted universe multiverse"' 
+docker exec -ti $DOCKER_CONTAINER_ID /bin/bash -xec "apt-get update"  
+docker exec -ti $DOCKER_CONTAINER_ID /bin/bash -xec "apt-get -y install libcairo2-dev:ppc64el libpango1.0-dev:ppc64el libglib2.0-dev:ppc64el libatk1.0-dev:ppc64el libfreetype6-dev:ppc64el libasound2-dev:ppc64el libperl-dev:ppc64el libqt4-dev:ppc64el libgtk2.0-dev:ppc64el libusb-dev:ppc64el zlib1g-dev:ppc64el gir1.2-atk-1.0:ppc64el gir1.2-gtk-2.0:ppc64el gir1.2-pango-1.0:ppc64el libgdk-pixbuf2.0-dev:ppc64el gir1.2-gdkpixbuf-2.0:ppc64el gir1.2-freedesktop:ppc64el gir1.2-glib-2.0:ppc64el  libgirepository-1.0-1:ppc64el libusb-1.0-0-dev:ppc64el libusb-dev:ppc64el libxcb1-dev:ppc64el libxcb1:ppc64el"  
+
+if [[ "$PROJ" =~ cuda ]]; then
+   echo "Setting up for cuda build"
+   #cp /usr/include/cudnn.h /usr/local/cuda/include/  
+   python $TRAVIS_BUILD_DIR/ci/gDownload.py 0B2xpvMUzviShdFlrT3FtNXFSRW8 $HOME/downloads/cudappc64.tar
```
saudet: How did you create that `cudappc64.tar`? Maybe we can just put the few commands for this here...

vb216:
Yeah I was just trading off time vs getting something out there. They're all possible, though the last one its quite a pain - its deconstructing the .deb file and repackaging it up, I guess could be put in a neat one liner but I didn't see anything immediate. Won't be able to do anything for a few weeks

saudet:
This is working remarkably well, great job! While fixing a couple of things, I also discovered an unrelated issue with Leptonica not linking to the image libraries, so that's good too.

Now that this is done, as far as CI goes, I think the priority would be to update the http://bytedeco.org/builds/ page with recent information about Travis CI and AppVeyor. Feel free to start sending pull requests there when you get some time in a few weeks, but no rush, take your time. Thanks!

PR #485, opened 2017-11-24
Intel QSV with ffmpeg

n-kai-cj:
I'd like to build `libmfx` into ffmpeg on windows platform.
`libmfx` enables Intel QSV codec and I've checked this pull request without any errors on windows-x86_64.
Can you merge it?


saudet:
Thanks! Would it be possible to add it to the other platforms as well?

n-kai-cj:
Yes, linux can be.
But sorry, I don't have any platforms except windows so I can't test it.

saudet:
That's alright, Travis CI will test it :) Thanks

saudet:
So it builds fine on Windows, please update for the other platforms! Thanks

saudet:
Could you try to add the following lines for the other platforms as well and see what happens?
```bash
        cd ../mfx_dispatch-$MFX_VERSION
        autoreconf -fiv
        PKG_CONFIG_PATH="../lib/pkgconfig" ./configure --prefix="`pwd`/.." --disable-shared --enable-static --enable-fast-install
        make
        make install
```

n-kai-cj:
I only have windows-x86_64 platform, so I can't try on other platforms.
But it seems `libmfx` cannot be built into ffmpeg on linux because official ffmpeg on linux doesn't contain `libmfx`.
Android and macos don't apply `libmfx` as well.
So I removed `--enable-libmfx` from ENABLE args and committed.

The only modification on windows can't be merged?

saudet:
We can build libmfx on other platforms as easily as on Windows. Just add the same kind of build commands as you did for Windows, and then we can check if there are any builds issues here:
https://travis-ci.org/bytedeco/javacpp-presets/pull_requests
You don't need to install anything to try the build on other platforms because Travis CI does it for you. Please give it a try!

n-kai-cj:
Really? That's a good news.
So I have to add your lines for `android-x86`, `linux-x86`, `linux-x86_64`, `macosx-x86_64`, `windows-x86`, and `windows-x86_64` right? Or add for all platforms?

saudet:
Yes, like that, thanks! It looks like we'll need to resolve the conflict before Travis CI runs your pull request though.

saudet:
Unfortunately, it fails on all platforms except Windows :(
https://travis-ci.org/bytedeco/javacpp-presets/builds/309838539
https://ci.appveyor.com/project/Bytedeco/javacpp-presets/build/289
If you could take a look at the error messages and try to fix this, it would be great!

n-kai-cj:
I knew it...
As I said `libmfx` is built into official ffmpeg for only windows.
It seems linux needs to install Intel Media Server Studio, and Intel QSV uses Core iX CPU, so arm or ppc processor cannot use Intel QSV.

Anyway, I think it cannot be helped.
The only modification on windows can't be merged?

saudet:
No, the errors are just normal build errors. We don't need any special
hardware or software...

saudet:
But it looks like it wouldn't do anything on other platforms than linux-x86 and linux-x86_64. So let's get it working on at least those 2?

PR #487, opened 2017-11-29
Enabled AMR-WB encoding support to FFmpeg

canelzio:
- Added commands for download, untar and install vo-amrwbenc library
- Enabled AMR-WB encoder in FFmpeg using --enable-libvo-amrwbenc parameter


saudet:
Your additions are passing with flying colors on all platforms! Awesome! Thanks a lot

canelzio:
Thanks to you for support!

Issue #389, opened 2017-03-07
Hardware Accell in ffmpeg


CraziFuzzy:
Is there a technical or licensing reason that would prevent enabling the various hardware accelerated codecs in the ffmpeg.jar?

saudet:
If we can build them easily and FFmpeg can fall back on software when the hardware is missing, that is fine. Check pull #388, for example, this is one being investigate for ARM at the moment.

CraziFuzzy:
Hmmm.  I'll have to investigate that. I don't believe ffmpeg ever will fallback to using a codec that is not specified.  nvenc_h264, h264_qsv, and h264 are all completely separate encoders.  It would probably be up to the application to make sure it's requesting the right encoder.

saudet:
If they show up as additional encoders, yeah, doesn't sound like that would
cause any issues.

CraziFuzzy:
Yeah, as far as I know, they are pretty independent from the existing software codecs.  What I don't know is how difficult it would be to build, I have always used pre-built binaries, on both windows and linux.  I don't know if it's as simple as enabling an additional switch, or if other 3rd party libraries need to be included as well.  The first source of info I know of on it is here: https://trac.ffmpeg.org/wiki/HWAccelIntro  Might make more sense to you than I.

saudet:
Ok, let's keep this issue open so others can check this out as well and weigh in. Thanks for the link!

saudet:
FYI, support for Intel QSV as been merged with pull https://github.com/bytedeco/javacpp-presets/pull/485.

saudet:
MMAL and OpenMAX acceleration has also been merged with pull #388 for ARM platforms. Is there any other supported APIs worth adding?

n-kai-cj:
I committed Nvidia NVENC as well.
As far as I know, there is no other hardware codecs in ffmpeg.
I think this issue can be closed.

saudet:
There are a couple of others like OpenCL, but I remember having issues with that one previously...

n-kai-cj:
As for OpenCL, it's currently only used for filtering, encoding/decoding is not supported for now.
https://trac.ffmpeg.org/wiki/HWAccelIntro

And official ffmpeg seems not to be built with `--enable-opencl` for windows platform.
Apparently I can't help this.

mifritscher:
[...]

PR #492, opened 2017-12-10
ffmpeg with nvenc

n-kai-cj:
I'd like to build `nvenc`, use NVIDIA GPU for encoding or decoding as GPGPU, into ffmpeg as well, so we need `CUDA` to use `nvenc`
To get ffmpeg with `nvenc`, all we have to do is adding `--enable-nvenc` to ffmpeg configure option.
I've checked the build made success on windows-x86_64 platform.

I think this commit relates with #389 .
Thank you.

### DIFF
--- a/ffmpeg/cppbuild.sh
+++ b/ffmpeg/cppbuild.sh
@@ -262,7 +262,7 @@ case $PLATFORM in
         fi
         cd ../ffmpeg-$FFMPEG_VERSION
         patch -Np1 < ../../../ffmpeg-linux.patch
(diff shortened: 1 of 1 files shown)


saudet:
Does this create a dependency on CUDA? Or does it link statically or do
something dynamically otherwise?

saudet:
No, it doesn't appear to add a dependency on CUDA, interesting.

n-kai-cj:
Sorry for late and thank you @saudet !
Actually I have checked only build success, but not checked whether work or not.

PR #515, opened 2018-01-29
Build OpenBlas and mkl for ios (x86_64 and aarch64)

johanvos:
With this PR, I can build the jars required for iOS deployment:

mvn -DskipTests -Dmaven.javadoc.skip=true -Djavacpp.platform=ios-x86_64  -Djavacpp.cppbuild.skip=true  install
mvn -DskipTests -Dmaven.javadoc.skip=true -Djavacpp.platform=ios-arm64  -Djavacpp.cppbuild.skip=true  install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,6 +89,27 @@ matrix:
         install: true
         script: ./ci/install-travis.sh
 
+      - os: osx
+        osx_image: xcode7.3
+        env: PROJ=openblas OS=ios-arm
+        install: true
+        script: ./ci/install-travis.sh
+      - os: osx
+        osx_image: xcode7.3
+        env: PROJ=openblas OS=ios-arm64
+        install: true
+        script: ./ci/install-travis.sh
+      - os: osx
+        osx_image: xcode7.3
+        env: PROJ=openblas OS=ios-x86
+        install: true
+        script: ./ci/install-travis.sh
+      - os: osx
+        osx_image: xcode7.3
+        env: PROJ=openblas OS=ios-x86_64
+        install: true
+        script: ./ci/install-travis.sh
+
       - os: linux
         env: PROJ=opencv,flandmark OS=linux-armhf
         install: true
--- a/openblas/cppbuild.sh
+++ b/openblas/cppbuild.sh
@@ -120,6 +120,16 @@ case $PLATFORM in
         export TARGET=HASWELL
         export NO_AVX2=1
         ;;
+    ios-x86_64)
+        export CC="clang++"
+        export FC="clang++"
+        export BINARY=64
+        ;;
+    ios-arm64)
+        export CC="clang++"
+        export FC="clang++"
+        export BINARY=64
+        ;;
     *)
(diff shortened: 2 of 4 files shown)


saudet:
As noted in pull #444 we might want to create new presets for "netlib" to be able to use implementations like Accelerate that have separate libraries for BLAS and LAPACK, but then again since Accelerate doesn't implement LAPACKE it might not matter as much:
https://stackoverflow.com/questions/38548113/lapacke-on-accelerate-framework

So, BLAS-only like on Android, that looks good enough for now. Let's make Accelerate optional though and give users the option of actually using OpenBLAS if they so choose. Did you encounter any problems building LAPACK-less OpenBLAS for iOS?

johanvos:
It didn't run out of the box but I can't remember if the issues would be easy to solve or not.
However, since accelerate is available as a Framework on iOS, it sounds like the path of least resistance.

saudet:
Ok, but let's put the steps to build it, even if it doesn't run for now,
and leave it up to the user at link time which one to use :)

And mkl isn't available on iOS for sure, so let's not add it to the pom.xml
file for iOS.

johanvos:
* added basic step to at least kick off the build of openblas
* removed mkl

saudet:
Ok, I've mode some changes to JavaCPP (https://github.com/bytedeco/javacpp/commit/944f6b1d99985a0b0c54b531716c8ce9966e4f58) and this PR (https://github.com/bytedeco/javacpp-presets/pull/515/commits/16ae0780aa64af53e7a55fa034144ff5c267ddd9) so it actually builds and bundles OpenBLAS now. That seems to work for me here. You'll find `libopenblas.a` in the JAR file, but you can ignore it (until if you find that Accelerate is buggy). It should still be able to link with Accelerate too.

Now this doesn't include LAPACKE, so the Java wrappers that this generates won't work with ND4J. The trick is to run a `mvn install` for another platform _after_ the ones for iOS, say "macosx-x86_64", so that the Java wrappers get overwritten by something that does map LAPACKE. This way, after that, we can build ND4J. At runtime, LAPACK functionality just won't be available on iOS (or Android).

saudet:
Ok, I spent a bit more time fixing the builds and now it should work. With the latest commit above, can you try to run *exactly* these commands inside the `openblas` subdirectory and let me know if everything appears to work properly for you as well?
```
mvn -U clean
mvn install -Djavacpp.platform=ios-arm
mvn install -Djavacpp.platform=ios-arm64
mvn install -Djavacpp.platform=ios-x86
mvn install -Djavacpp.platform=ios-x86_64
mvn install -Djavacpp.platform=macosx-x86_64
```

saudet:
Well, Travis CI seems to be happy with it, so let's get this merged and it will start deploying artifacts automatically! Let me know if you encounter any issues with them. Thanks

saudet:
Fixed final linking issues in commit https://github.com/bytedeco/javacpp-presets/commit/aec02659a3575e3d3762e48a23ab333466b004de. It should actually work now :)

PR #522, opened 2018-02-06
Tesseract 4.0.0

amalow:
Changes to Presets for Tesseract 4.0.0 with (LSTM), only required changes on StringReadCallback, CharReadCallback and IntReadCallback.

Notes: We could probably change git to pull from a specific commit of Tesseract. Unfortunately due to 4.0.0 being an Alpha on the project's master branch that's an issue we have to get around in-case of non-backwards compatible changes upstream.


saudet:
Thanks!! Looks like the build currently fails on all platforms though :(

amalow:
Bugger, thought we'd get it first go ;). I've had a quick skim through Travis' logs and it looks like something is going on between the automake and the configure steps. 

This is being thrown from the Configure step:
```
checking host system type... x86_64-apple-darwin15.6.0
./configure: line 4243: syntax error near unexpected token `-mavx,'
./configure: line 4243: `AX_CHECK_COMPILE_FLAG(-mavx, avx=true, avx=false)'
```

This is also a problem I run into on AWS instances when I try and build tesseract from source, I'll look into a fix for it.

amalow:
Alright, looks like we may need autoconf-archive as a dependency
see: [](https://github.com/tesseract-ocr/tesseract/issues/777

amalow:
May have to revert to an earlier version of autoconf available in tar.gz, tar.xv is didn't work for android-arm

saudet:
I'm pretty sure we don't need to build autoconf from source, but if we do, let's try to use the build files for CMake instead. They look a bit more sane.

amalow:
Yeah, automake doesn't look to have fixed it in the Mac build on travis either. I'll give CMake a try.

saudet:
Hi, any updates on this? Are you having issues with anything?

amalow:
Sorry I haven't had a chance to look at CMake yet.

saudet:
Looks like Tesseract 4.0.0-beta.1 has been released! Let's get the presets updated with that version.

saudet:
FYI, installing `autoconf-archive` allowed me to proceed with Autoconf: https://github.com/tesseract-ocr/tesseract/issues/777

PR #525, opened 2018-02-09
Skia on iOS

nornagon:
This builds, but I haven't tested the resulting binaries as I don't have a test project that runs Java on iOS.


nornagon:
@radxe could you pull this branch and try it out?

saudet:
Awesome, thanks! Could we do Android as well possibly?

nornagon:
Could do, but Android has a Skia library [built in](https://developer.android.com/reference/android/graphics/Canvas.html), so I don't know why someone would want that.

saudet:
Good point, probably not useful...

nornagon:
I guess it might be good if you're doing something cross-platform and don't
want to switch API or build an abstraction?

saudet:
Maybe... In any case, the new iOS builds seem to work fine, but they're breaking Linux and Mac:
https://travis-ci.org/bytedeco/javacpp-presets/builds/339662327

nornagon:
Updated. 🤞

nornagon:
Looks like the appveyor build failure was a transient network issue, so this should be good to merge now I think

PR #540, opened 2018-03-17
add Libraw

crose24:


crose24:
This was an accident, not done yet.

PR #547, opened 2018-04-03
ONNX (v1.2.2) preset

EmergentOrder:


EmergentOrder:
Done!
Thanks for the help along the way

EmergentOrder:
@saudet Just when I think I finally have these build issues fixed, something new pops up.
Specifically, 
```
[ERROR] /javacpp-presets/onnx/src/main/java/org/bytedeco/javacpp/onnx.java:[1187,16] repeated annotations are not supported in -source {0}
[ERROR] (use -source 8 or higher to enable repeated annotations)
[ERROR] /javacpp-presets/onnx/src/main/java/org/bytedeco/javacpp/onnx.java:[1189,16] org.bytedeco.javacpp.annotation.StdString is not a repeatable annotation type
```
This is happening today for the first time, only with JavaCPP 1.4.2-SNAPSHOT. When I point to 1.4.1, no compile error (locally).
I can log an issue, but I thought I'd just give you the heads up first (as I think this will fail my next CI build here)

saudet:
Oops, sorry about, fixed: https://github.com/bytedeco/javacpp/commit/be49bde573801e2899ce8149b434c2ccc4af317a

BTW, installing miniconda on the host with apt-get probably won't work. The builds happen in Docker containers.

Also, FYI, for faster turnaround with CI we can temporarily disable all other builds in `.travis.yml` and `.appveyor.yml`.

EmergentOrder:
Yeah, the build was taking forever, thanks for the suggestion, I disabled the other projects.
Also, I actually don't think I need apt-get or anything at all in install-travis.sh now.
There are a lot of commits here now, from tangling with the CI. If you want, I can put this all in a single commit (once it passes) to clean up the history.

saudet:
We can "squash" the commits when merging with GitHub, so don't worry about
that.

EmergentOrder:
@saudet I moved this to use protobuf3 and now I am getting : 
```[ERROR] Failed to execute JavaCPP Builder: /javacpp-presets/onnx/cppbuild/linux-x86_64/include/google/protobuf/generated_message_table_driven.h:126:Unexpected token '}```
I see that in the caffe preset you include this same file... so any ideas on why I get this error and apparently the caffe preset doesn't? I've tried this with protobuf 3.3.0 and 3.5.X

**Comment by EmergentOrder** [comment]
Success! I ended up switching back to protobuf 2.6.1 and building from source.
I'll let it pass one more time after a bit of cleanup, then I'll reinstate the rest of the projects in .appveyor.yml and .travis.yml and let the whole thing run.

File: onnx/autogen.sh.patch [hunk 1]
```diff
@@ -0,0 +1,14 @@
+--- protobuf-2.6.1/autogen.sh	2014-10-20 20:06:06.000000000 -0400
++++ autogen.sh	2018-04-15 19:38:24.032970600 -0400
+@@ -19,8 +19,9 @@
+ # directory is set up as an SVN external.
+ if test ! -e gtest; then
+   echo "Google Test not present.  Fetching gtest-1.5.0 from the web..."
+-  curl http://googletest.googlecode.com/files/gtest-1.5.0.tar.bz2 | tar jx
```
saudet: This is probably fixed with recent versions of protobuf. Let's try to get that running maybe? Please show me the kind of errors you get with 3.5.x.

onnx/cppbuild.sh
```diff
@@ -0,0 +1,72 @@
...
+
+if [[ $PLATFORM == windows* ]]; then
+    #No Windows support yet
+    :
+else
+
+    export PROTO=2.6.1
+    export INSTALL_PATH=`pwd`/$PLATFORM
+    wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O ~/miniconda.sh
+    bash ~/miniconda.sh -f -b -p $HOME/miniconda
```
saudet: Does Miniconda need to be installed in the $HOME directory? What happens if we place it in $INSTALL_PATH?

onnx/cppbuild.sh
```diff
@@ -0,0 +1,72 @@
...
+    #No Windows support yet
+    :
+else
+
+    export PROTO=2.6.1
+    export INSTALL_PATH=`pwd`/$PLATFORM
+    wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O ~/miniconda.sh
+    bash ~/miniconda.sh -f -b -p $HOME/miniconda 
+    export PATH="$HOME/miniconda/bin:$PATH"
+    conda install -y -c conda-forge protobuf=2.6.1 numpy setuptools
```
saudet: Does this need to create files in $HOME as hinted by the uninstallation procedure?
https://conda.io/docs/user-guide/install/linux.html#uninstalling-anaconda-or-miniconda
If we cannot get this to live somewhere else than $HOME, let's let users install it by themselves.
That is, place these commands in install-travis.sh for the CI servers as well as document them here:
https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments#linux-x86-and-x86_64

saudet:
BTW, according to https://github.com/onnx/onnx#source we don't need either Anaconda or Miniconda, we just need NumPy, and that's already available in CentOS:
https://github.com/bytedeco/javacpp-presets/blob/master/ci/install-travis.sh#L40
Have you tried to use that version of NumPy to build ONNX? Did it give you any errors you could not resolve?

EmergentOrder:
Moved to use $INSTALL_PATH for miniconda. 
Did not try installing NumPy by pip, because originally I needed the protobuf from conda-forge before I ended up building protoc myself.
Protobuf 2.6.1 is working fine for my purposes, I'm going to leave it at that for now. Feel free to follow up with 3.X if you can get it to work.

saudet:
We shouldn't even need to install numpy with pip. The default that comes
with CentOS should work...

saudet:
Ok, now everything builds alright with Python 3.5 and without installing Miniconda!

Now let's get protobuf updated... Would you have some time to tackle that? We probably just need to skip over a couple of "internal" from the header files.

EmergentOrder:
I still get the same error I posted above: ````[ERROR] Failed to execute JavaCPP Builder: /javacpp-presets/onnx/cppbuild/linux-x86_64/include/google/protobuf/generated_message_table_driven.h:126:Unexpected token '}'````
Already (trying to) skip over the containing union (AuxillaryParseTableField).

EmergentOrder:
If I follow an approach closer to the Tensorflow preset, I get many errors like : 
` /javacpp-presets/onnx/target/classes/org/bytedeco/javacpp/jnionnx.cpp:35614:5: error: ‘_NodeProto_default_instance_’ was not declared in this scope
     _NodeProto_default_instance_ = *ptr0;
`

EmergentOrder:
@saudet ok, I got past the default_instance declaration errors by skipping. I still get these 2 errors though:

```
javacpp-presets/onnx/target/classes/org/bytedeco/javacpp/jnionnx.cpp:1452:51: error: expression cannot be used as a function
         StringAdapter radapter(ptr->string_value_());
```

and 

```
javacpp-presets/onnx/target/classes/org/bytedeco/javacpp/jnionnx.cpp:1487:46: error: expression cannot be used as a function
         rptr = ptr->string_value_((char*)ptr0);```

**Comment by EmergentOrder** [comment]
once again, it looks like skip did the trick. I'll push my changes now

**Comment by EmergentOrder** [comment]
@saudet should be good to go

**Comment by saudet** [comment]
Thanks! Looks like ONNX 1.2.2 is out. Could we update the presets?

**Comment by saudet** [comment]
Would you also have some sample code we could put in a README.md file?

**Comment by EmergentOrder** [comment]
I had a branch with 1.1.0 ready to go, looks like I can bump that to 1.1.2 without issue. I'll push the merged changes from that now.
1.2.2 has errors, so I'll look at those at some point when I have a moment.

**Comment by EmergentOrder** [comment]
1.2.2 builds now but throws a runtime error in my first test with the jar

**Comment by EmergentOrder** [comment]
looks like an issue with an rvalue reference that was introduced

**Comment by saudet** [comment]
I've been fixing and cleaning this up, and now it looks good to merge! I've also changed the license to match the other presets, so please let me know if this not OK with you. Thanks!

**Comment by EmergentOrder** [comment]
Thanks for that, it looks good at a first glance.

 I'm still getting the runtime error I mentioned before, on this call:
`org.bytedeco.javacpp.onnx.OpSchemaRegistry.get_all_schemas()`
This call is essential for my application. Works in 1.1.2 and older, pre-rvalue.

The other issue is new with your changes, will comment inline

File: onnx/src/main/java/org/bytedeco/javacpp/onnx.java [hunk 4]
```diff
@@ -1194,7 +1192,7 @@ private native void allocate(
   // Get output formal parameters.
   public native @Const @ByRef FormalParameterVector outputs();
 
-  public native @Const @ByRef TypeConstraintParamVector typeConstraintParams();
+  public native @StdVector TypeConstraintParam typeConstraintParams();
```
EmergentOrder: Why does this no longer return a vector (TypeConstraintParamVector) ? This type change breaks my code.

saudet:
Ok, I've fixed that and it seems to work fine now. Thanks for testing!

EmergentOrder:
Great, both issues are fixed now, thanks!
Now I only get errors due to changes in ONNX internals in the more recent versions, but that's something else entirely...
I approve the license change.
Good to go.

PR #555, opened 2018-04-18
Added OpenCV Saliency Support

Skazzino:
Hi, as discussed in the issue [#958](https://github.com/bytedeco/javacv/issues/958) on the JavaCV repository, I've added the Saliency library of OpenCV3. However, since this is my first approach to JavaCV building, I want to know if everything is correct. Thanks.


saudet:
Looks good, thanks! Let's see what the CI servers have to say about it...

Issue #558, opened 2018-04-30
LLVM: Missing bindings for creating new 'CatchPad' instructions.


RobertStanforth:
The LLVM presets (great utility btw) lacks bindings for the following functions relating to the new 'CatchPad' style of exception handling:

LLVMBuildCatchPad
LLVMBuildCatchSwitch
LLVMBuildCatchRet
LLVMBuildCleanupPad
LLVMBuildCleanupRet

Would it be possible to implement these?

saudet:
I'm sure they can! BTW, contributions are welcome.

PR #559, opened 2018-05-01
Issue #558 - Bindings for creating new 'CatchPad' instructions

RobertStanforth:
Add bindings for:
LLVMBuildCatchPad
LLVMBuildCatchSwitch
LLVMBuildCatchRet
LLVMBuildCleanupPad
LLVMBuildCleanupRet


saudet:
This doesn't work. Those functions aren't part of LLVM 6.0.0. Where are
they from?

RobertStanforth:
Oh - it looks as if they're new in LLVM 7 (which is not yet released).
http://llvm.org/doxygen/IR_2Core_8cpp_source.html#l02561
I hadn't realised they were quite so cutting edge.
Thanks for taking a look.

saudet:
LLVM 7.0.0 has been released, the presets updated in commit https://github.com/bytedeco/javacpp-presets/commit/665537549df8538bf808c601a82ee604de611292 now contain wrappers for these functions, and snapshot builds are also available until the next release:
http://bytedeco.org/builds/
https://github.com/bytedeco/javacpp-presets/tree/master/llvm
Enjoy!

Issue #409, opened 2017-04-18
Sophisticated example for JavaCPP - Tensorflow


szingerpeter:
Dear All,

I have seen the [example](https://github.com/bytedeco/javacpp-presets/tree/master/tensorflow#the-srcmainjavaexampletrainerjava-source-file) for using Tensorflow with Java. I did not really manage to understand what was going on in the code. Would it be possible to have some more documentation on it? 
Moreover, I would be interested in a bit more sophisticated example or at least some guidance how I should / could use Tensorflow in Java. I think it is a great initiative, but I feel a bit lost on how I could implement for example a feed-forward neural network with for example 2 hidden layers and 5 classes by using Tensorflow JavaCPP bindings. 
Thank you for any help!

Cheers,
Peter

saudet:
It basically maps the C++ API, so for example we could easily port this example to Java:
https://github.com/tensorflow/tensorflow/tree/master/tensorflow/examples/label_image
That should be pretty easy to do! Sounds good?

szingerpeter:
Thank you for your reply.
Unfortunately, I am not familiar with C++, but will try to make sense out of the example you sent me!

saudet:
It shouldn't be too hard. If you have any questions, please post to the mailing list or ask on Gitter! Thanks

allenjack:
I agree with @szingerpeter. It is better to provide an example about how to train a simple neural network. I am also facing this problem that how to start from scratch with Java API. There are few materials can be referred to.

saudet:
@allenjack That's cool, contributions are welcome!

saudet:
@allenjack FYI, here is a complete training example using only the C++ API:
https://matrices.io/training-a-deep-neural-network-using-only-tensorflow-c/
It should be pretty straightforward to translate this to Java, but please let me know if you encounter any issues. Thanks for your interest!

allenjack:
Thank you so much. I will have a try and let you know.

saudet:
Many thanks to @Neiko2002, we now have a Java translation of that complete training example:
https://github.com/bytedeco/javacpp-presets/blob/master/tensorflow/samples/src/main/java/org/bytedeco/javacpp/samples/tensorflow/CarPricePredictionExample.java
Enjoy!

saudet:
BTW, it's now also possible to train using Keras:
https://github.com/bytedeco/javacpp-presets/blob/master/tensorflow/samples/KerasMNIST.java
[...]

PR #563, opened 2018-05-21
Add OutputVector class and a more complex example to Tensorflow

Neiko2002:
I have created a smaller version of the ExampleTrainer script and was working on a complete training example. In order to finish it some additional features were needed. e.g. the AddN constructor needed an InputList which itself  requires a vector of outputs. The vector does not exists right now, as far as I know.
```
public AddN(@Const @ByRef Scope scope, @ByVal InputList inputs) 
...
public InputList(@StdVector Output out)
```

The sames goes for the AddSymbolicGradients method
```
public static native @ByVal Status AddSymbolicGradients(@Const @ByRef Scope scope,
                            @StdVector Output outputs,
                            @StdVector Output inputs,
                            @StdVector Output grad_outputs)
```

This PR creates a OutputVector class and adds three examples: a simpler version of the ExampleTrainer script, a InputListExample and the more sophisticated examples requested in issue #409.


saudet:
Looks good, thanks! Since the samples are not run as tests though, could we put them under the `samples` subdirectory in their own little project with a `pom.xml` file to make them more accessible? (FYI, I'd eventually like to run samples automatically as part of tests, but don't worry about that for now.)

PR #567, opened 2018-05-30
Add tensorflow with gpu support for windows

Neiko2002:
Tested with cuda 9.0 + cudnn 7.0 and cuda 9.2 + cudnn 7.1

Is it possible to merge attributes from other platform annotation? Right now I had to copy the "link" attributes from the "windows" platform to the "windows-x86_64" with "-gpu" extension.

Furthermore it might be better to exchange the "-gpu" extension with "-cuda90" or "-cuda92". On the other hand all the different configurations are horrible (e.g. cuda version, cudnn version, cpu instruction set) to maintain.

### DIFF
--- a/tensorflow/README.md
+++ b/tensorflow/README.md
@@ -60,6 +60,12 @@ We can use [Maven 3](http://maven.apache.org/) to download and install automatic
             <version>1.8.0-1.4.2-SNAPSHOT</version>
             <classifier>macosx-x86_64-gpu</classifier>
         </dependency>
+        <dependency>
+            <groupId>org.bytedeco.javacpp-presets</groupId>
+            <artifactId>tensorflow</artifactId>
+            <version>1.8.0-1.4.2-SNAPSHOT</version>
+            <classifier>windows-x86_64-gpu</classifier>
+        </dependency>
 
     </dependencies>
 </project>
--- a/tensorflow/cppbuild.sh
+++ b/tensorflow/cppbuild.sh
@@ -145,7 +145,9 @@ case $PLATFORM in
         mkdir -p ../build
         cd ../build
         "$CMAKE" -A x64 -DCMAKE_BUILD_TYPE=Release -DPYTHON_EXECUTABLE="C:/Python27/python.exe" -Dtensorflow_BUILD_PYTHON_BINDINGS=OFF -Dtensorflow_BUILD_SHARED_LIB=ON -Dtensorflow_WIN_CPU_SIMD_OPTIONS=/arch:AVX -G"Visual Studio 14" $CMAKE_GPU_FLAGS -DCUDNN_HOME="$CUDA_PATH" ../tensorflow-$TENSORFLOW_VERSION/tensorflow/contrib/cmake
-        MSBuild.exe //p:Configuration=Release /maxcpucount:$MAKEJ tf_core_gpu_kernels.vcxproj
+        if [[ "$EXTENSION" == *gpu ]]; then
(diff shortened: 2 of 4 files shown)


saudet:
Unfortunately, it's not currently possible to "merge" annotation values as it would make it hard to override them. If you have a good idea to enhance JavaCPP that would allow both concatenating and overriding results easily with annotations like this, let me know!

Other than that, we'll need to leave ` -DPYTHON_EXECUTABLE="C:/Python27/python.exe" ` in there because it won't build with the version of Python that comes with MSYS2.

Neiko2002:
Added the PYTHON_EXECUTABLE again.

saudet:
I can't get it to build here. The Python script don't seem to be running properly. Which distribution of Python are you using?

Neiko2002:
Python 3.6 was in the path of the system. Which python script has problems?

saudet:
No, it doesn't seem to be because of the version of Python, it doesn't work with 3.6.5 either. I'm getting this kind of error:
```
CMake Error: Could not open file for write in copy operation C:/msys64/tmp/javacpp-presets/tensorflow/cppbuild/windows-x86_64-gpu/tensorflow-1.8.0/tensorflow/contrib/cmake/build/CMakeFiles/tf_core_gpu_kernels.dir/__/__/core/kernels/tf_core_gpu_kernels_generated_cwise_op_gpu_squared_difference.cu.cc.obj.cmake.pre-gen.tmp
CMake Error: : System Error: No such file or directory
CMake Error at C:/msys64/mingw64/share/cmake-3.8/Modules/FindCUDA.cmake:1511 (configure_file):
  configure_file Problem configuring file
Call Stack (most recent call first):
  C:/msys64/mingw64/share/cmake-3.8/Modules/FindCUDA.cmake:1723 (CUDA_WRAP_SRCS)
  tf_core_kernels.cmake:217 (cuda_add_library)
  CMakeLists.txt:448 (include)
```
A Python script somewhere is supposed to create those .cmake.pre-gen.tmp files, but they don't get (all) created for some reason here on my installation of Windows 10.

saudet:
All the failing paths are longer than 260 characters, so this has something to do with the MAX_PATH limit:
https://bugs.python.org/issue18199
https://bugs.python.org/issue27731
But enabling support for it in the registry does not help, hum...

Neiko2002:
I had path length problems somewhere else, so my directory is just called: f:/G/JP/tensorflow/. But if this is really an issue we should check it somewhere and tell the user beforehand.

saudet:
This is working well for me, thanks! I've uploaded snapshots manually and will keep doing that until we figure out what to do about the CI service (issue #568).

Let's hope they get Bazel working before long; a build time of 4 hours is pretty ridiculous. As far as I know, we're only missing support for static libraries: https://github.com/bazelbuild/bazel/issues/1920

BTW, keep an eye on https://github.com/tensorflow/tensorflow/issues/17390. Google is expending effort to build a community around support for Java.

Neiko2002:
@saudet Thanks for pointing me to #17390. There are too many Tensorflow Github tickets to follow. I will be at a conference in Yokohama (Japan) next week and have same holidays afterwards. When I'm back in office I could start working on examples for feature.pb.h, example.pb.h, record_reader.h, record_writer.h.

The other day I was building Tensorflow with Bazel but had the feeling it takes much longer on Windows and uses less available resources as cmake.

saudet:
Which conference is that BTW? I live in Japan, not too far from Yokohama/Tokyo...

Issue #587, opened 2018-07-12
FFmpeg Android improvements


tmm1:
I maintain the android mediacodec decoder in FFmpeg, and would love to see the ffmpeg javacpp preset support it out of the box. The only thing that should be required to make this happen is adding `--enable-jni --enable-mediacodec` to the configure invocation for android builds.

Another thing I noticed while poking around here.. the `ffmpeg-android.patch` can be dropped in favor of a simple `--disable-symver`.

tmm1:
> Another thing I noticed while poking around here.. the ffmpeg-android.patch can be dropped in favor of a simple --disable-symver.

Looks like `--disable-symver` is already being used. I believe the patch can be removed, but perhaps I'm mistaken.

saudet:
Sounds great! Please send a pull request and Travis CI will tell us if
there are any build issues.

PR #588, opened 2018-07-12
Remove unnecessary ffmpeg-android.patch

tmm1:
cc #587


saudet:
Without the patch, I'm still seeing library versioning happening:
```
-rw-r--r--. 1 saudet saudet 3462256 Jul 13 14:18 libavutil.a
lrwxrwxrwx. 1 saudet saudet      22 Jul 13 14:18 libavutil.so -> libavutil.so.56.14.100
lrwxrwxrwx. 1 saudet saudet      22 Jul 13 14:18 libavutil.so.56 -> libavutil.so.56.14.100
-rwxr-xr-x. 1 saudet saudet  394976 Jul 13 14:18 libavutil.so.56.14.100
```
Is there a way to disable that without using a patch? Android doesn't officially support that: https://issuetracker.google.com/issues/36971591

tmm1:
I didn't see those when looking at ffmpeg/cppbuild/android-arm/lib 

What directory is your listing from?

saudet:
It's doing that for at least `android-arm64`...

saudet:
According to the logs on Travis CI, it's happening for all architectures:
https://api.travis-ci.org/v3/job/403299556/log.txt
https://api.travis-ci.org/v3/job/403299566/log.txt
https://api.travis-ci.org/v3/job/403299585/log.txt
https://api.travis-ci.org/v3/job/403299595/log.txt

tmm1:
Looks like this also requires `--target-os=android`, and it was only working for me because I tested a branch that had both of my PRs merged together.

saudet:
Indeed seems to work fine that way, so let's merge this with pull #589 and test the build together?

PR #589, opened 2018-07-12
Build ffmpeg for android with mediacodec support

tmm1:
cc #587


saudet:
I'm getting an `ERROR: jni not found` message from FFmpeg's configure script with that, at least for `android-arm64`, with thess details in the `config.log`:
```
/home/saudet/Android/android-ndk//toolchains/aarch64-linux-android-4.9/prebuilt/linux-x86_64/bin/aarch64-linux-android-gcc -Wl,--no-undefined -z text -L../lib/ -L/home/saudet/Android/android-ndk//sources/cxx-stl/gnu-libstdc++/4.9//libs/arm64-v8a/ -Wl,--no-undefined -z text --sysroot=/home/saudet/Android/android-ndk//platforms/android-21/arch-arm64/ -Wl,--as-needed -Wl,-z,noexecstack -o /tmp/ffconf.tD50XJ6R/test /tmp/ffconf.tD50XJ6R/test.o -lgnustl_static -lgcc -ldl -lz -lm -lc
/tmp/ffconf.tD50XJ6R/test.o: In function `foo':
test.c:(.text.foo+0x7c): undefined reference to `cexp'
collect2: error: ld returned 1 exit status
ERROR: jni not found
```
You're not getting this?

tmm1:
I was seeing that with an earlier commit before I changed target-os to android.

Let me try with android-arm64 here..

tmm1:
The requirements on the ffmpeg side are pretty simple: https://github.com/FFmpeg/FFmpeg/blob/master/configure#L5995

The build works on my Mac:

<code>
External libraries:
jni                                libvo_amrwbenc
</code>

Maybe I'm not running it right?

<code>
ANDROID_NDK=$HOME/Downloads/android-ndk-r15c
mvn install --projects .,ffmpeg -Djavacpp.platform=android-arm64 -Djavacpp.platform.compiler=$HOME/Downloads/android-ndk-r15c/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/bin/aarch64-linux-android-g++ -Djavacpp.platform.root=$HOME/Downloads/android-ndk-r15c/
</code>

Or maybe I have jni.h in my environment? It exists in the NDK sysroot too so it seems like it would still work.

tmm1:
cppbuild.sh directly works too:

>ANDROID_NDK=~/Downloads/android-ndk-r15c bash cppbuild.sh -platform android-arm64 install ffmpeg

Should I try on Linux?

saudet:
Yeah, seems something specific to Linux for some reason.

saudet:
But it looks like it works on Travis CI, strange...

saudet:
Ah, it looks like I was missing `--target-os=android`. With that, it seems to build fine for me locally on Linux as well, thanks!

tmm1:
Great! So this should be good to do.

PR #596, opened 2018-08-06
enable python_bindings and link `tf_python_protos_cc`, `tf_c_python_api`

wumo:
commit e61ec03e1f76c07276b6e52df7dae9f25d7e9256 requires additional lib `tf_c_python_api` to be linked and `tf_c_python_api` depends on `tf_python_protos_cc`.

 `tf_c_python_api`  lies in `${tensorflow.path}/../build/tf_c_python_api.dir/Release/`
`tf_python_protos_cc` lies in `${tensorflow.path}/../build/Release/`

so in this change, we enabled python_bindings and linked `tf_python_protos_cc`, `tf_c_python_api`.


saudet:
I've updated the build scripts with more minimal changes so that it should work on AppVeyor if we had enough build minutes, but we don't... Anyway, running the build for `tf_c_python_api.vcxproj` after `tensorflow_static.vcxproj` takes only a couple of minutes, so the change in build time is really minimal. I've also deployed manually snapshot artifacts from my local machine, so please try them out! And let me know if this is good to merge @wumo @Neiko2002 Thanks!

PR #598, opened 2018-08-14
Samples for clang

Neiko2002:
Here are some samples on how to use the clang preset. Each java file works on its own, no dependencies to other libraries (except the llvm preset of course) or other java files in the project.


saudet:
Looks good! Thanks

PR #607, opened 2018-08-21
Build with Traditional ML support (ONNX-ML)

EmergentOrder:
This builds without error, but it still doesn't return the Traditional ML ops on call ```org.bytedeco.javacpp.onnx.OpSchemaRegistry.get_all_schemas()```.
It returns 115 ops, whereas doing the same thing in python returns 133.

Digging in a bit more, calling `org.bytedeco.javacpp.onnx.RegisterOnnxMLOperatorSetSchema()` 
results in:

`java.lang.UnsatisfiedLinkError: org.bytedeco.javacpp.onnx.RegisterOnnxMLOperatorSetSchema()V
  at org.bytedeco.javacpp.onnx.RegisterOnnxMLOperatorSetSchema(Native Method)`

Which is strange, since we already link to both of the .so files the ONNX build produces.

I think the root cause is again the rvalue thing, specifically this line in schema.h : ```void RegisterSchema(OpSchema&& schema)```.
Is it possible to apply `std::move` to get the rvalue?

Any ideas?

### DIFF
--- a/onnx/README.md
+++ b/onnx/README.md
@@ -57,7 +57,9 @@ import static org.bytedeco.javacpp.onnx.*;
 
 public class LoadModel {
     public static void main(String[] args) throws Exception {
-        OpSchemaRegistry.get_all_schemas();
+        RegisterOnnxMLOperatorSetSchema();
+        OpSchemaVector allSchemas = OpSchemaRegistry.get_all_schemas();
+        System.out.println(allSchemas.size());
 
         byte[] bytes = Files.readAllBytes(Paths.get("examples/resources/single_relu.onnx"));
 
--- a/onnx/README.md
+++ b/onnx/README.md
@@ -57,7 +57,6 @@ import static org.bytedeco.javacpp.onnx.*;
 
 public class LoadModel {
     public static void main(String[] args) throws Exception {
-        RegisterOnnxMLOperatorSetSchema();
         OpSchemaVector allSchemas = OpSchemaRegistry.get_all_schemas();
         System.out.println(allSchemas.size());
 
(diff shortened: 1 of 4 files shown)


saudet:
Maybe, but we're not calling `RegisterSchema()` anywhere, right? AFAIK, this gets called internally. So I think it's something with the build...

EmergentOrder:
Well no, not explicitly calling it anywhere from client / JavaCPP code. It is called in schema.h from `RegisterOpSetSchema` which is called from the operator-sets .h files, and the result should be population of the OpSchemaRegistry. 
This mechanism for opset population is new in 1.2.X. 
 
But the build fails if we don't skip `std::function<void(OpSchema&&)`, on line ` public static native void ForEachSchema(@ByVal std::function<void(OpSchema&&)> fn);` in the generated Java and `onnx::RegisterSchema` , with ` error: cannot bind rvalue reference of type ‘onnx::OpSchema&&’ to lvalue of type ‘onnx::OpSchema’
         onnx::RegisterSchema(*ptr0);
`.
I mentioned it works from Python, just to clarify, this is even in the case when installed from the same JavaCPP-Presets/ONNX build .

EmergentOrder:
See https://github.com/onnx/onnx/pull/894

saudet:
Well, it works fine here. I've updated the sample in the README.md file and it display "133" for me. So, good to merge?

EmergentOrder:
Hmm, ok there is something funky with a local install then. I got the error by following the README : `mvn install --projects ../onnx/platform -Djavacpp.platform.host`. 
When I just do a standard `mvn clean package` and copy the jars to my project lib directory, no error.
Actually, looking at it now after wiping my .m2 repo, I see it downloading the jars from the Sonatype snapshots instead of using my local.
Also, it looks like the call you added to the sample to `RegisterOnnxMLOperatorSetSchema()` is now redundant, as that call happens internally already, as it should (I get 133 now without it).

EmergentOrder:
So yeah, good to merge!

PR #609, opened 2018-08-25
added make options for gpu build to mxnet

bitstormGER:
I have added options for gpu / cuda build.
To use cuda set CUDA_HOME to the folder containing cuda (e.g. /opt/cuda for arch linux) in the environment of the build
To use cudnn set USE_CUDNN=1

The has only been tested on arch linux_x86_64 where it compiles


saudet:
Thanks! That's a good start. Let's include it as an "extension" just like with Caffe and TensorFlow (check their `cppbuild.sh`, `presets/caffe.java`, and `presets/tensorflow.java` files), and add a separate entry for it to `.travis.yml` as well, so that CUDA gets installed when it gets to build it.

bitstormGER:
I have added the extension handling but I am not sure how to test it. Please check if CUDA home needs to be changed

saudet:
Thanks! Let's wait and see what Travis CI has to say about it...

saudet:
The build passes, but it doesn't appear to be compiling any CUDA kernels...?

bitstormGER:
I think the problem is the different CUDA_HOME path. I have seen that the cuda project expects the standard /usr/local/cuda, I had /opt/cuda so it worked only on arch.
I have changed this now and it a test build with a softlink works fine here.

Sorry for missing that.

saudet:
Looks like the build is failing:
```
In file included from src/operator/contrib/./../tensor/sort_op.h:104:0,
                 from src/operator/contrib/./bounding_box-inl.h:37,
                 from src/operator/contrib/bounding_box.cu:27:
src/operator/contrib/./../tensor/./sort_op-inl.cuh:34:44: fatal error: cub/device/device_radix_sort.cuh: No such file or directory
 #include <cub/device/device_radix_sort.cuh>
```
https://travis-ci.org/bytedeco/javacpp-presets/jobs/421178460

BTW, it's basically building inside a Docker container using "nvidia/cuda:9.2-cudnn7-devel-centos7" as image, so you could run with that locally to make sure it works alright.

saudet:
Using the release archives seems to have fixed that issue!

PR #615, opened 2018-09-07
Upgrade presets for ONNX 1.3.0

EmergentOrder:


EmergentOrder:
Thanks for fixing it up. I almost had that, just missing the @Function annotation, but I fell back to the patchfile, figured you could find the missing piece faster than I would.

saudet:
Let me know when this is good to merge! Thanks

EmergentOrder:
I get `Unexpected token 'EOF'` at optimize.h:99

EmergentOrder:
And in onnxifi.h, it seems to think we are on windows:

```
#if defined(_WIN32) && defined(_M_IX86)
/* Windows x86 */
#define ONNXIFI_ABI __stdcall
```

results in: 
`public static final int ONNXIFI_ABI = __stdcall;`
which clearly doesn't compile.

saudet:
I see. BTW, I wrote a guide recently about what to do in general:
https://github.com/bytedeco/javacpp/wiki/Mapping-Recipes

EmergentOrder:
I saw that, much appreciated.
Fixed the macro issue using the instructions provided there.
The EOF seems to be due to comments at the end of a file. That needs to be fixed in JavaCPP no?

EmergentOrder:
Not due to the comment I guess, as schema.h has it too, but I'm still not sure how to handle it

saudet:
`optimize.h` includes a lot of other files, it's probably missing some definitions from those.

EmergentOrder:
Ok, you can go ahead and merge this.
I may add the other headers at some point down the line, in another PR.

PR #619, opened 2018-10-02
Add X265 multilib support for osx and linux-x86*

renderdude:
X265 supports API selectable internal compression depths of 8,10, or 12 bits. This commit enables building of multilib support for X265 on osx and linux-x86* and enables this feature in ffmpeg.


saudet:
Thanks for the contribution! But it doesn't build:
```
ERROR: libx265 not found
```
The `$INSTALL_PATH` needs to be used to have make install them in the right place. BTW, instead of using a patch to modify `multilib.sh`, we could directly launch these multiple builds from `cppbuild.sh`. That would make it easier to test and fix.

renderdude:
Which platform failed?

saudet:
Linux, Fedora 27, but you'll see it on Travis CI with CentOS 6.

saudet:
See here: https://travis-ci.org/bytedeco/javacpp-presets/jobs/436238096
That's failing because it doesn't find `cmake` though, we need to call `cmake3` on CentOS.

renderdude:
I’m in the process of rewriting the code to inline it in cppbuild.sh and use INSTALL_PATH and $CMAKE. Should have a new pull request later today.

saudet:
We can skip things like AVX for 32-bit builds. Those binaries are not used for performance anyway.

renderdude:
So the last patch fixed the compile, but now it's erroring in deploy (XCode). As it doesn't appear to be related to this patch, I assume I can ignore.

saudet:
Yes, it's looking good, thanks! I'd like to do the same with other platforms before merging this though. If you could give it a try, that'd be great.

saudet:
Looking good! Builds are passing on Windows. Thanks!

saudet:
This is pretty much it: https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments#android-arm-and-x86

It doesn't work? What message do you get with your crash?

renderdude:
It gets to OpenSSL (I think, we really need a print statement indicating each step of the build), and dies with this:

SIXTY_FOUR_BIT_LONG mode
RC4 uses unsigned char

Configured for android64.
clang70clang70: error: unknown argument: '-mandroid'
: error: unknown argument: '-mandroid'
clang70: warning: -z text: 'linker' input unused [-Wunused-command-line-argument]

renderdude:
It gets to OpenSSL (I think, we really need a print statement indicating each step of the build), and dies with this:

SIXTY_FOUR_BIT_LONG mode
RC4 uses unsigned char

Configured for android64.
clang70clang70: error: unknown argument: '-mandroid'
: error: unknown argument: '-mandroid'
clang70: warning: -z text: 'linker' input unused [-Wunused-command-line-argument]

saudet:
Ah, you're trying to build with Clang, that won't work, keep using GCC.

You mean like echo statements in the script? We can add those sure, or just
enabling echo of the commands would be easier I think.

renderdude:
The issue is that I’m using the cppbuild straight out of the box. So where it’s picking up clang I don’t know. My guess is that it’s an install issue. I’m using r18

Rather than echoing commands, I would have it output something like:

-------------------
Building zlib
-------------------

… compile output …

-------------------
Building Lame
-------------------

… compile output

At least that way it’s easier to detect which package failed, rather than hunting through the compile output looking for an include or install directory.

renderdude:
The issue is that I’m using the cppbuild straight out of the box. So where it’s picking up clang I don’t know. My guess is that it’s an install issue. I’m using r18.

Rather than echoing commands, I would have it output something like:

-------------------
Building zlib
-------------------

… compile output …

-------------------
Building Lame
-------------------

… compile output

At least that way it’s easier to detect which package failed, rather than hunting through the compile output looking for an include or install directory.

saudet:
r18 doesn't have GCC anymore, use r15c as per these instructions:
https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments#android-arm-and-x86

saudet:
Everything should work pretty much as is with r15c, for now. We'll need to do something about Clang eventually when other projects like FFmpeg and TensorFlow start supporting it, but it's not really the case for now, unfortunately.

saudet:
We can echo commands here and there too that's fine :+1: Thanks

renderdude:
All of the android versions have been committed.

Mark

PR #622, opened 2018-10-12
Adds ONNX Optimizer, Version Converter and IR.

EmergentOrder:
The goal here is being able to ["polish"](https://github.com/onnx/onnx/blob/v1.3.0/onnx/utils.py#L14) a model, as well as convert it to the latest version, via the preset.
Some notes:

- The patchfile for optimize.h illustrates that JavaCPP expects semicolons where none are strictly necessary.
- The patchfile for optimize_pass.h illustrates an issue with the combination of `virtual` and `inline` for constructor declarations
- I had to skip lots of things, some of which might be important, I have yet to test.


EmergentOrder:
Thanks for the fixes!

I tested check_model and added it to the README.

When I try Optimize(...) I get : 
```
java.lang.UnsatisfiedLinkError: org.bytedeco.javacpp.onnx.Optimize(Lorg/bytedeco/javacpp/onnx$ModelProto;Lorg/bytedeco/javacpp/onnx$StringVector;)Lorg/bytedeco/javacpp/onnx$ModelProto;
  at org.bytedeco.javacpp.onnx.Optimize(Native Method)
```

EmergentOrder:
Similarly for InferShape(...) : 
```
java.lang.UnsatisfiedLinkError: org.bytedeco.javacpp.onnx.InferShapes(Lorg/bytedeco/javacpp/onnx$ModelProto;)V
  at org.bytedeco.javacpp.onnx.InferShapes(Native Method)
```

I still need to test version conversion.

saudet:
That sounds like the build is missing some options that would include those symbols in the library.

EmergentOrder:
Yeah, that was my first thought too. The thing is, there don't seem to be any build options for this, it should already be there in the libonnx.so.

There's no mention of build options in the docs:
https://github.com/onnx/onnx/blob/v1.3.0/docs/Optimizer.md
https://github.com/onnx/onnx/blob/v1.3.0/docs/ShapeInference.md

And in fact the .so matches when I grep for the function names ( `GetAvailablePasses` from the optimizer for example).
Maybe the generated function signature doesn't match what's in the .so ?

saudet:
Check with `readelf -s`, they probably who up as "undefined".

EmergentOrder:
Hmm, ok I'm thinking this related to what I reported in one of the previous PRs, 
where `mvn install --projects ../onnx/platform -Djavacpp.platform.host` as per README pulls jars from sonatype instead of using the locally built ones. 
So after wiping my local maven repo, I now get 
`error: not found: value GetAvailablePasses` instead of the link error.

EmergentOrder:
Actually, that was installing from platform without installing the local build first.
When I do it that way (and immediately, or using the maven option `--no-snapshot-updates`), I'm back to the link error and seeing that GetAvailablePasses IS in fact defined in the .so (I checked in my local maven repo after install). 
So I'm stumped.

EmergentOrder:
Damn.. It turns out that, yep, it was the same issue as a few months ago, my local was not being used.
This time I determined that it actually installed ok to the local .m2 repo, so now I'm looking in sbt's direction instead of mvn.

EmergentOrder:
Ready to merge!

still haven't figured out how to maven install platform without getting the `UnsatisfiedLinkError` (offline mode and no snapshot updates don't seem to do it), but working around with manual jar copy to the lib dir works.

saudet:
Yup, it's working fine here! Thanks. Good to merge?

EmergentOrder:
Yep, good to merge.

PR #642, opened 2018-11-19
nGraph Preset initial commit

EmergentOrder:
Another preset incoming!

So I've managed to get it down to just one error on the jningraph.cpp file:

```
jningraph.cpp:9879:60: error: ‘position1’ was not declared in this scope
         env->SetLongField(arg1, JavaCPP_limitFID, rsize1 + position1);
```

Any idea what's going on here? This is in the class NgraphResultVector. There are several other NgraphSOMEVector, defined similarly, which don't seem to have the same issue.


saudet:
Thanks! The issue might have something to do with recent changes in commit https://github.com/bytedeco/javacpp/commit/f214f5c3aa14d2231bfe4e8d0fbab394a7a50a30? But there isn't any buffers used in the class... I guess I'll have to look at this.

EmergentOrder:
After adding `runtime::Tensor` and `runtime::Backend`, now I get 2 `position` related errors.
Also, we are missing the `from<SOME>` functions from `ngraph::element::Type`, because they failed to parse (see `element_type.hpp.patch`).

saudet:
BTW, add an entry to `.travis.yml` to have the build tested at least on Linux.

EmergentOrder:
Added BackendManager, which raises these new errors:

``` 
error: ‘void*’ is not a pointer-to-object type
         ::ngraph::onnxifi::BackendManager::unregister(*ptr0);`
```

```
 error: ‘void*’ is not a pointer-to-object type
         rptr = (const ::ngraph::onnxifi::Backend*)&::ngraph::onnxifi::BackendManager::get(*ptr0);
```

saudet:
I'm not able to build nGraph with `cppbuild.sh`, it fails with this error (Fedora 27):
```
[ 60%] Linking CXX shared library libcodegen.so
/usr/bin/ld: cannot find -ltinfo
collect2: error: ld returned 1 exit status
make[2]: *** [src/ngraph/codegen/CMakeFiles/codegen.dir/build.make:188: src/ngraph/codegen/libcodegen.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:759: src/ngraph/codegen/CMakeFiles/codegen.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 61%] Linking CXX shared library ../../libinterpreter_backend.so
[ 61%] Built target interpreter_backend
make: *** [Makefile:152: all] Error 2
```
Could you first get the `cppbuild.sh` script working? Also, you're missing the "header" to ensure that it builds under the `cppbuild` subdirectory, and let's use the `download` function instead of `git` as per what we did for ONNX:
https://github.com/bytedeco/javacpp-presets/blob/master/onnx/cppbuild.sh
Thanks for your time on this!

EmergentOrder:
Followed the [instructions](http://ngraph.nervanasys.com/index.html/buildlb.html#centos) for CentOS, on Fedora 27 (in Docker) and I get a successful build (up to the errors I reported above).

Without installing the required packages in Fedora, I got errors before even reaching `cannot find -ltinfo`, but I guess you have those requirements already. It looks like `tinfo` is in `ncurses`.

So I'll `make install` ncurses in `cppbuild.sh`, not sure if that's the only thing that I need to add though.

EmergentOrder:
Works for me now on Fedora 27 without package `ncurses-devel`.
Working on your end?

saudet:
In other news, I've fixed compilation failing with `error: ‘position1’ was not declared in this scope` in commit https://github.com/bytedeco/javacpp/commit/99835a5b8cc26e2162f30ba4b8dbf188e389cb0f. For the other one with `onnxBackendID`, the problem is that it's an `@Opaque` type and it's trying to call functions that need its definition with `@ByVal`. You'll need to either skip those functions, or include a header file with the definition for `onnxBackendID`.

EmergentOrder:
Actually, it looks like it parses `element_type.hpp` ok, but then fails on the generated java.

saudet:
We just need to give them names like:
```java
.put(new Info("ngraph::element::from<char>").javaNames("fromChar"));
...
```
See https://github.com/bytedeco/javacpp/wiki/Mapping-Recipes#specifying-names-to-use-in-java and https://github.com/bytedeco/javacpp/wiki/Mapping-Recipes#creating-instances-of-c-templates

BTW, since it appears to install everything properly in the install path, we probably don't need to copy everything manually.

saudet:
Another build issue, the `ngraph` module is missing from the parent `cppbuild.sh` and `pom.xml` files:
https://travis-ci.org/bytedeco/javacpp-presets/jobs/460812836
Also, it most likely isn't going to build on CentOS 6, so let's add it to the list for CentOS 7 here:
https://github.com/bytedeco/javacpp-presets/blob/master/ci/install-travis.sh#L49

pom.xml
```diff
@@ -535,6 +535,7 @@
         <module>tensorrt</module>
         <module>ale</module>
         <module>onnx</module>
+        <module>ngraph></module>
```
saudet: Please fix the syntax error here.

saudet:
The unit tests are not building for some reason, but we can probably just skip those:
https://travis-ci.org/bytedeco/javacpp-presets/jobs/462493888

saudet:
It's not jut the unit tests, it failing to build any executables like `nbench` now:
https://travis-ci.org/bytedeco/javacpp-presets/jobs/463607246

saudet:
Now it's failing on this command:
> ```
> rm: cannot remove '../../lib/libtbb.so': No such file or directory
> ```
https://travis-ci.org/bytedeco/javacpp-presets/jobs/463895675

BTW, is there a reason we're keeping all these lines of `cp`, `rm`, and `ln`?
Does the build fail without them?

Can we also remove `element_type.hpp.patch` since it's no longer used?

saudet:
Getting further, now the build can't find where nGraph's libraries are:
```
/usr/bin/ld: cannot find -ltbb
/usr/bin/ld: cannot find -lcodegen
/usr/bin/ld: cannot find -lcpu_backend
/usr/bin/ld: cannot find -lonnxifi-ngraph
/usr/bin/ld: cannot find -lngraph
/usr/bin/ld: cannot find -lonnxifi
```
https://travis-ci.org/bytedeco/javacpp-presets/jobs/464387300

EmergentOrder:
I know that it's just CentOS 7 in Docker on Travis. Thing is, I already had this working on Fedora 7 and Ubuntu and was hoping it would *just work*. Also didn't have much time this week to look at this.
Anyway I just tried it inside Docker on my machine, I get other issues (specifically not being able to curl to https). So I just disabled the other builds in Travis and AppVeyor to speed things up.

saudet:
Make sure to run these commands to install what's needed:
https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments#linux-x86-and-x86_64

saudet:
It's building now, great! But it's not copying TBB with it's version number so it probably won't run:
```
[INFO] Copying /home/travis/build/javacpp-presets/ngraph/cppbuild/linux-x86_64/lib/libtbb.so
```
We'll need to specify the version in the presets like `@Platform(..., link = {"tbb@.2", ...}, ...)`.

EmergentOrder:
K, so next steps, sample usage would be good.

I'm looking to use the ONNX ops in `onnx_import`, as their ONNXIFI support is not ready yet. [Haven't committed the changes to get there, yet].
 By itself, this doesn't require linking to `libonnx.so`. But, it looks like in order to get the definitions I need in order to instantiate NodeProto, ModelProto and GraphProto, we need the dependency on ONNX. I tried to add this directly, but I'm getting linking errors on functions I've skipped, so something isn't right.

 Is there a way I can use the definitions we already have from the ONNX preset, and swap them in for the empty "forward declarations" in the nGraph code? I couldn't find a way to skip those definitions without also skipping other references, for a start. Then there's getting the import in the generated nGraph Java.

 Ideas?

saudet:
We can add a dependency on the ONNX presets, yes. We'll probably need to bundle its header files for that to work first. Take a look at Caffe and OpenCV, for example.

EmergentOrder:
Ok I was able to get the Proto classes from ONNX.

Now I'm blocked on parsing errors on `ngraph/frontend/onnx_import/core/weight.hpp` , `ngraph/frontend/onnx_import/core/value_info.hpp` and `ngraph/frontend/onnx_import/core/tensor.hpp`.
specifically, on declarations with empty bodies.

saudet:
Ok, I'll have to look into that. BTW, are those header files from ONNX? If that's the case, they should be parsed as part of the ONNX presets.

EmergentOrder:
No, they're not, they're part of ngraph.

saudet:
Ok, I've fixed the issue in commit https://github.com/bytedeco/javacpp/commit/fbf9e958e564bc0c93dec201e8e6d650f10524ef.
It's parsing fine now. Thanks for reporting!

EmergentOrder:
Great, thanks for the fix.
One more file failing to parse now: `ngraph/op/constant.hpp`, could you take a look at that too?

saudet:
Sure thing, fixed in commit https://github.com/bytedeco/javacpp/commit/2cf75e542f1a2756e5a918872a90720e895fe239!

EmergentOrder:
The last build failed, not on parsing this time, but on the generated Java.

saudet:
Yes, we need more Info.

saudet:
The current error shouldn't be hard to resolve, but as usual, let me know if you need help to make progress.

EmergentOrder:
No problem, I'll take care of that.

Another issue, brought by further changes I haven't check in yet:
`error: cannot convert ‘const Operator* {aka const std::function<ngraph::NodeVector(const ngraph::onnx_import::Node&)>*}’ to ‘ngraph::NodeVector (*)(const ngraph::onnx_import::Node&)’ in assignment
`

Where I am declaring my own FunctionPointer for Operator.
I tried adding `@Const` to the declaration, but I get a bunch of other errors related to const/volatile.

Any ideas?

saudet:
We can't convert an std::function into a function pointer or a function
object because the target may not be either a function pointer or a
function object. Only creating an std::function from a function pointer or
a function object makes sense, not the other way around.

EmergentOrder:
hmm, ok, looking for an alternative to `get_operator`,
I found `@Namespace("ngraph::onnx_import") public static native @ByVal NgraphFunctionVector load_onnx_model(@Cast("std::istream*") @ByRef Pointer sin);`

How can I load a file into a Pointer containing an istream so I can call this?

saudet:
Enhancing JavaCPP to generate custom call() methods for instances of
std::function might become necessary at some point, but I haven't
encountered that need yet. It's always possible to define the interface
manually anyway.

The same for C++ streams, we'd have to map that API first, but it's very
inefficient so libraries typically offer alternatives such as reading from
a memory buffer or something anyway.

EmergentOrder:
This is what I had for Operator:
`        public static class Operator extends FunctionPointer {
        static { Loader.load(); }
        public    Operator(Pointer p) { super(p); }
        protected Operator() { allocate(); }
        private native void allocate();
        public @ByVal native @Cast("ngraph::NodeVector*") Pointer call(@ByRef @Cast("const ngraph::onnx_import::Node*") Pointer a);
   }
`
so I did define it manually, but got the error when trying to use it. 
I'm not clear on how generated custom call() methods would help the `cannot convert` error I hit. It looked to me like the issue was just not being able to assign something to const to something non-const in the generated C++.

saudet:
No, I mean defining mappings for the `std::function` instance itself, like 
```java
@Name("std::function<something>") class SomeFunction {
    @Name("operator()") public native void call(...);
    ...
}
```
And to access generated types with it, either as part of a helper class or with some `Info.javaText`: 
https://github.com/bytedeco/javacpp/wiki/Mapping-Recipes#mapping-a-declaration-to-custom-code
https://github.com/bytedeco/javacpp/wiki/Mapping-Recipes#writing-additional-code-in-a-helper-class

ngraph/CMakeLists.txt.patch
```diff
@@ -0,0 +1,11 @@
+--- CMakeLists.txt.ORIG	2018-12-01 16:28:12.228687268 -0500
++++ CMakeLists.txt	2018-12-01 16:28:28.676543930 -0500
+@@ -195,7 +195,7 @@
+     endif()
+ 
+     add_dependencies(cpu_backend ext_mkldnn ext_eigen)
+-    target_link_libraries(cpu_backend PUBLIC ngraph libmkldnn libeigen libjson libtbb)
```
saudet: Wouldn't we need TBB here?

saudet: Ah, I see, it's commented out in the build script. Let's clean this up!

ngraph/cppbuild.sh
```diff
@@ -0,0 +1,94 @@
...
+#mkdir ../../include/onnx
+
+cp -r ../src/ngraph/frontend/onnx_import/core/ ../../include/
+cp -r ../src/ngraph/frontend/onnx_import/utils/ ../../include/
+cp -r ../src/ngraph/frontend/onnx_import/utils/ ../../include/ngraph/frontend/onnx_import/op/
+cp -r ../src/ngraph/frontend/onnx_import/exceptions.hpp ../../include/
+
+
+patch ../../include/ngraph/frontend/onnxifi/backend.hpp ../../../../backend.hpp.patch
+patch ../../include/ngraph/frontend/onnxifi/backend_manager.hpp ../../../../backend_manager.hpp.patch
```
saudet: Is it JavaCPP that is failing without those patches? If that's the case, let's try to fix it instead of applying those.

pom.xml
```diff
@@ -1241,6 +1243,7 @@
         <module>tensorflow</module>
         <module>ale</module>
         <module>onnx</module>
+        <module>ngraph</module>
```
saudet: Let's remove it from the list for Mac here, unless it's supposed to work already?

PR #647, opened 2018-11-28
OpenCV: Added support for nD to AbstractMat

HGuillemet:
AbstractArray.createIndexer doesn't support Mat with more than 2 dimensions.
Added AbstractMat.createIndexer that does.
The channels are considered as an additional dimension in the indexer, but this dimension
can be omitted in a long[] array of indices for mats with only 1 channel.
One limitation : Mat theoretically accepts steps that are not multiple of the element size, but indexer does not. An exception is thrown by createIndexer if this occurs.
In JavaCPP, Indexer.rows, cols, width, heads, channels should also be deprecated or modified to return -1 if the number of dimensions of the indexer is not 3.

### DIFF
--- a/opencv/src/main/java/org/bytedeco/javacpp/helper/opencv_core.java
+++ b/opencv/src/main/java/org/bytedeco/javacpp/helper/opencv_core.java
@@ -104,20 +104,6 @@ import static org.bytedeco.javacpp.opencv_core.cvScalar;
 
 public class opencv_core extends org.bytedeco.javacpp.presets.opencv_core {
 
-    public static int CV_ELEM_SIZE1(int type) {
-        switch (type & CV_MAT_DEPTH_MASK) {
-            case CV_8U:
-            case CV_8S:  return 1;
-            case CV_16U:
-            case CV_16S: return 2;
-            case CV_32S:
-            case CV_32F: return 4;
-            case CV_64F: return 8;
-            default: assert false;
-        }
-        return 0;
-    }
-
     public static abstract class AbstractArray extends Pointer implements Indexable {
         static { Loader.load(); }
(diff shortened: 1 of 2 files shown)


opencv/src/main/java/org/bytedeco/javacpp/helper/opencv_core.java
```diff
@@ -1829,6 +1843,45 @@ public static double cvNorm(CvArr arr1, CvArr arr2) {
         @Override public int arrayStep() { return step(0); }
 
         public static final Mat EMPTY = null;
+
+        @Override public <I extends Indexer> I createIndexer(boolean direct) {
+            BytePointer ptr = arrayData();
+            int size = arraySize();
+            int dims = dims();
+            int depth = depth();
+            int elemSize = CV_ELEM_SIZE1(depth);
```
saudet: I think we could just call `elemSize1()` here instead?

saudet:
BTW, `IplImage` used to support steps that are not a multiple of the element size, but I don't think `Mat` supports that anymore, because `step1()` always returns `step/elemSize1()` without throwing an exception or failing in some other way.

opencv/src/main/java/org/bytedeco/javacpp/helper/opencv_core.java
HGuillemet: Ok, I didn't spot this method. It'll trigger a JNI call but I guess we don't care here.
CV_ELEM_SIZE1 definition can be deleted then.

HGuillemet:
I think it may theoretically still happen. But in practice that would be a very weird data layout. I'd keep the check and the exception thrown.

PR #649, opened 2018-11-30
Modular Layout

HGuillemet:
Modified master pom and opencv sub-module to use the new package-targetting feature and allow modularisation.


saudet:
This is looking good! Let's merge it after we get `module-info` "classes" in place so that people can start testing this with `jlink` on the snapshots?

HGuillemet:
Adding of module-info now works for both java and native jars, using moditech. Snapshot version of moditech is needed (see https://github.com/moditect/moditect/issues/75), until beta2 is released.
The extra execution of the compiler with a jdk > 8 is not needed anymore.

HGuillemet:
sur travis ? J'ai l'impression que c'est le snapshot moditect qui est compilé avec java 8+ et sur la plateforme qui passe pas tu utilises le jdk 7

saudet:
So, we'll need to update that! Can it still compile code for Java 7 though?

HGuillemet:
moditect doesn't compile with java 7. Is Java 8 available on ppc64le ? I guess it's not a problem to request java 8 pour the building of presets.

saudet:
I'm sure it's available, but what I want to do know is, does it produce class files that are incompatible with Java 7? If we can't create libraries for Java 7 with ModiTect, we won't be able to use it.

HGuillemet:
moditect only creates the module-info.class, so what problem are you thinking about ?
What is blocking Travis is that maven is run with JDK 7 on ppc64le and that the moditect maven plugin uses lambda expressions and needs Java 8 to run. So maven must run with JDK 8. This is a problem related to the building tools only.

saudet:
Ok, sounds good then. Could you try to update to JDK 8 for linux-ppc64le as well?

opencv/pom.xml
```diff
@@ -70,41 +71,47 @@
...
-                <exclude>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/share/**/*.dylib</exclude>
-                <exclude>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/share/**/*.so*</exclude>
-                <exclude>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/x86/**/*.lib</exclude>
-                <exclude>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/x86/**/*.dll</exclude>
-                <exclude>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/x86/**/*.exe</exclude>
-                <exclude>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/x64/**/*.lib</exclude>
-                <exclude>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/x64/**/*.dll</exclude>
-                <exclude>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/x64/**/*.exe</exclude>
-                <exclude>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/**/*cascade*</exclude>
+                <exclude>lib/*.so*</exclude>
```
saudet: I don't think we need to change any of those directories anymore. Could you try to revert all those changes back?

HGuillemet: Right. Done.

saudet: It's still the same. Let me know when you've reverted the paths here.

pom.xml
```diff
@@ -164,7 +165,7 @@
             </goals>
             <configuration>
               <includes>
-                <include>org/bytedeco/javacpp/presets/*.java</include>
+                <include>org/bytedeco/javacpp/**/*_presets.java</include>
```
saudet: I think we can simplify this by just not compiling what's in `${project.basedir}/src/gen/java`. Can you try that?

opencv/pom.xml
saudet: Can you either revert the changes to these excludes, or remove them if you find that they are no longer necessary?

opencv/src/main/java9/module-info.java
```diff
@@ -0,0 +1,41 @@
+module org.bytedeco.javacpp.opencv {
```
saudet: We'll also need to add `org.opencv` to the list.

opencv/src/main/java/org/bytedeco/javacpp/cvkernels.java
```diff
@@ -1,98 +0,0 @@
...
-import static org.bytedeco.javacpp.Loader.*;
-import static org.bytedeco.javacpp.opencv_core.*;
-
-/**
- *
- * @author Samuel Audet
- */
-@Properties(inherit=opencv_core.class, value={
-    @Platform(define={"MAX_SIZE 16", "CV_INLINE static inline"}, include="cvkernels.h", compiler="fastfpu") })
-public class cvkernels {
```
saudet: We'll need to add back this file as well. It's a few custom operations that I've added before that are used in JavaCV...

pom.xml
```diff
@@ -342,21 +338,66 @@
...
+              <archive>
+                <manifestEntries>
+                  <MultiReleaseOutput>true</MultiReleaseOutput>
+                </manifestEntries>
+              </archive>
               <excludes>
-                <exclude>org/bytedeco/javacpp/windows-*/*.exp</exclude>
-                <exclude>org/bytedeco/javacpp/windows-*/*.lib</exclude>
-                <exclude>org/bytedeco/javacpp/windows-*/*.obj</exclude>
+                <exclude>/windows-*/*.exp</exclude>
```
saudet: Same here, are these excludes still required? I think they are, so let's at least revert the changes so they are effective.

pom.xml
```diff
@@ -285,8 +286,9 @@
               <goal>build</goal>
             </goals>
             <configuration>
+              <outputDirectory>${project.build.directory}/native/org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}</outputDirectory>
```
saudet: This isn't going to work for Android though. We need to place the libraries in `${javacpp.platform.library.path}` for that platform. We should be able to use the profiles below in the pom.xml file for that though. Could you give this a try?

opencv/pom.xml
HGuillemet: Reverted

pom.xml
HGuillemet: This prevent for compiling helper classes too (that may depend on generated classes that will be generated later in the build process)

opencv/src/main/java9/module-info.java
HGuillemet: Right. Added.

pom.xml
HGuillemet: Ok. Reverted.

pom.xml
saudet: The helper classes are not part of the generated classes. It's not very important, but if the generated classes contain syntax errors, this prevents the build from passing until we erase them manually, so we shouldn't try to compile them at this point.

pom.xml
saudet: Should be enough, let's see what that gives, thanks!

`javacpp.platform.library.path` isn't available for any other platform. If neither it nor `outputDirectory` is set, the effective output directory gets based on the scanned classes, so we can't really set something like `javacpp.platform.library.path` before actually parsing everything in those cases.

pom.xml
HGuillemet: We don't. At this point `compileSourceRoot` is not set, so defaults to `src/main/java` and the `includes` limits the source file to the presets file. Thus neither the generated classes nor the helper classes are compiled. I don't understand what's wrong in this configuration.

saudet: Ah, yes, it's not part of the same execution. I got mixed up. It looks alright.

pom.xml
```diff
@@ -342,12 +339,13 @@
             </goals>
             <configuration>
               <classifier>${javacpp.platform}${javacpp.platform.extension}</classifier>
-              <skipIfEmpty>true</skipIfEmpty>
-              <includes>
-                <include>${javacpp.platform.library.path}/</include>
-                <include>${javacpp.platform}${javacpp.platform.extension}/</include>
-                <include>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/</include>
-              </includes>
+              <classesDirectory>${project.build.directory}/native</classesDirectory>
```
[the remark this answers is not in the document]
HGuillemet: `native/` will only contain the output of the javacpp compiler. So no more need to filter out .java and various outputs from other building stage and that's why I removed it. But if you think normal situations may arise where, after some crosscompiling and no cleaning up, there are compiler outputs for different platforms in the directory, we might bring this includes back.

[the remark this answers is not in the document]
HGuillemet: Ok, added `<include>${javacpp.platform.nativeOutputPath}/</include>`, that should be enough.

pom.xml
```diff
@@ -392,6 +437,7 @@
             <configuration>
               <minmemory>128m</minmemory>
               <maxmemory>1024m</maxmemory>
+              <sourcepath>${project.basedir}/src/gen/java:${project.build.sourceDirectory}</sourcepath>
```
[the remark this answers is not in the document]
HGuillemet: I don't understand how maven retrieves the `compileSourceRoots` from a specific execution configuration of the compiler, but it seems to do. I get the warnings too when I remove `sourcepath`. I suggest to add the path to org.opencv source path by overriding `sourcepath` in `opencv/pom.xml`, just like `compileSourceRoots` is overridden to add this path.

saudet:
Let's not add any extra settings if we don't need to. Does it work better
if we specify the directories manually? If not, let's remove it!

HGuillemet:
Yes, no more warnings.

HGuillemet:
> Have you also tried changing the compileSourceRoots in the same order?

Yes, no changes.

Issue #593, opened 2018-07-28
Possible to do a headless packaging of OpenCV?


harukizaemon:
I'm running on a system with no GTK which is needed by highgui (and probably others). However, I have no need to run any GUI and getting GTK installed on the target system is at present, not possible.

```
Exception in thread "main" java.lang.UnsatisfiedLinkError: no jniopencv_highgui in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1122)
	at org.bytedeco.javacpp.Loader.loadLibrary(Loader.java:1225)
	at org.bytedeco.javacpp.Loader.load(Loader.java:983)
	at org.bytedeco.javacpp.Loader.load(Loader.java:882)
	at org.bytedeco.javacpp.opencv_highgui.<clinit>(opencv_highgui.java:15)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at org.bytedeco.javacpp.Loader.load(Loader.java:941)
	at org.bytedeco.javacpp.Loader.load(Loader.java:898)
	...
Caused by: java.lang.UnsatisfiedLinkError: some.jar/org/bytedeco/javacpp/linux-x86_64/libjniopencv_highgui.so: libgtk-x11-2.0.so.0: cannot open shared object file: No such file or directory
	at java.lang.ClassLoader$NativeLibrary.load(Native Method)
	at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1941)
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1824)
	at java.lang.Runtime.load0(Runtime.java:809)
	at java.lang.System.load(System.java:1086)
	at org.bytedeco.javacpp.Loader.loadLibrary(Loader.java:1205)
	...
```

When I remove all references to highgui from the generated jar, I predictably get a different error:

```
Exception in thread "main" java.lang.ArrayStoreException: sun.reflect.annotation.TypeNotPresentExceptionProxy
	at sun.reflect.annotation.AnnotationParser.parseClassArray(AnnotationParser.java:724)
	at sun.reflect.annotation.AnnotationParser.parseArray(AnnotationParser.java:531)
	at sun.reflect.annotation.AnnotationParser.parseMemberValue(AnnotationParser.java:355)
[...]

PR #667, opened 2019-01-09
Support building OpenCV headless

harukizaemon:
Addresses #593: If the environment variable `HEADLESS=yes` then we will build OpenCV without QT, GTK, and WIN32UI.


saudet:
Looks good, but unfortunately breaks the normal builds:
> ../cppbuild.sh: line 56: HEADLESS: unbound variable

https://travis-ci.org/bytedeco/javacpp-presets/jobs/477373066

harukizaemon:
> Looks good, but unfortunately breaks the normal builds:

That's totally my bad. I only ran the OpenCV build so I didn't see the `set -eu`.

I have pushed another change that assigns default value of `no`. I'm sure there are other ways to handle this but I liked that this makes the default behaviour pretty clear.

harukizaemon:
I noticed the AppVeyor build was cancelled. The Travis CI build now passes. Is there something else I need to do?

saudet:
No, looks fine! Thanks

PR #668, opened 2019-01-10
Set compile source roots using build-helper plugin

HGuillemet:
Currently, the pom uses `<compileSourceRoots>` in the configuration of the default execution of the maven compiler to add the generated sources and the possible java sources included in the library distribution.
This configuration seems to remain local to the compiler and is not used by the javadoc and maven-source plugins.
I suggest to replace this (quite undocumented) config by the use of the `add-source` goal of `build-helper-maven-plugin`.
This is yet another plugin to add to each pom, but it seems cleaner and prevent the reconfiguration of the source paths for javadoc and maven-source.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,26 +143,6 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>build-helper-maven-plugin</artifactId>
-        <version>3.0.0</version>
-        <executions>
-          <execution>
-            <id>add-source</id>
-            <phase>process-sources</phase>
-            <goals>
-              <goal>add-source</goal>
-            </goals>
-          </execution>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -142,6 +142,26 @@
...
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <version>3.0.0</version>
+        <executions>
+          <execution>
+            <id>add-source</id>
+            <phase>generate-sources</phase>
```
saudet: We should be adding the generated sources after actually generating them, so let's try something after `generate-sources`, like `process-sources` maybe?

HGuillemet: ok

PR #674, opened 2019-01-18
Adding Qt presets

freastro:
Basic presets for QtCore, QtGui, and QtWidgets. Only tested on macOS.


saudet:
Looks great! Before merging this though, could we add a build for Qt itself inside cppbuild.sh to avoid having to rely on the system for a specific version of Qt?

freastro:
I've made the requested changes and added some documentation.

PR #679, opened 2019-01-28
mod to leptonica cppbuild to all it to build properly on linux_arm64.…

kyle-redyeti:
… Still fails to build tesseract


saudet:
This pull request doesn't include any changes. Maybe you picked the wrong branch?

kyle-redyeti:
Probably, I decided to try making a forked version so I could then make the
pull request that way.

So an issue I realized that there are some limitations for versions of
tesseract and leptonica that run on Ubuntu 16.04. So I can use 3.04
tesseract and 1.71 Leptonica. Currently Nvidia does not support 18.04 on
the Jetson TX2 so I may put this on the back burner for a bit.  I do still
want to get librealsense for the D435 working through opencv but that looks
to be a bit tricky too.

I will get my initial code for tesseract pushed shortly I hope.

Thanks,

Kyle

saudet:
Leptonica and Tesseract build just fine now on Ubuntu 16.04 (Xenial) for linux-arm64:
https://travis-ci.org/bytedeco/javacpp-presets/jobs/551346066
Please give those binaries a try and let me know if there are any issues with them.
Instructions to download and use them are here: http://bytedeco.org/builds/ Thanks!

PR #681, opened 2019-01-31
Initial commit for JPMS support

HGuillemet:
Contains modification to the poms for the new package layout, modularized opencv and ffmpeg.

I didn't include the opencv generated Java, since if I pushed mine, the cuda related classes would be missing.


saudet:
It's looking good, thanks! Though, it seems like the formatting fixes I made to the abstract helper classes have been reverted. Could you undo those reverts, that is, not change the formatting? There's also a lot of missing `@Properties(inherit = ...)` annotations on the helper classes; we'll need to add those back.

saudet:
Now let's give this a try with the 1.5-SNAPSHOT builds!
http://bytedeco.org/builds/

/cc @SamCarlberg

PR #686, opened 2019-02-08
Fix regression in op schema attributes access

EmergentOrder:
I noticed this regression when building ONNX-Scala (via ONNXAlgebraGenerator), which depends on op schema attributes for API generation.


onnx/src/main/java/org/bytedeco/javacpp/presets/onnx.java
```diff
@@ -87,17 +87,18 @@ public void map(InfoMap infoMap) {
...
                              "onnx::Attributes<onnx::Node>", "Symbol", "std::reverse_iterator<onnx::ArrayRef<onnx::Node::Value*>::iterator>",
                              "const_graph_node_list_iterator", "const_graph_node_list", "onnx::toString", "onnx::ResourceGuard", "onnx::GraphInferencer",
                              "onnx::shape_inference::GraphInferenceContext", "onnx::optimization::FullGraphBasedPass",
                              "onnx::optimization::ImmutablePass", "PROTOBUF_INTERNAL_EXPORT_protobuf_onnx_2fonnx_2dml_2eproto",
                              "PROTOBUF_INTERNAL_EXPORT_protobuf_onnx_2fonnx_2doperators_2dml_2eproto", "google::protobuf::FindAllExtensions").skip())
                .put(new Info("onnx::shape_inference::InferenceContextImpl").skip())
                .put(new Info("std::set<int>").pointerTypes("IntSet").define())
-               .put(new Info("std::map<std::string,std::unique_ptr<onnx::optimization::OptimizePass> >", "std::unique_ptr<onnx::optimization::OptimizePass>").skip())
+               .put(new Info("onnx::optimization::GlobalPassRegistry").skip()) //A trade-off to get schema attributes back
```
saudet: Why does skipping this brings attributes back?

EmergentOrder: I wasn't sure exactly, but it looks like skipping `std::map<std::string,std::unique_ptr<onnx::optimization::OptimizePass> >`
also skipped `std::map<std::string, Attribute>&` ?
And GlobalPassRegistry used the former.

Attributes disappeared in https://github.com/bytedeco/javacpp-presets/pull/622. 

Turns out `OptimizePass` was also renamed to just `Pass`, but that just happened in ONNX 1.4.0.

Anyway, I was able to restore both GlobalPassRegistry and the schema attributes.

saudet:
Do you have a good reason to change the name of `std::optimization::Pass`? Let's just leave it as `Pass`...

PR #693, opened 2019-02-25
Replace jar GAV by jar filename in moditect config

HGuillemet:
Ensure moditect adds the module-info to the target jar, and not a jar fetched from maven repository.
Also commented out an <include> that seems useless.

### DIFF
--- a/cuda/pom.xml
+++ b/cuda/pom.xml
@@ -81,12 +81,7 @@
             <configuration>
               <modules>
                 <module>
-                  <artifact>
-                    <groupId>${project.groupId}</groupId>
-                    <artifactId>${project.artifactId}</artifactId>
-                    <version>${project.version}</version>
-                    <classifier>${javacpp.platform}-redist</classifier>
-                  </artifact>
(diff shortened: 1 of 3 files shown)


saudet:
The include isn't useless, you removed it also before, see one of your previous pull requests for an explanation. If it's not clear why it's there, let's add a comment.

Thanks for simplifying the Moditect config. Could you do the same for cuda/pom.xml and mkl/pom.xml? Thanks!

HGuillemet:
> The include isn't useless, you removed it also before, see one of your previous pull requests for an explanation. If it's not clear why it's there, let's add a comment.

Hmm, yes a comment will be welcome, since I don't remember why it's here and I didn't find reference for it in previous PR. Do you have an idea ? What file in `${project.build.directory}/native` and not in `${javacpp.platform.nativeOutputPath}/` would we want to exclude from the jar with this include ?
For opencv on linux at least, the generated jar is the same with and without this.

Another commit coming for replacing GAV in mkl and cuda.

saudet:
Try to build for multiple platforms, such as Android, without `clean`ing between each build.

PR #696, opened 2019-02-27
Modularization of */platform

HGuillemet:


opencv/platform/pom.xml
```diff
@@ -122,6 +122,48 @@
...
+            <phase>package</phase>
+            <goals>
+              <goal>add-module-info</goal>
+            </goals>
+            <configuration>
+              <modules>
+                <module>
+                  <file>${project.build.directory}/${project.artifactId}.jar</file>
+                  <moduleInfoSource>
+                    module org.bytedeco.opencv.platform {
```
saudet: Will adding all those "requires" prevent us from including only 1 platform? If we're not able to make them optional, we'll need to leave them out.

HGuillemet: If someone wants to pull only 1 platform, say linux.x86_64, [s]he could just require`org.bytedeco.opencv.linux.x86_64` instead of `org.bytedeco.opencv.platform`.

saudet: Sure, but what if they also want Leptonica, Tesseract, etc as part of JavaCV?

HGuillemet: Don't you have a javacv-linux-x86_64 artifact ?

HGuillemet: Apparently not. So how do you select a single platform of javacv, I don't see the use case.

saudet: With the Maven profiles, for example `mvn -Djavacpp.platform=linux-x86_64 ...`

saudet: Unfortunately, this scheme does not appear to work for JPMS. It doesn't work for Gradle or sbt either and people are not complaining too much, so until JPMS becomes more flexible, let's just include the lot here as well. Let's not include the -gpu and -redist ones though: They are a bit too big.

PR #706, opened 2019-03-26
Disable NGRAPH_USE_PREBUILT_LLVM; Fixes UnsatisfiedLinkError at runti…

EmergentOrder:
…me on LLVM via libcodegen


saudet:
When does that happen? I'm not getting that...

EmergentOrder:
I was getting a similar error on MKL when trying to use the nGraph preset until your recent fix. Once past that, I got this error. Tried deleting my local ivy and maven caches, got the latest snapshots, same. On latest Ubuntu LTS.
Building locally with this setting changed got me a successful run  (running the "free" subproject of [ONNX-Scala](https://github.com/EmergentOrder/onnx-scala) ), but I will double check that my previous local build(s) without it wasn't catching old snapshot jars or something.

saudet:
Works for now, but eventually, ideally, we should get nGraph to use the binaries prebuilt with the presets for LLVM, MKL-DNN, and ONNX.

PR #716, opened 2019-04-16
Fix missing exports

HGuillemet:


saudet:
Wouldn't a general `exports org.bytedeco.<moduleName>.presets` do just as well?

HGuillemet:
Rather than exporting to javacpp only ? 
Sure, it's just a matter of encapsulation and if you consider presets.* to be part of user API.

saudet:
It's sometimes required, yes, as discussed previously: https://github.com/bytedeco/javacpp/pull/252#issuecomment-439584607

Issue #717, opened 2019-04-19
OpenCV `Vec4iVector` should be built on `IntPointer`


jpsacha:
In C++ code `Vec4i` is defined as a vector of integers:
```c++
typedef Vec<int, 4> cv::Vec4i
```
In `org.bytedeco.opencv.presets.opencv_imgproc` (line 53) it is defined based on `FloatPointer`:
```java
new Info(..., "cv::Vec4i", ...).cast().pointerTypes("FloatPointer"))
```
it should be rather:
```java
new Info("cv::Vec4i").cast().pointerTypes("IntPointer"))
```
A problem with current implementation can be seen when using `HoughLinesP`. It will produce lines' coordinates as Vec4i. Those coordinates of lines cannot be correctly transferred from the C++ side.

saudet:
Yes, that's an error that slipped there. In any case, to work around that with the current release, we can easily cast to `IntPointer` to access it properly.

jpsacha:
Not clear how workaround with casting to `IntPointer` should be done. 

Here is some sample code that is using `HoughLinesP` to get line fragments(`lines`):
```java
Vec4iVector lines = new Vec4iVector()
HoughLinesP(binary, lines, deltaRho, deltaTheta, minVotes, minLength, minGap)
FloatPointer l = lines.get(i)
float x1 = l.get(0)
```
Due to the bug, the value of `x1` is not extracted correctly. Where the casting to `IntPointer` should happen?

saudet:
Something like this:
```
FloatPointer l = new IntPointer(lines.get(i));
int x1 = l.get(0);
```

jpsacha:
The workaround works. I added it to the LineFinder example.

saudet:
Fix included in version 1.5.1. Thanks for reporting!

PR #718, opened 2019-04-19
Correct generation of Vec4iVector

jpsacha:
Fix for #717


saudet:
Actually, I don't think we need it there at all because it's already here: https://github.com/bytedeco/javacpp-presets/blob/master/opencv/src/main/java/org/bytedeco/opencv/presets/opencv_core.java#L360
Can you check that?

jpsacha:
`HoughLinesP` is declared with `Vec4iVector`, see:
https://github.com/bytedeco/javacpp-presets/blob/c0ccd115c19bac69b78cf298a87f67fe1caba80e/opencv/src/main/java/org/bytedeco/opencv/presets/opencv_imgproc.java#L69

The line you pointed declares `Scalar4i`
https://github.com/bytedeco/javacpp-presets/blob/31ab1070fcc9ae1590344d70eacb6731349c7585/opencv/src/main/java/org/bytedeco/opencv/presets/opencv_core.java#L360
It does not resolve issue with `Vec4iVector`/`HoughLinesP`

May be `Vec4iVector` should be deleted and `HoughLinesP` declared with `Scalar4i`?

saudet:
Yes, but it should still work. Did you try to build it that way?

jpsacha:
I do not rebuild it. I am using released versions.

saudet:
Fixed in commit https://github.com/bytedeco/javacpp-presets/commit/03bc1166462eaafdba694853cb15398bfa0038ea, thanks for reporting!

PR #741, opened 2019-06-27
Strip out MKLML. MKL-DNN no longer has a dependency on it, due to its…

EmergentOrder:
… non-free license and the full MKL is already available under the same license.

"The direction that we are taking with Intel MKL-DNN though is providing full performance without dependency on non-free software (per definition of Free Software Foundation)."

See: https://github.com/intel/mkl-dnn/issues/341


saudet:
I imagine that will happen with MKL-DNN 1.0, so can you try to upgrade the version to 1.0-rc as well?

/cc @blueberry

EmergentOrder:
@saudet @blueberry
Updated to 1.0-rc

blueberry:
@EmergentOrder Thanks!

mkl-dnn/cppbuild.sh
```diff
@@ -17,45 +16,31 @@ INSTALL_PATH=`pwd`
 echo "Decompressing archives..."
 tar --totals -xf ../mkl-dnn-$MKLDNN_VERSION.tar.bz2
 cd mkl-dnn-$MKLDNN_VERSION
-patch -Np1 < ../../../mkl-dnn.patch
+#patch -Np1 < ../../../mkl-dnn.patch
```
saudet: The build fails on Mac without this patch:
https://travis-ci.org/bytedeco/javacpp-presets/jobs/552522693
We'll need to put that back.

saudet:
It's also not building on Windows, for different reasons:
https://ci.appveyor.com/project/Bytedeco/javacpp-presets/builds/25648345/job/str9gw67vf9hamhq
We can't use `long` in C/C++ because it's 32-bit on Windows. We need to use `long long` everywhere.

saudet:
Builds still failing on Mac and Windows:
https://travis-ci.org/bytedeco/javacpp-presets/jobs/553331979
https://ci.appveyor.com/project/Bytedeco/javacpp-presets/builds/25716655/job/31c14q1f0tmi7w7j

saudet:
@EmergentOrder BTW, it is possible to log into an interactive session to debug things like that. Let me know if you would like to try.

EmergentOrder:
Working now on Mac and Windows

EmergentOrder:
Looks like the nGraph build timed out.
It's working for me locally.
This is good to merge from my side, will restore the other CI builds and the MKL-DNN dep in nGraph here first if you agree.

saudet:
Ok, MKL-DNN itself seems to be building fine everywhere now, thanks! But nGraph isn't:
https://travis-ci.org/bytedeco/javacpp-presets/jobs/554029222
We'll also need to disable MKLML in MXNet and TensorFlow, but that will probably happen when they update to MKL-DNN 1.0 anyway, so we can wait for that to happen upstream before merging this PR.

EmergentOrder:
You're welcome!
Look at the build before that though, nGraph passes. Still using the older MKL-DNN preset dependency, but important because nGraph saw its own changes: 
https://travis-ci.org/bytedeco/javacpp-presets/builds/553937687

The build failure you linked to is because I bumped the MKL-DNN dependency to the latest in nGraph, and that makes Travis unable to resolve it because the new version hasn't been published. This seems to be a general problem with the CI for inter-preset dependencies (MXNet and TensorFlow fail for the same reason). Even in cases when it builds(if version didn't change), it's actually using the last published snapshot, not the one from the current PR. Maybe you know how to fix?

saudet:
Once their dependency on MKLML is removed, we can also remove the dependency on the artifact for MKL-DNN because they all compile their own version from source and link with it statically.

EmergentOrder:
Sounds good, I'll strip the dep from the nGraph preset then.

blueberry:
We seem to be very near to the final release of MKL-DNN 1.0. Please see https://github.com/intel/mkl-dnn/pull/384#issuecomment-508861754

saudet:
We still want MXNet and TensorFlow to use MKL-DNN, just not MKLML, so they need to start using MKL-DNN 1.0 as well.

EmergentOrder:
That's what I did with MXNet here. It still uses MKL-DNN, but not MKLML (on my local I double checked by looking for symbol references in the .so file). Doesn't require MKL-DNN 1.0 to do so.
It looks like in Travis it is still failing on missing MKLML though, so I'll look into that.

EmergentOrder:
Same result. For some reason `libmxnet.so` in Travis depends on `libmklml_intel.so`, but not on my local.
Any idea why that might be?

EmergentOrder:
Fixed linux-x86_64.

Similar to the inter-preset issue, the linux-x86 and gpu builds appear to fail only because they pull pre-existing binaries for the mxnet preset (which need MKLML).

saudet:
Yes, but if we're not redistributing MKLML, it wouldn't work anyway. Maybe
it's better to bring that up upstream and help them fix it there. I'm sure
if we ask someone over there could easily get this done.

EmergentOrder:
MXNet updating to MKL-DNN v1.0 and MKLML discussed here:
https://github.com/apache/incubator-mxnet/issues/15294

saudet:
I wonder if the guys at TensorFlow are going to say the same thing:
https://github.com/tensorflow/tensorflow/issues/23420#issuecomment-509882486

blueberry:
I am not really sure that I understand exactly what is the level of dependence on mainline tensorflow and mxnet for the javacpp MKL-DNN releases, but  just FYI, MKL-DNN version 1.0 has officially been released a few days ago.

saudet:
Do you guys happen to know if MKL-DNN 0.20 also doesn't really need MKLML? If so, we could indeed patch MXNet and TensorFlow to use that version and remove their dependencies on MKLML, leaving us free to upgrade to 1.0. That would work.

saudet:
You're right, MKL-DNN 0.20 still relies on MKL for a few operations. So we'll need to wait after MXNet, TensorFlow, and nGraph to update their code before we can merge this to master. It doesn't make sense to hack them to remove MKLML as it will hurt their performance.

@EmergentOrder Can you remove all modifications concerning MKLML and only leave the changes for MKL-DNN itself? Then we can merge this to a new "mkl-dnn" branch and get snapshots out at least for MKL-DNN itself.

EmergentOrder:
I have to disagree, 0.20 does not require (but does default to) MKLML.
See: https://github.com/intel/mkl-dnn/tree/rls-v0.20#using-intel-mkl-optional

 It seems it only actually uses MKL on older platforms: 

"Intel MKL-DNN includes an optimized matrix-matrix multiplication (GEMM) implementation for modern platforms. The library can also take advantage of GEMM functions from Intel MKL to improve performance with older versions of compilers or on older platforms. "

And in 1.0 you can still use MKL/MKLML if you really, really want: https://github.com/intel/mkl-dnn/blob/v1.0/src/cpu/gemm/os_blas.hpp, but default is not to.

saudet:
Documentation is rarely accurate. That's why I was asking you guys for tangible information. Look at this, for example: https://github.com/intel/mkl-dnn/blob/v0.20.1/src/cpu/rnn/rnn_reorders.hpp#L108
There are a lot of places like that in the code, but they are all gone with 1.0.

EmergentOrder:
True enough. I was wondering if you had found something like that, I took a look around but did not.

EmergentOrder:
Ok, I'll split out the changes to MKL-DNN itself to bump it to 1.0 (but that will include removing the bundling of MKLML there) to a new PR.
This one was about removing MKLML everywhere, so I will leave it open until we see the changes we need in the upstream projects.

EmergentOrder:
To track movement to MKL-DNN 1.0:

- [MXNet](https://github.com/apache/incubator-mxnet/pulls?utf8=%E2%9C%93&q=is%3Apr+%5Bmkldnn-v1.0%5D)

- [nGraph](https://github.com/NervanaSystems/ngraph/pull/3227)

- [TensorFlow](https://github.com/tensorflow/tensorflow/pulls?utf8=%E2%9C%93&q=is%3Apr+mkl-dnn+%5BINTEL+MKL%5D)

saudet:
Oh, look at that. They are renaming the library to DNNL:
https://github.com/intel/mkl-dnn
So I guess we'll just keep both the 0.x and 1.x series :)

blueberry:
@saudet Great news! If I understand correctly, now we do not have to wait for tensorflow, mxnet, etc. to release the upgrade, but can release MKL-DNN v1 AKA DNNL 1.1+ as its own independent javacpp package?

saudet:
@blueberry Yes, that's what it means, and here it is:
https://github.com/bytedeco/javacpp-presets/tree/master/dnnl

blueberry:
> @blueberry Yes, that's what it means, and here it is:
> https://github.com/bytedeco/javacpp-presets/tree/master/dnnl

Impressive!
Thank you.

EmergentOrder:
Closing this PR, the move to DNNL / off of MKLML can happen on a per preset basis in fresh PRs.

PR #765, opened 2019-07-20
Update MKL-DNN Preset to 1.0; Remove MKLML

EmergentOrder:


saudet:
Ok, this is looking pretty good, thanks! I think the only thing left before we can merge this is to update the sample based on its updated version here:
https://github.com/intel/mkl-dnn/blob/master/examples/cpu_cnn_inference_int8.cpp
@blueberry Want to work on that?

blueberry:
@saudet Yes, but this would have to wait a bit since I'm in the middle of a large pile of work. I'll also need to install the Java dev environment from scratch and switch my brain context to Java, that's why I can't just insert it in front of the queue.

blueberry:
Just a short notification that I have built the mkl-dnn branch with v1.0, and am slowly updating my existing code. So far, everything goes smooth, considering the fact that there have been many breaking changes in the mkl-dnn itself. I do not have any major issue to report so far. 

Thanks!

PR #768, opened 2019-07-25
LLVM on ARM64

loicottet:
I'd need to have the ARM64 binaries included in the org.bytedeco:llvm releases. For now I have managed to compile them locally with these changes, is there a CI I should pass to check it?


saudet:
Yes, please try to add it to the .travis.yml file.

saudet:
Hum, looks like we'll need to do a bit more to get cross-compilation working for LLVM:
https://llvm.org/docs/HowToCrossCompileLLVM.html

saudet:
@vb216 Would you like to help with that? It looks like it would work for linux-armhf as well, so we can start there...

vb216:
Looks an interesting one, but too many other commitments on at the moment to offer much help right now

loicottet:
Great! Thanks a lot!

saudet:
@loicottet BTW, I see that you've been using these bindings to implement an LLVM backend for Graal's Substrate VM. That's really great! I'm happy to see that JavaCPP is useful for the GraalVM project. If you think of anything else that I can do to help with this or anything else, please let me know!

loicottet:
@saudet The JavaCPP bindings have definitely been a great help in this project! Being able to target ARM is a big milestone for us and we'll probably need your help when targeting new architectures in the future, I'll let you know!

PR #787, opened 2019-09-02
Androïd API number in cppbuild.sh.

jbigdata-git:
Add the Android API release as a parameter of the _cppbuild.sh_ command line, i.e. _-api 14_.

- This parameter has sens only with Android platform, use ANDROID_API in _cppbuild.sh_, value comes from command line option

- Replace -D__ANDROID_API__=21 by -D__ANDROID_API__=$ANDROID_API

- Add API number in the end jar, i.e. _leptonica-1.78.0-1.5.1-android-arm64.jar_ -> _leptonica-1.78.0-1.5.1-android-arm64-**api14**.jar_

So every platform are covered and for Android every OS release.


saudet:
If we just set the API level to 14, the builds fail. We need to provide patches to work around some of the limitations of old versions of the API.

jbigdata-git:
", i.e. -api 14." 14 was an example.
Looking at [https://search.maven.org/search?q=tesseract](https://search.maven.org/search?q=tesseract) there's no file  leptonica-1.78.0-1.5.1-android-arm64-api**21**.jar

This request can't solve the limitations of old version, it is almost for now (around 20) and future API releases. Its aim is to provide severals API release of arm64 jars.

PR #793, opened 2019-09-25
initial preset for imbs-mt

netbrain:


saudet:
This looks good in general, but it's missing some sample code in the `README.md` file and in the `samples` subdirectory, so I can't test it. Could you add something there?

Could you also try to replace all the values in `@Platform(inherit = ...)` with just `opencv_features2d.class`? That seems to work fine here, but again, I don't have anything to test it. What error messages are you getting if you do that?

And please update your fork to OpenCV 4.1.1! We'll need to do that before we can merge. Thanks

PR #804, opened 2019-10-13
Bump nGraph to 0.26.0; Use MKL-DNN v1.0; Use OpenBLAS preset in place…

EmergentOrder:
… of mklml-intel

### DIFF
--- a/ngraph/cppbuild.sh
+++ b/ngraph/cppbuild.sh
@@ -65,7 +65,7 @@ $CMAKE -DCMAKE_INSTALL_PREFIX=$INSTALL_PATH -DCMAKE_INSTALL_LIBDIR=lib -DNGRAPH_
 make -j $MAKEJ
 make install/strip
 
-cp onnx/bin/libonnxifi.so ../lib/
+cp onnx/bin/libonnxifi.* ../lib/
 cp onnx/src/onnx/onnxifi.h ../include/ngraph/frontend/onnxifi/
 sedinplace  's/#include <onnxifi.h>/#include "onnxifi.h"/g' ../include/ngraph/frontend/onnxifi/backend_manager.hpp
 sedinplace  's/#include <onnx\/onnxifi.h>/#include "onnxifi.h"/g' ../include/ngraph/frontend/onnxifi/backend_manager.hpp
--- a/ngraph/cppbuild.sh
+++ b/ngraph/cppbuild.sh
@@ -14,7 +14,7 @@ if [[ $PLATFORM == windows* ]]; then
 fi
 
 NCURSES=6.1
-NGRAPH=0.25.0
+NGRAPH=0.26.0
 download https://ftp.gnu.org/pub/gnu/ncurses/ncurses-$NCURSES.tar.gz ncurses-$NCURSES.tar.gz
 download https://github.com/NervanaSystems/ngraph/archive/v$NGRAPH.tar.gz ngraph-$NGRAPH.tar.gz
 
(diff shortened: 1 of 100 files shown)


saudet:
MKLML was only there for MKL-DNN. AFAIK, nGraph itself doesn't need BLAS, right?

ngraph/cppbuild.sh
```diff
@@ -44,13 +44,28 @@ patch -Np1 < ../../../ngraph.patch
...
+sedinplace 's/CopyMKL//g' cmake/external_mkldnn_v1.cmake
+
 sedinplace '/NGRAPH_VERSION_PARTS/d' CMakeLists.txt
-$CMAKE -DCMAKE_INSTALL_PREFIX=$INSTALL_PATH -DCMAKE_INSTALL_LIBDIR=lib -DNGRAPH_UNIT_TEST_ENABLE=FALSE -DNGRAPH_TOOLS_ENABLE=FALSE -DNGRAPH_ONNX_IMPORT_ENABLE=ON -DNGRAPH_ONNXIFI_ENABLE=TRUE .
+$CMAKE -DCMAKE_INSTALL_PREFIX=$INSTALL_PATH -DCMAKE_INSTALL_LIBDIR=lib -DNGRAPH_USE_LEGACY_MKLDNN=FALSE -DNGRAPH_UNIT_TEST_ENABLE=FALSE -DNGRAPH_TOOLS_ENABLE=FALSE -DNGRAPH_ONNX_IMPORT_ENABLE=ON -DNGRAPH_ONNXIFI_ENABLE=TRUE .
 make -j $MAKEJ
 make install/strip
 
-cp onnx/bin/libonnxifi.* ../lib/
+cp onnx/bin/libonnxifi.so ../lib/
```
saudet: This won't work on Mac. Don't change what you don't need to change.

ngraph/src/main/java/org/bytedeco/ngraph/presets/ngraph.java
```diff
@@ -91,8 +92,9 @@
 //        "core/tensor.hpp",
         "ngraph/frontend/onnx_import/onnx.hpp"
     },
-    preload = {"iomp5", "mklml", "mklml_intel", "mkldnn", "ncurses@.6", "onnxifi"}, preloadresource = "/org/bytedeco/mkldnn/",
-    link = {"ngraph", "onnxifi-ngraph", "codegen", "tbb@.2", "cpu_backend"}
+    preload = {"iomp5"}, preloadresource = "/org/bytedeco/mkldnn/",
+    link = {"mkldnn", "ncurses@.6", "onnxifi", "ngraph", "onnxifi-ngraph", "codegen", "tbb@.2", "cpu_backend"}
```
saudet: This also won't work on Mac. Again, don't change what you don't need to change.

saudet:
Actually, it looks like MKL-DNN 1.0/DNNL 1.1 need the full version of MKL for some operations:
```
                    << "Dependency on Intel(R) MKL version 2019u2 or newer is "
                       "required for int8 RNN"
```
https://github.com/intel/mkl-dnn/blob/master/examples/cpu_rnn_inference_int8.cpp#L345

Should we make the presets for DNNL and thus nGraph depend on the full version of MKL? It sounds like what we should be doing...

/cc @blueberry

EmergentOrder:
I only see that in one example and one test, perhaps outdated?. The stated direction for MKL-DNN / DNNL is "full performance without dependency on non-free software", an aim I support / am trying to follow through on here. My understanding was they had reached that aim prior to the transition to DNNL. Just to point out, we don't actually use the DNNL preset here (I can remove the ref), because nGraph builds against MKLDNN 1.0, not DNNL 1.1

EmergentOrder:
As for nGraph needing BLAS, see : https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/builder/matmul_bias.cpp#L84

saudet:
@EmergentOrder Sounds like a worthy goal. Could you send a pull request upstream that adds a build flag to prevent needing MKL instead of just hacking this here? :)

EmergentOrder:
Sounds good, I'll look into that.

saudet:
Yeah, no, your MKL-less patch doesn't work on Mac:
https://travis-ci.org/bytedeco/javacpp-presets/jobs/597431585
We can achieve the same effect on Linux with `-Wl,--no-undefined` if you'd like to debug this.

saudet:
nGraph hasn't done any releases since January, is it dead?

EmergentOrder:
It's not dead, still being actively developed. In fact the contribution graph shows an uptick in the last month or so. 
But their release cycle does seem to have ground to a halt. As you noted, no RC since January, and no proper release since October. 
For that and a few other reasons, I've dropped it in ONNX-Scala in favor of ORT.

EmergentOrder:
Another of those reasons, re: the aim of this PR: ORT is already MKL-free w/ the default or DNNL CPU execution providers.

PR #819, opened 2019-11-22
Additional native requires

HGuillemet:
When using JPMS, presets that depend on other presets should include a `requires` from the native module of the dependent to the native module of the dependency. 
Without such `requires` dependency native module wont be brought into the module graph.

Adding the `nativeRequires` property to simplify this.

I set this property to all presets with a maven dependency to another presets. Maybe some of these should be `static` (optional) requires instead ?


saudet:
We're getting build errors like this for at least MXNet and TensorFlow::
```
[ERROR] Failed to execute goal org.moditect:moditect-maven-plugin:1.0.0.Beta2:add-module-info (add-module-infos) on project tensorflow: Execution add-module-infos of goal org.moditect:moditect-maven-plugin:1.0.0.Beta2:add-module-info failed: (line 3,col 45) Parse error. Found "-", expected ";"
[ERROR] Problem stacktrace : 
[ERROR]   com.github.javaparser.GeneratedJavaParser.generateParseException(GeneratedJavaParser.java:10419)
[ERROR]   com.github.javaparser.GeneratedJavaParser.jj_consume_token(GeneratedJavaParser.java:10278)
[ERROR]   com.github.javaparser.GeneratedJavaParser.ModuleStmt(GeneratedJavaParser.java:5505)
[ERROR]   com.github.javaparser.GeneratedJavaParser.ModuleDeclaration(GeneratedJavaParser.java:5654)
[ERROR]   com.github.javaparser.GeneratedJavaParser.CompilationUnit(GeneratedJavaParser.java:322)
[ERROR]   com.github.javaparser.JavaParser.parse(JavaParser.java:136)
[ERROR]   com.github.javaparser.JavaParser.simplifiedParse(JavaParser.java:342)
[ERROR]   com.github.javaparser.JavaParser.parse(JavaParser.java:314)
[ERROR]   org.moditect.internal.compiler.ModuleInfoCompiler.parseModuleInfo(ModuleInfoCompiler.java:69)
```

HGuillemet:
Fixed.
Any module-info line could be specified in the `nativeRequires` property. So we could find a better name for the property. Even though I doubt there will be cases were we need to add something else than a `requires`.

PR #832, opened 2019-12-23
opencv 4.2.0 with support for cuda4dnn - (opencv/opencv#16218)

netbrain:
It's a work in  progress, but here's what i've had to do in order to create a working opencv 4.2.0 release. Some of the changes i've done i'm not sure if is the proper solution.

TODO:
- :heavy_check_mark:  Target 1.5.3/Master 
- :heavy_check_mark:  Verify that patch changes to `modules/core/include/opencv2/core/types_c.h` are correct
- :heavy_check_mark:  Reintroduce opencv_cudacodec preset (deleted it as it was the cause of incorrect generated code)

For now, merry christmas! :)

**Note on compilation**
Compiled using, [linux-x86-and-x86_64](https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments#linux-x86-and-x86_64)
```
mvn clean install -Djavacpp.platform=linux-x86_64  --projects .,openblas
mvn clean install -Djavacpp.platform=linux-x86_64 -Dmaven.javadoc.skip=true -Djavacpp.compiler.skip=false -Djavacpp.cppbuild.skip=false -Djavacpp.platform.extension="-gpu" --projects opencv
```

### DIFF
--- a/opencv/16218.patch
+++ b/opencv/16218.patch
@@ -1,0 +1,807 @@
+From ac470873e2f92c6f817e3284e53516fcd422e418 Mon Sep 17 00:00:00 2001
+From: Julien Maille <julien.maille@data-pixel.com>
+Date: Sat, 21 Dec 2019 02:07:57 +0100
+Subject: [PATCH 1/3] Enable cuda4dnn on hardware without support for half (ie.
+ compute capability < 5.3)
+
+Update CMakeLists.txt
+
+Lowered minimum CC to 3.0
+---
+ modules/dnn/CMakeLists.txt              |  4 ++--
+ modules/dnn/src/cuda/activations.cu     | 22 +++++++++++++++++++
+ modules/dnn/src/cuda/atomics.hpp        |  2 +-
+ modules/dnn/src/cuda/bias_activation.cu | 14 +++++++++++++
+ modules/dnn/src/cuda/concat.cu          |  2 ++
+ modules/dnn/src/cuda/crop_and_resize.cu |  2 ++
+ modules/dnn/src/cuda/eltwise_ops.cu     | 10 +++++++++
+ modules/dnn/src/cuda/fill_copy.cu       |  2 ++
+ modules/dnn/src/cuda/limits.hpp         |  2 ++
+ modules/dnn/src/cuda/math.hpp           | 28 +++++++++++++++++++++++++
+ modules/dnn/src/cuda/max_unpooling.cu   |  4 ++++
+ modules/dnn/src/cuda/normalize.cu       |  2 ++
+ modules/dnn/src/cuda/padding.cu         |  2 ++
+ modules/dnn/src/cuda/permute.cu         |  2 ++
+ modules/dnn/src/cuda/prior_box.cu       |  2 ++
+ modules/dnn/src/cuda/region.cu          |  2 ++
+ modules/dnn/src/cuda/resize.cu          |  4 ++++
+ modules/dnn/src/cuda/roi_pooling.cu     |  2 ++
+ modules/dnn/src/cuda/scale_shift.cu     | 12 +++++++++++
+ modules/dnn/src/cuda/slice.cu           |  2 ++
+ modules/dnn/src/cuda4dnn/csl/fp16.hpp   |  2 ++
+ 21 files changed, 121 insertions(+), 3 deletions(-)
+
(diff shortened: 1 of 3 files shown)


saudet:
I've upgraded to OpenCV 4.2.0 in the last commit, but CUDA support for the DNN module isn't enabled by default because it doesn't support compute capabilities 3.x, which is not a problem in itself, but prevents all other modules from using compute capabilities 3.x as well, which is not acceptable. When something like https://github.com/opencv/opencv/pull/16218 gets merged, we'll be able to enable it by default. We could even apply that patch in this repository right away. If that sounds good, please give it a try and update your pull request! Thanks for your time on this

netbrain:
Sweet! ill look into it next week!

saudet:
BTW, for next time, please open WIP pull requests as "draft".

netbrain:
> BTW, for next time, please open WIP pull requests as "draft".
 
So the title of the PR should have been the following?
`draft opencv 4.2.0 `?

saudet:
No, the title doesn't matter, it's something we can set when opening a PR:
https://github.blog/2019-02-14-introducing-draft-pull-requests/

netbrain:
Aha, i see. I'll remember that for next time.

netbrain:
Will the build of this PR be available on sonatype as a snapshot?

saudet:
> Will the build of this PR be available on sonatype as a snapshot?

Not before it gets merged, no. You could set up AppVeyor and/or Travis CI on your fork to do that though.

netbrain:
Works fine on my linux machine, going to compile it on windows to see if it works there.

netbrain:
So what now? I see some checks have failed on travis, but I suspect that it's been working on old codebase.

saudet:
It's kind of weird, yes. Let me try to clear the build caches and see if that helps...

netbrain:
> It's kind of weird, yes. Let me try to clear the build caches and see if that helps...

Any update on this?

saudet:
It doesn't look like the caches were the problem. Builds here are still failing:
https://travis-ci.org/bytedeco/javacpp-presets/builds/633196365
But they pass just fine on the master branch:
https://travis-ci.org/bytedeco/javacpp-presets/builds/633980200
In any case, there's something wrong, and we'll need to fix it before we can merge this.

saudet:
There's something missing from your patch. See this build log, for example:
```
can't find file to patch at input line 336
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|diff --git a/modules/dnn/src/cuda/fill_copy.cu b/modules/dnn/src/cuda/fill_copy.cu
|index 2304e423460..a056d08ca14 100644
|--- a/modules/dnn/src/cuda/fill_copy.cu
|+++ b/modules/dnn/src/cuda/fill_copy.cu
--------------------------
File to patch: 
```
https://api.travis-ci.org/v3/job/633196448/log.txt

netbrain:
Yes, thats an old issue that was fixed. When i fixed it i did a forced update so maybe travis is on a old changeset? is there a git cache that can be cleared?

appveyor seems to go green, which it shouldn't if they truly are compiling the same codebase.

saudet:
You're right. Not sure where Travis CI picked up that commit. Could you rebase on the current master branch and force push again?

netbrain:
Sure thing

saudet:
Travis CI still isn't picking up the right commit. Please rebase and force push.

saudet:
It looks like the patches don't work with some versions of `patch`... Could you try to fix this by maybe renaming manually `fill.cu` and `fill.hpp` to `fill_copy.cu` and `fill_copy.hpp`?

netbrain:
looks like https://github.com/opencv/opencv/pull/16218 is just about to be merged into master. Which in large makes this PR not so valid i guess. Will however have to wait for 4.2.1. What do you think?

saudet:
I wonder when 4.2.1 is due...

netbrain:
@saudet i'm at a loss of whats going on here.

saudet:
It's not able to find the files to patch. The renaming needs to happen after, not before.

saudet:
But if the file has already been renamed, we don't need to rename it. We have to do something about that as well...

saudet:
OpenCV 4.3.0 has been released and contains the necessary changes. Thanks anyway for the effort!

PR #839, opened 2020-01-22
Spinnaker 1.27.0.48

jpsacha:
This upgrades Spinnaker to v.1.27.0.48.

Libraries for building on Ubuntu are updated. Windows are not, since is I do not now how they are created (ones used currently they are not the original binaries from FLIR).


saudet:
The SDKs for FlyCapture and Spinnaker are now publicly downloadable so we don't need to put them somewhere else like Google Drive anymore: https://www.flir.com/support-center/iis/machine-vision/downloads/spinnaker-sdk-flycapture-and-firmware-download/
Please try to use curl with those URLs to download them!

The executables for Spinnaker don't support unattended installation, but it extracts MSI files that do support unattended installation. We just need to double click on the EXE files to get those MSI files...

saudet:
Actually, since we need to have those MSI files extracted, we'll still need to put them up somewhere like Google Drive... ugh... Ignore what I said in the first paragraph. :)

jpsacha:
@saudet How would you download directly from FLIR website. I do not see a way to get the link to the file to download programatically). For instance the file `spinnaker-1.27.0.48-Ubuntu16.04-amd64-pkg.tar.gz` (https://flir.app.boxcn.net/v/SpinnakerSDK/folder/74727114471)?

jpsacha:
@saudet Can you clarify how you think to extract the MSI files?

saudet:
After downloading an EXE file for Spinnaker from the FLIR site, double click on it, and it will extract the MSI files somewhere in your temporary directories.

jpsacha:
You actually need to let the installer run, so it can put `*.msi` in `C:\ProgramData\Package Cache` and then get them from there. 

I updated the `*.msi` downloads, waiting for tests to run...

saudet:
I'm sure the installer changes a bit from version to version. Thanks for figuring out the new version! At least we're getting MSI files somehow...

jpsacha:
The AppVeyor build failed with missing header file that was added in new version. It may be problem with cache - there is no version number in *.MSI file names. I will rename them to add version number.

saudet:
BTW, if you could cook up a `SpinnakerFrameGrabber` for JavaCV, please do!

jpsacha:
I will take a look at creating `SpinnakerFrameGrabber`

PR #841, opened 2020-02-12
Add ORT C++ API support and sample code; Use DNNL everywhere & enable…

EmergentOrder:
… execution provider at runtime


onnxruntime/cppbuild.sh
```diff
@@ -30,7 +30,33 @@ git submodule update --init --recursive --jobs $MAKEJ
 git submodule foreach --recursive git reset --hard
 patch -p1 < ../../../onnxruntime.patch
 which ctest3 &> /dev/null && CTEST="ctest3" || CTEST="ctest"
-MAKEFLAGS="-j $MAKEJ" bash build.sh --cmake_path "$CMAKE" --ctest_path "$CTEST" --config Release --use_dnnl --use_mklml --build_shared_lib
+#MAKEFLAGS="-j $MAKEJ" bash build.sh --cmake_path "$CMAKE" --ctest_path "$CTEST" --config Release --use_dnnl --build_shared_lib
```
saudet: You've commented out the execution of the build script. That's not going to work.

EmergentOrder: Clearly. Meant to uncomment that, commented it out to avoid re-building locally each time when it wasn't necessary.

saudet:
Build fails on Mac: https://travis-ci.org/bytedeco/javacpp-presets/jobs/649548227
We should try to get it working without modifying the code as much as possible...

EmergentOrder:
I agree. Most of the seds were because of parsing fails though. 
The line `explicit TypeInfo(OrtTypeInfo* p) : Base<OrtTypeInfo>{p} {}` for example fails to parse.
If you can fix those, or know another way to make fewer modifications and still have it working, feel free.
For now though, I used a workaround for sed that should fix it for mac os.

EmergentOrder:
@saudet Thanks for the assist. 
Can we make Session::Run take a ValueVector or something similar to support multiple inputs and reflect the fact that both inputs and outputs are Value* in the preallocated version of Run?

saudet:
We can add a helper method for that, but it's not like the original C++ API makes that and a lot of other things clear anyway...

saudet:
For the `Session::Run()` method specifically, if we overload it with `ValueVector` parameters, people may assume that it can resize it to fit the number of outputs, but the underlying function itself doesn't support that, so it would still be confusing to use. Anyway, we can always design at some later point in time an API that is easier to use, and that also does not add any overhead...

PR #849, opened 2020-03-01
Hyperscan 5.2.1 integration

SoulDefender:
Hi @saudet 
I have added the preset for the Intel Hyperscan regex library (https://github.com/intel/hyperscan).
Could you review the changes and let me know if the preset can be merged into the main branch?

By the way, could you help me with adding the Ragel to linux-x86_64 build, because currently, it is failing because of the Ragel missing?


saudet:
Thanks for the contribution! Before we can merge though, we'll need to get the builds to pass on the CI machines. To speed up the process, you can temporarily comment out all the other unrelated builds.

saudet:
It looks like it's now failing on Windows too with a similar error:
> CMake Error at CMakeLists.txt:97 (message):
>  Ragel state machine compiler not found

https://ci.appveyor.com/project/Bytedeco/javacpp-presets/builds/31260209/job/n0ejtgqim93urxqf

Where does that compiler need to come from?

SoulDefender:
@saudet Initially the Ragel package was fetched from http://repo.msys2.org:
`curl -L -o mingw-w64-x86_64-ragel-6.9-2-any.pkg.tar.xz http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-ragel-6.9-2-any.pkg.tar.xz`
Now, it seems, that http://repo.msys2.org is down, so I am fetching the Ragel directly with pacman (to use available mirrors):
`pacman -S --noconfirm mingw-w64-x86_64-ragel`

SoulDefender:
@saudet Looks like the build for hyperscan is fine. The general build is failing due to some external reasons.

PR #856, opened 2020-03-28
[WIP] a subset of cudf APIs [skip ci]

razajafri:
I have started the process of creating the javacpp-presets for cudf.


razajafri:
@saudet. The only code change I had to do was change the enum: bool to enum: int32_t. I have filed a bug for that against javacpp project [here](https://github.com/bytedeco/javacpp/issues/388). 

I am also unable to get the right result when I run the program. I think I am not using the Pointer class properly. Can you please take a look? This is the first step in generating the bindings for hte entire code base.

saudet:
Are you talking about `cudfSample.java`? It looks alright, but I can't say for sure without comparing it to the equivalent in C++. What does a working example look like in C++?

razajafri:
Yes I am talking about cudfSample.java. I think I have found the problem. The template functions are being skipped. Can you please show me how to translate a template function? 

``` 
class column_view_base {
 public:
 template <typename T = void>
  T const* head() const noexcept {
    return static_cast<T const*>(_data);
  }
}
```

It will be used for all the primitive types.

saudet:
Of course, that's explained here: https://github.com/bytedeco/javacpp/wiki/Mapping-Recipes#creating-instances-of-c-templates

And check how it looks like for Arrow here: https://github.com/bytedeco/javacpp-presets/blob/master/arrow/src/main/java/org/bytedeco/arrow/presets/arrow.java

Arrow is bit weird, there's no recognizable structure to the templates, but when there is, we can easily do that in for loops as I have done with, for example, Caffe: https://github.com/bytedeco/javacpp-presets/blob/master/caffe/src/main/java/org/bytedeco/caffe/presets/caffe.java

saudet:
Also, please disable AppVeyor and Travis CI until the builds actually do something, thanks.

razajafri:
How do I disable it? I've added [skip ci] to the title does that work?

I had read the docs before I messaged here. I can't seem to make that work. I will give it another shot

saudet:
The only way that I know of is to remove the .travis.yml and .appveyor.yml
files.

razajafri:
How do I let the generated cpp file not get deleted?

razajafri:
@saudet I finally figured it out.. This is now a working code that handles the unary ops only.

saudet:
> How do I let the generated cpp file not get deleted?

There's a Mojo parameter for this:
http://bytedeco.org/javacpp/apidocs/org/bytedeco/javacpp/tools/BuildMojo.html#deleteJniFiles

> @saudet I finally figured it out.. This is now a working code that handles the unary ops only.

That's cool, but still please disable AppVeyor and Travis CI, at least until you've got something to build.

razajafri:
@saudet this code builds and I want to see if the CI will pass. Is that OK or do you still want me to get rid of CI?

saudet:
If that's the case, then yes, please do update `.appveyor.yml` and `.travis.yml` accordingly. To get feedback faster, we can temporarily comment out all the other builds.

saudet:
Please apply the following patch to your branch: [patch.txt](https://github.com/bytedeco/javacpp-presets/files/4486031/patch.txt)

Also, it looks like we need Anaconda or something to build cuDF, so please add the commands required to download and install it as part of the `ci/install-travis.sh` script. Thanks!

PR #864, opened 2020-04-04
LLVM Polly optimizer support

yukoba:
This is a pull request for LLVM Polly optimizer.
This pull request depends on https://github.com/bytedeco/javacpp/pull/389 .
LLVM 10 uses C++14.

After many trials, I decided not to add Java API.
For my opinion, the public API of Polly is only command line arguments.
Many important variables are in the file scope and cannot write from outside.
Therefore, I control through LLVMParseCommandLineOptions.

LLVM uses dynamic loading on UNIX and static loading on Windows.
So I added ```-DLLVM_POLLY_LINK_INTO_TOOLS=ON``` to the Windows build option.
Also I added polly/LinkAllPasses.h . This is necessary to avoid the compiler elimination.

I tested Linux and Windows, but I have not tested on macOS and ARM.

Please see samples/MatMulBenchmark/MatMulBenchmark.java for the usage.
As I want to add pom.xml for MatMulBenchmark,
I created samples/Fac/ and samples/Clang/ and moved sample codes to there.


saudet:
LLVM 10 uses C++14, yes, but since we're mapping only the C API, it shouldn't affect anything JavaCPP is doing. Where do we need C++14? Is it just for `polly/LinkAllPasses.h`?

If LLVMParseCommandLineOptions() is the way to access Polly's functionality from the C API, that looks fine to me.

BTW, instead of using JNA, we could use JavaCPP to map that function pointer. We could also create presets for libffi itself, which unlike JNA we could manage with JavaCPP. It's a pretty small library.

/cc @Neiko2002

yukoba:
> LLVM 10 uses C++14, yes, but since we're mapping only the C API, it shouldn't affect anything JavaCPP is doing. Where do we need C++14? Is it just for polly/LinkAllPasses.h?

Yes. polly/LinkAllPasses.h includes C++14 header files.
I don't need to change clang.java to C++14 https://github.com/bytedeco/javacpp-presets/pull/864/commits/cef020cdfa4e41add1282dd1aaa9c072050efb7a#diff-f71d00b84ee3dc122ff4dc9bba03a2faR32 , but I changed it for consistency.

> instead of using JNA, we could use JavaCPP to map that function pointer

Is it possible without libffi or do I need to port libffi?
Sorry, I'm not understanding what you want to say.

saudet:
> > instead of using JNA, we could use JavaCPP to map that function pointer
> 
> Is it possible without libffi or do I need to port libffi?
> Sorry, I'm not understanding what you want to say.

Yes, with `FunctionPointer` as shown in this unit test for `strlen()`: https://github.com/bytedeco/javacpp/blob/master/src/test/java/org/bytedeco/javacpp/PointerTest.java#L87

Since we're defining a new `native` function though, we have to execute JavaCPP on that, ideally via the Maven plugin.

yukoba:
I see. This is the solution https://github.com/yukoba/javacpp-presets/commit/114bdbc6b4ab9b810746f4daf333b75dc4d45170 .
But it is complicated. Could forgive me to use JNA?

saudet:
Yes, that works. I don't think it's complicated, but let's put both versions? :) Maybe someone else will find this complicated and that will motivate them in helping making this easier to use... What do you think?

Neiko2002:
I just have read the changes without compiling them. For me the version without JNA looks better since it is adding some type-safety when calling MatMulFunction. It might be nice to have a static function in the MatMulFunction class which handles the instantiation and storing of the pointer. 

```java
public static MatMulFunction create(Pointer address) {  
  MatMulFunction func = new MatMulFunction();  
  func.put(address);  
  return func;  
}  
```

yukoba:
If I use ```Builder.build()```, is it possible to run without Visual Studio 2017 on Windows?

saudet:
> If I use `Builder.build()`, is it possible to run without Visual Studio 2017 on Windows?
> [yukoba@114bdbc#diff-fd9ea1da4821f979dec59387d963ee99R69](https://github.com/yukoba/javacpp-presets/commit/114bdbc6b4ab9b810746f4daf333b75dc4d45170#diff-fd9ea1da4821f979dec59387d963ee99R69)

No, it's not possible. That's what a library like libffi is for. I find it surprising that LLVM itself doesn't offer this kind of functionality, but it doesn't.

saudet:
BTW, the reason why your benchmark is slower than natively compiled code with clang is probably because of JNA and libffi. Those are known to be very slow. Since your goal is to increase performance, you shouldn't try to use libraries like libffi and JNA, and compile everything with a C++ compiler. That's why JavaCPP does that, it's for performance reasons.

yukoba:
> Fix build on Mac

Thank you!

> No, it's not possible. That's what a library like libffi is for.

Then, if I do not use JNA, you need the JavaCPP building environment to run this sample. I think most people do not prepare the JavaCPP building environment and impossible to run.

> BTW, the reason why your benchmark is slower 

```clang -O3 -march=native -mllvm -polly -mllvm -polly-vectorizer=stripmine```
can use Intel AVX-512 related optimization on Intel Skylake Xeon.

However, I have to add ```-march=native``` thing to here
https://github.com/bytedeco/javacpp-presets/pull/864/files#diff-fd9ea1da4821f979dec59387d963ee99R240
but I cannot find a way to do so. I searched for many hours but I cannot find.

By ```-O3 -march=native```, they convert
```s += a[m * K + k] * b[k * N + n]```
to this,
```
s += a[m * K + k] * b[k * N + n] +
a[m * K + (k + 1)] * b[(k + 1) * N + n] +
a[m * K + (k + 2)] * b[(k + 2) * N + n] +
a[m * K + (k + 3)] * b[(k + 3) * N + n];
```
but this optimization is impossible on my sample code.
Polly and LLVM do not understand that my CPU can use Intel AVX-512.

However, I think this optimization part is using Intel AVX-512.
https://github.com/bytedeco/javacpp-presets/pull/864/files#diff-fd9ea1da4821f979dec59387d963ee99R263

saudet:
Maybe we just need to enable some option for AVX-512 support? Did you look?

saudet:
That functionality may be missing from the C API: https://stackoverflow.com/questions/38738123/determining-and-setting-host-target-triple-and-instruction-extensions-in-llvm-c/38801586

yukoba:
Thank you for merging.

> That functionality may be missing from the C API:

I solved the optimization problem.
I have to do the same thing of ```main()``` in llvm/tools/opt/opt.cpp .

https://github.com/bytedeco/javacpp-presets/commit/acbfeeb43cd6e21df44ff950fc26eeda1b51fb8d

I will use this personally for a week, and if there is no problem I want to send a pull request.

saudet:
Thanks! Please consider contributing this to the C API upstream though.

saudet:
I know it's hard to contribute new code to LLVM, so if you're not able to, please do send a pull request to have this patch added to the presets here at least. Thanks!

yukoba:
>  if you're not able to, please do send a pull request to have this patch added to the presets here at least.

@saudet OK. I sent https://github.com/bytedeco/javacpp-presets/pull/869.

PR #869, opened 2020-04-16
LLVM full optimization

yukoba:
This pull request is for LLVM full optimization.

I added two functions.

1. ```JavaCPPLLVMOptimizeModule()```: This function does the standard LLVM optimization, and based on main() of llvm/tools/opt/opt.cpp.
2. ```JavaCPPLLVMCreateOptimizedJITCompilerForModule()```: This function is similar to LLVMCreateJITCompilerForModule() but does CPU specific optimization.

I added ```JavaCPPLLVM``` prefix to avoid the conflict against LLVM-C function names.


saudet:
I don't see any `LLVMOptimizeModule()` or `LLVMCreateOptimizedJITCompilerForModule()` functions.
How do those names conflict?

yukoba:
Yes, LLVM does not have ```LLVMOptimizeModule()``` and ```LLVMCreateOptimizedJITCompilerForModule()```. But if LLVM creates those functions in the future, JavaCPP has to change the name and cannot keep the upper compatibility. Therefore I added JavaCPPLLVM prefix.

saudet:
Right, but we're not adding to the C API, so we don't need to follow its conventions. We can also make it more object-oriented. Let me propose something.

yukoba:
Thank you. Please merge.

PR #900, opened 2020-06-27
Remove requires in *-platform modules

HGuillemet:
This PR allows an app to have a `*-platform` module in its module graph without bringing into the graph all existing implementations of the native library.

Meant to work with the use of `javacpp.plaform` property and the new additions to javafx-maven-plugin.

See [Add option to populate module path with all maven dependencies #91](https://github.com/openjfx/javafx-maven-plugin/issues/91)


HGuillemet:
I'm not sure if it's normal than the requires towards the macos module is commented out in the cuda presets.

saudet:
> I'm not sure if it's normal than the requires towards the macos module is commented out in the cuda presets.

Yes, CUDA no longer supports Mac, but some users may want to backport some stuff to older versions, you never know...

PR #913, opened 2020-07-23
upgrade opencv 4.4.0

n-kai-cj:
upgrade new released opencv `4.4.0`


saudet:
Thanks, but it's already been done! FYI, we need to update versions in the pom.xml file as well, see commit https://github.com/bytedeco/javacpp-presets/commit/17a1f75ee9587151b40f58066bfad74561c5313d.

n-kai-cj:
Oh I didn't notice, sorry for duplication.

PR #928, opened 2020-08-17
Update module-info.java

AKMANSoft:
helper package was missing in the file. Fix Error: module org.bytedeco.opencv does not export org.bytedeco.opencv.helper

### DIFF
--- a/cpython/src/main/java9/module-info.java
+++ b/cpython/src/main/java9/module-info.java
@@ -1,6 +1,7 @@
 module org.bytedeco.cpython {
   requires transitive org.bytedeco.javacpp;
   exports org.bytedeco.cpython.global;
+  exports org.bytedeco.cpython.helper;
   exports org.bytedeco.cpython.presets;
   exports org.bytedeco.cpython;
 }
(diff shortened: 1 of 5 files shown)


saudet:
@HGuillemet Let's append to this all other missing helper modules like that.

HGuillemet:
I wrote a script to identify source directories without a corresponding export in module-info. It found:
```
org/bytedeco/cpython/helper
org/bytedeco/opencv/helper
org/bytedeco/qt/helper
```
and the following directories, which seems to have been introduced after the creation of the module-info:
```
org/bytedeco/llvm/program
org/bytedeco/tesseract/program
```
Do you think these 5 directories correspond to packages which must be exported (ie contain Java type that could be accessed by users) ?

saudet:
Yes, they should be accessible. Thanks!

HGuillemet:
@AKMANSoft, could you add to your PR these 5 exports to their corresponding module-info ?
I don't think I can do it myself without submitting another PR.
Thanks,

AKMANSoft:
Yes no problem. I have added the exports.

Issue #932, opened 2020-08-23
Supporting non-host LLVM Targets


supergrecko:
**Supporting other Stable Targets**

As of right now (1.5.4-SNAPSHOT) the LLVM build is set to only enable the target the host we're building for<sup>(1)</sup>. Is there a specific reason we're only enabling the host target? This makes the current LLVM binary unable to cross-compile because it doesn't have the other backends.

**Supporting Experimental Targets**

As of LLVM 10.0.1 LLVM has 19<sup>(2)</sup> targets, 17<sup>(3)</sup> of which are "stable", experimental backends being AVR and ARC. If we do decide to enable all the stable targets, we should also consider providing support for the experimental backends.

We should also note that AVR has become a stable backend as of LLVM 11.0.0-rc1.

**Footnotes**

1. [In the cppbuild.sh](https://github.com/bytedeco/javacpp-presets/blob/master/llvm/cppbuild.sh#L36), the Cmake `-DLLVM_TARGETS_TO_BUILD` variable is set to `host`, or whichever target if we are cross-compiling LLVM itself. One would assume this enables the target equal to the one we're building LLVM on, but I couldn't find anything regarding `host` [in the LLVM Cmake build](https://github.com/llvm/llvm-project/blob/master/llvm/CMakeLists.txt). I do believe it is currently only enabling the targets for the host arch though, because `LLVMGetFirstTarget()` with `LLVMGetNextTarget()` only yields the platform the binary was built for.

2. List of targets https://github.com/llvm/llvm-project/tree/llvmorg-10.0.1/llvm/lib/Target

3. List of stable targets https://github.com/llvm/llvm-project/blob/llvmorg-10.0.1/llvm/CMakeLists.txt#L288

Would love to hear your thoughts about this.

saudet:
What's the size of those binaries? If they are like 10 times larger, we'll have to see if enough people need this to enable them by default.

supergrecko:
I build LLVM 10.0.1 for linux-x86_64 with all targets enabled and compared the library sizes. I used the binary for linux-x86_64 which the presets ship with for comparison.

The current `libLLVM-10.so` is 51 megabytes which has one target included.

[...]

Issue #935, opened 2020-08-24
Missing LLVM Target Initialization APIs


supergrecko:
###### Partially related to #932

In the [`llvm-c/Target.h`](https://github.com/llvm/llvm-project/blob/llvmorg-10.0.1/llvm/include/llvm-c/Target.h) header there are six macros which declare functions for target initialization. These are ran for each target to generate a set of APIs exposed to LLVM-C.

Javacpp is currently not properly read though this (probably because the `llvm/Config/Targets.def` file is created during build generation.

**Fix**

I'm not sure how we want to go about adding these APIs as they *should* only be available if the target is enabled. With the current build setup, that would be one target, but a different one for each platform we build the binaries for.

1. Creating a stub file (similar to #869) in the resources directory and add it to the preset for ALL targets
This solution means that the initialization API for every target will be "available" regardless of which target the user's binary has enabled. This is probably not wanted behavior because calling into a function which doesn't actually exist will crash the JVM.

2. Parse the `llvm-c/Target.h` file after build generator step, assuming the Javacpp parser is able to perform macro expansion.
This requires the parser to go over the files after the build has been generated.

I have created an example showing how this could be done if we go with option 1 [here](https://github.com/supergrecko/javacpp-presets/tree/llvm/targets-header-all)

saudet:
Is there a reason why the `LLVMInitializeAllTargets()`, etc functions not enough?

supergrecko:
Generally yeah, they're more than enough. I just noticed they were missing from the javacpp output and wanted to raise awareness.  There would be some unnecessary initalization if one doesn't use all the targets, but it's a very minor thing.

saudet:
The problem isn't macro expansion, it's because JavaCPP doesn't process `#include` directives, the idea being that users best know in what order they should be parsed, but it doesn't work when the same file is included multiple times in another file like this with different macro values.

saudet:
Released with version 1.5.4! Thank you for your contribution

PR #937, opened 2020-08-25
Build all stable LLVM targets

supergrecko:
As discussed in #932 I have enabled LLVM to build all stable targets.


saudet:
While we're at it, let's also include your code for issue #935? It looks alright.

saudet:
Apart from the experimental targets, it looks like the following functions are not actually defined:
```
   Creating library jniLLVM.lib and object jniLLVM.exp
jniLLVM.obj : error LNK2001: unresolved external symbol "void __cdecl LLVMInitializeXCoreAsmParser(void)" (?LLVMInitializeXCoreAsmParser@@YAXXZ)
jniLLVM.obj : error LNK2001: unresolved external symbol "void __cdecl LLVMInitializeNVPTXAsmParser(void)" (?LLVMInitializeNVPTXAsmParser@@YAXXZ)
jniLLVM.obj : error LNK2001: unresolved external symbol "void __cdecl LLVMInitializeNVPTXDisassembler(void)" (?LLVMInitializeNVPTXDisassembler@@YAXXZ)
jniLLVM.dll : fatal error LNK1120: 3 unresolved externals
```
Do we just disable them for all targets?

supergrecko:
Yeah, it seems like those targets don't export those functions for some reason. We could either omit the initialization functions for those targets entirely or we just omit those three functions manually ([example gist](https://gist.github.com/supergrecko/70c5c5c3597d58151b1488e296c362f5))

PR #946, opened 2020-09-18
Updated librealsense2 to version 2.38.1

cansik:
This PR updates the librealsense2 library to version 2.38.1 which supports newer devices like the RealSense L515 (LIDAR). There were not that many changes necessary, mainly the `libtm` which has been replaced with `tm2`.

I could just test the library on MacOS but there the new devices are recognized and an output image could be captured.

I was not sure if it's necessary to update the number `1.5.4` to `1.5.5`, but I did it. Maybe you would like to change it (back or to another version).


saudet:
Thanks for the update! Use 1.5.5-SNAPSHOT for the version until it gets released.
Before we can merge though we'll need the builds to pass. They are currently on Linux and Windows:
https://travis-ci.org/github/bytedeco/javacpp-presets/builds/728431118
https://ci.appveyor.com/project/Bytedeco/javacpp-presets/builds/35285006

saudet:
On Linux, we can get the video.h, etc files as with this patch used for librealsense:
https://github.com/bytedeco/javacpp-presets/blob/master/librealsense/librealsense.patch
For Windows, we'll need to skip the missing symbols, which are undefined anyway, so we don't lose anything.

cansik:
I will have a look at it tomorrow! Thx for the investigation.

saudet:
Actually, the patch is there for librealsense2 as well, but it's a bit
broken...

Issue #419, opened 2017-05-14
newbie question on licensing [GPL]


devdoomari:
Hello!
I'm planning to use javacpp-presets for ffmpeg for my educational app.
However, I'm a bit confused about javacpp-presets being Apache 2 while ffmpeg being GPL.

Is it OK to use javacpp-presets + ffmpeg for a closed-source app?
(ie. does javacpp-presets library dynamically-link to ffmpeg? also, is it ok to use dynamically-linked GPL-code for a closed-source app?)

I'm not sure whether this is the right place to ask, but I just want to make some informed decision before diving in.

Thank you!

saudet:
This would not work with GPL, but most of FFmpeg is LGPL, so as long as you distribute, with your application, unmodified non-GPL modules, it should be OK for use in a closed-source application, yes.

devdoomari:
so the default ffmpeg build I get when I do:

```scala
libraryDependencies ++= "org.bytedeco.javacpp-presets" % "ffmpeg" % "3.2.1-1.3"
```
is LGPL?
Or do I need to do something else to get a LGPL-build?

saudet:
It contains some GPL components...

devdoomari:
oh... ok.
then how about using version-postfix for gpl/lgpl version? 
e.g) 3.2.1-1.3-lgpl --> lgpl pre-built
       3.2.1-1.3-lgpl_no-nonfree --> lgpl without non-free
        3.2.1-1.3-gpl --> gpl pre-built

saudet:
Sure, contributions as welcome!

devdoomari:
I guess 
https://github.com/bytedeco/javacpp-presets/blob/master/ffmpeg/cppbuild.sh
this file looks like the start...
but how do you guys build javacpp-presets-ffmpeg?
e.g. via a CI (e.g. CircleCI?)?

should I add an "if git repo branch has 'LGPL' then ... / else : ..." logic?
Or would you like to have an option for the build?
(e.g. ./cppbuild.sh --lgpl ?)

saudet:
Use an environment variable, something like ENABLE_GPL=0 so we can use it
from the parent cppbuild.sh script as well, with a default of 1, along with
others like MAKEJ sounds good.

We've started to use AppVeyor and Travis CI for now...

devdoomari:
[...]

PR #949, opened 2020-09-20
Add FFmpeg without GPL module (#419)

tmurakam:
- Add ffmpeg_nogpl module. Same artifactId but changed version (-nogpl suffix)
- cppbuilds.sh: Disable gpl modules
- postproc.java: Disable postproc


saudet:
Please try to implement this as an "extension" instead. Thanks for the
contribution!

tmurakam:
Ok, I'll try it.

PR #951, opened 2020-09-23
Added support for armhf and arm64 in librealsense2

ICH88:
Added all the necessary changes to successfully compile for the armhf platform.

Can be cross-compiled with: 
mvn clean install -Djavacpp.platform=linux-armhf -Djavacpp.platform.compiler=arm-linux-gnueabihf-g++ -Dmaven.javadoc.skip=true -pl :librealsense2

(Sorry for the chaotic commits, I was in a hurry)


saudet:
While we're at it, let's do the same for librealsense 1.x!

saudet:
Let's also add builds for linux-arm64 here. BTW, you don't need to test that everything is working. We can let Travis CI do that.

ICH88:
Both armhf and arm64 are building and working correctly now.
Had no time for librealsense 1 so far.

saudet:
Thanks! It builds fine with GCC 5.4 from Ubuntu 16.04:
https://travis-ci.org/github/bytedeco/javacpp-presets/jobs/730305554
But not with GCC 4.9 from the Raspberry Tools:
https://travis-ci.org/github/bytedeco/javacpp-presets/jobs/730305543
Looks like we'll need to patch this a bit...

saudet:
BTW, are 32-bit Raspberry Pi still ARMv6 only? Or can they do ARMv7 too? If they accept ARMv7 binaries as well these days, we could cross-compile using GCC from Ubuntu instead of trying to patch code from librealsense2.

/cc @vb216

vb216:
I think the early Pi1, and Pi Zero are physically ARMv6 chips, some of the newer models running 32bit OS are ARMv7 but may report as ARMv6 still. 
Tricky one though, realistically I suspect anyone doing heavier compute work would be on a Pi3/Pi4 nowadays, but you'd lose some older platform compatibility fully switching to v7.

ICH88:
The newer Pis are V7 compatible, I actually compiled on Ubuntu and it runs fine on the ARM V7 boards

saudet:
Interesting, these days they actually recommend using Ubuntu for cross-compiling:
https://www.raspberrypi.org/documentation/linux/kernel/building.md
And it's supposed to work for "Pi 1, Pi Zero, Pi Zero W, or Compute Module". 
Not sure how that works, but let's start by making builds with Ubuntu and we'll see if anyone has any issues.

PR #955, opened 2020-10-12
Update LLVM presets to 11.0.0

supergrecko:
This updates the LLVM presets to the recently released 11.0.0 version. The version number in the README and poms have been updated from 1.5.4 to 1.5.5-SNAPSHOT. The build and generation was ran on Ubuntu 1804 on x86_64.


README.md
```diff
@@ -134,7 +134,7 @@ Each child module in turn relies by default on the included [`cppbuild.sh` scrip
  * NumPy 1.19.x  https://github.com/numpy/numpy
  * SciPy 1.5.x  https://github.com/scipy/scipy
  * Gym 0.17.x  https://github.com/openai/gym
- * LLVM 10.0.x  http://llvm.org/releases/download.html
+ * LLVM 11.0.x  http://llvm.org/releases/download.html
```
supergrecko, line 137: I suppose I could change this to 11.0.0 as this would be more precise, but I left it as is

saudet:
Thanks! Looks like it's not building on Windows for some reason. We'll need to do something about this before we can merge:
```
ninja: error: '/c/projects/javacpp-presets/.git/logs/HEAD', needed by 'include/llvm/Support/VCSRevision.h', missing and no known rule to make it
```
https://ci.appveyor.com/project/Bytedeco/javacpp-presets/builds/35711190/job/2qs8csd81x71ko2h

supergrecko:
There's a -DLLVM_APPENC_VC_REV cmake flag which toggles this header so I went ahead and disabled that because we're not pulling the source code from git, we're downloading the sources directly.

saudet:
We'll also need to update the version of the libraries:
```
/usr/lib/gcc-cross/arm-linux-gnueabihf/5/../../../../arm-linux-gnueabihf/bin/ld: cannot find -lLLVM-10
```
https://travis-ci.org/github/bytedeco/javacpp-presets/jobs/735160594

saudet:
It still doesn't build on Windows, same error:
https://ci.appveyor.com/project/Bytedeco/javacpp-presets/builds/35721871/job/ilfavo432ruj4nqv

supergrecko:
Yeah, just noticed that as well, I was surprised it still looks for it even though the option is disabled, but it looks like LLVM doesn't actually do any checks for the option and just [includes the VersionFromVCS.cmake file](https://github.com/llvm/llvm-project/blob/llvmorg-11.0.0/llvm/CMakeLists.txt#L218) which doesn't care about the option at all and determines whether we should build VCS revision based on whether the host has git installed or not (https://github.com/llvm/llvm-project/blob/master/llvm/cmake/modules/VersionFromVCS.cmake#L7)

I suppose a temporary workaround could be forging a .git/logs/HEAD file ourselves?

saudet:
> I suppose a temporary workaround could be forging a .git/logs/HEAD file ourselves?

It sounds more robust to just patch the build file if that's the problem here?

supergrecko:
I can send a patch to LLVM, but that's not going to solve our current build issues as that'd arrive in 12.0.0 unless you were thinking of manually patching the Cmake file ourselves after it's downloaded

saudet:
Yes, just remove that line from the build file. There's already a line like this here for example:
https://github.com/bytedeco/javacpp-presets/blob/master/llvm/cppbuild.sh#L29

supergrecko:
Tried to just delete the include() file, but the build wouldn't go through as the function is no longer there. Instead I have tried to replace the include arg with a function declaration which hopefully fixes the issue.

saudet:
Nope, doesn't work, still the same error: 
https://ci.appveyor.com/project/Bytedeco/javacpp-presets/builds/35748347/job/3bam10rrndwtkoi3

saudet:
BTW, if you don't have access to a Windows machine, you can debug using AppVeyor for free from your own fork:
https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

saudet:
For something a bit less involved, we can also temporarily comment out all other builds from `.appveyor.yml` to get feedback faster.

saudet:
Your latest commit also broke the build on Mac:
```
CMake Error at CMakeLists.txt:218:
  Parse error.  Expected a newline, got identifier with text "n".
```
https://travis-ci.org/github/bytedeco/javacpp-presets/jobs/735727731

We can similarly comment out lines from `.travis.yml`...

supergrecko:
Looks like that CMakeLists is just the top of the iceberg. The header file is written by the following python script https://github.com/llvm/llvm-project/blob/master/llvm/utils/gn/build/write_vcsrevision.py which is called through GN which is a meta build system for Ninja which is probably why we're only hitting this error on the windows builds because that's the ones who use Ninja as Make target.

How the python file itself is called is something I'm yet to figure out.

saudet:
Maybe a regression of https://reviews.llvm.org/D38033?

supergrecko:
Yeah, possibly, setting up a windows env to test and look for workarounds.

saudet:
It might also do the same thing on other platforms with ninja. Have you tried?

supergrecko:
It should do it on any platform, I'm surprised it only triggers when using ninja. Essentially, inside [this function](https://github.com/llvm/llvm-project/blob/master/llvm/cmake/modules/AddLLVM.cmake#L2096), it git rev-parse's to find the closest git repository and then looks for /logs/HEAD which is the error we bump into. Because we fetched sources instead of the git repo, the closest repo is javacpp-presets.

I'm fairly certain the easiest way to solve this is to either clone LLVM from source (is there a specific reason we're downloading the source tarballs instead of cloning a tag except for manually moving polly and clang?) or to forge a git repository in the LLVM source directory. Patching LLVM for this is going to be tedious as this file is only one of the multiple places the build system attempts to find a git repository. Let me know what you think

saudet:
Hum, it works fine for me locally on Windows, so it's something special that AppVeyor is doing with Git.

The reason I avoid using Git is because Windows usually has problems like that with Git.

CMake still goes looking for Git even with the modifications you made. Let's see how we can disable that entirely...

saudet:
BTW, we don't need to update this list too?
https://github.com/bytedeco/javacpp-presets/blob/master/llvm/src/main/resources/org/bytedeco/llvm/include/TargetStubs.h

supergrecko:
Hmm, I was certain I updated that file but it looks like I didn't. I'll submit a patch.

PR #971, opened 2020-10-28
Update README.md

xwaeaewcrhomesysplug:
Based on issue about overall documentation, part 1. Added a section called Usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,15 @@ Further, in the case of Android, the JavaCPP Presets also rely on:
 
  * Android SDK API 21 or newer  http://developer.android.com/sdk/
 
+Usage
+-------------------
+Javacpp presents repo documentations have a two layer concept.This documentation readme you are reading is the upper layer,or simply Javacpp layer.
+However,each packages have some extra notes about usages,setup etc.As not every project have the same code quality,style,layout.
+Each (folder or packages)will have a respective readme.
+Once again:
+This document is the upper layer,(Javacpp-presents).It will mentions about javacpp wide layer settings like java runtime.Build,etc...
+The package is the lower layer,(eg.Opencv).It will mention about code,pom setting,required binary,package related documentations.
+
 
 Manual Installation
 -------------------


saudet:
Thank you for your contribution. I'm sorry I didn't get back to you sooner! I almost forgot about this PR.

I agree with you that adding a generic "Sample Usage" section in the main README.md file would be a good thing. However, the content of that section should be clear. Could you work on the wording of your paragraph to make it easier to understand?

xwaeaewcrhomesysplug:
Sorry will need to reopen a new issue with a new repository.The old readme cannot be override.

xwaeaewcrhomesysplug:
I created a draft.If it is ok.Then you can simply close this pr and I will create a new one with these>

Usage
-------------------
Javacpp presents repo documentations uses layering concept.
There are two layer in this repo.
1)This documentation readme you are reading is the upper layer,or simply Javacpp layer.
2)The 2nd layer is those folders or packages.
Now,you read ARE reading the (upper layer/Javacpp layer) which means you should have enough info to resolve some javacpp issues.
However,each (folder or packages)will have a respective readme.Which is some extra notes about usages,setup etc.As not every 3rd party ported project have the same code quality,style,layout.

Summary:
This document is the upper layer,(Javacpp-presents).It will mentions about javacpp wide layer settings like java runtime.Build,etc...
The package is the lower layer,(eg.Opencv).It will mention about code,pom setting,required binary,package related documentations.

saudet:
Since this basically says to go look at README.md files in each subdirectory, it sounds like we should should add this to the "Introduction" section. Do you agree?

If you have something a bit more substantial though, please let me know what you have in mind! Thanks

xwaeaewcrhomesysplug:
sure any way you like,I think telling users at introduction is a valid way.Sorry for late reply,sort some personal stuff.

xwaeaewcrhomesysplug:
ok closing this pr.Will do a new pr on the introduction.

PR #980, opened 2020-12-08
add opencv build without openblas dependencies

n-kai-cj:
I'd like to add an option to build OpenCV without OpenBLAS dependencies.

- Add `exclude` argument to root `cppbuild.sh`
- Confirmed build success and run opencv without openblas.jar
  - build and run on windows-x86_64, windows-x86_64-gpu

build
`bash cppbuild.sh -platform windows-x86_64 -exclude openblas install opencv`
`mvn install -Djavacpp.platform=windows-x86_64 --projects opencv`

### DIFF
--- a/cppbuild.sh
+++ b/cppbuild.sh
@@ -45,6 +45,7 @@ case $ARCH in
 esac
 PLATFORM=$OS-$ARCH
 EXTENSION=
+EXCLUDE=
 echo "Detected platform \"$PLATFORM\""
 
 while [[ $# > 0 ]]; do
@@ -62,6 +63,13 @@ while [[ $# > 0 ]]; do
         -extension)
             shift
             EXTENSION="$1"
+            ;;
+        -exclude=*)
+            EXCLUDE="${1#-exclude=}"
+            ;;
+        -exclude)
+            shift
+            EXCLUDE="$1"
             ;;
         install)
             OPERATION=install
--- a/opencv/cppbuild.sh
+++ b/opencv/cppbuild.sh
@@ -45,7 +45,10 @@ export PYTHON3_INCLUDE_DIR=
 export PYTHON3_LIBRARY=
 export PYTHON3_PACKAGES_PATH=
 if [[ -f "$CPYTHON_PATH/include/python3.7m/Python.h" ]]; then
-    export LD_LIBRARY_PATH="$OPENBLAS_PATH/lib/:$CPYTHON_PATH/lib/:$NUMPY_PATH/lib/:${LD_LIBRARY_PATH:-}"
+    export LD_LIBRARY_PATH="$CPYTHON_PATH/lib/:$NUMPY_PATH/lib/:${LD_LIBRARY_PATH:-}"
+    if [[ "$EXCLUDE" != *openblas ]]; then
+        export LD_LIBRARY_PATH="$OPENBLAS_PATH/lib/:${LD_LIBRARY_PATH:-}"
+    fi
     export PYTHON3_EXECUTABLE="$CPYTHON_PATH/bin/python3.7"
     export PYTHON3_INCLUDE_DIR="$CPYTHON_PATH/include/python3.7m/"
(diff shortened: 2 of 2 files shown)


saudet:
Please remove from the patch the files found under src/gen. They get regenerated anyway.

n-kai-cj:
commit them.
BTW, how do I get this jar from https://oss.sonatype.org/content/repositories/snapshots/org/bytedeco/ or maven repository?
maybe I need to add something in `.appveyor.yml` and `.travis.yml`, but not sure how to add new argument `exclude`.

saudet:
How about using an environment variable like HEADLESS so that we can run the build entering a single command like
`HEADLESS=1 mvn clean install`?

The builds are now done with GitHub Actions, so you can modify the files under `.github` to have the JAR files built from your fork deployed wherever you need them.

saudet:
What about using `NOOPENBLAS` as environment variable to tweak the build? Does that sound good to you?

n-kai-cj:
sorry for late.
Yes, that's better to me.

I was looking into the way to add something to `.github` files to make jar files. Ideally, I'd like to get the opencv jar without openblas by indicating like a `opencv_noopenblas-platforms` in pom.xml. But I'm not sure how to do that.

saudet:
> I was looking into the way to add something to `.github` files to make jar files. Ideally, I'd like to get the opencv jar without openblas by indicating like a `opencv_noopenblas-platforms` in pom.xml. But I'm not sure how to do that.

The closest thing to that would be an "extension", such as the one defined in "platform/gpu/pom.xml", but this doesn't allow us to change the dependencies of the modules. There's just no way to do that with Maven... We could work around that by providing the "excludes" as part of the platform artifact in something like "platform/noopenblas/pom.xml" instead patching the main pom.xml file. I'm not sure that would work well, but if you would like to try it out, please do!

n-kai-cj:
Yes, I've just noticed it seems not to be able to be achieved. If I succeed with your comments such as adding `exclude` and `platform/noopenblas/pom.xml`, we couldn't get opencv-gpu without openblas or we need more `platform/noopenblas/gpu/pom.xml`, right? I think it's not a good way.

Second ideal thing is to get each platforms of `opencv_noopenblas.jar` from https://oss.sonatype.org/content/repositories/snapshots/org/bytedeco/. Can we change the artifact jar file's name with command `mvn install` by editing pom.xml?

saudet:
> Yes, I've just noticed it seems not to be able to be achieved. If I succeed with your comments such as adding `exclude` and `platform/noopenblas/pom.xml`, we couldn't get opencv-gpu without openblas or we need more `platform/noopenblas/gpu/pom.xml`, right? I think it's not a good way.

Right, it doubles the number of packages.

> Second ideal thing is to get each platforms of `opencv_noopenblas.jar` from https://oss.sonatype.org/content/repositories/snapshots/org/bytedeco/. Can we change the artifact jar file's name with command `mvn install` by editing pom.xml?

That's determined by the classifier name here:
https://github.com/bytedeco/javacpp-presets/blob/master/pom.xml#L348

n-kai-cj:
Thank you! I'll try it later.

saudet:
BTW, why do you need a version of OpenCV without OpenBLAS? Are you experiencing any issues with OpenBLAS?

n-kai-cj:
No, I don't have particular reason. I just want to reduce the artifact size.

saudet:
That will only reduce the overall size by 1/3rd. To reduce more, you'll need to disable other modules from OpenCV anyway...

n-kai-cj:
Since openblas of master branch is upgraded, I have to change patch file.
I think it'll be unclear commit if I commit all changes from current master branch, so send another PR https://github.com/bytedeco/javacpp-presets/pull/987

Can I close this PR?

saudet:
We can rebase the branch in those cases, no need to open a new PR, but
since you've already opened one, we can close this one, yes.

PR #987, opened 2020-12-17
add opencv build without openblas dependencies

n-kai-cj:
I'd like to add an option to build OpenCV without OpenBLAS dependencies.
apply patch to exclude OpenBLAS dependencies when environment variable `NOOPENBLAS` is set to `yes`.

build
```
NOOPENBLAS=yes bash cppbuild.sh -platform windows-x86_64 install opencv
mvn clean install -Djavacpp.platform=windows-x86_64 -Djavacpp.cppbuild.skip=true --projects opencv
```
then `opencv-windows-x86_64-noopenblas.jar` is created.

It's a pity that we have to modify the patch file whenever we upgrade OpenBLAS versions.


saudet:
> It's a pity that we have to modify the patch file whenever we upgrade OpenBLAS versions.

We could use sed...

n-kai-cj:
fix to remove version number by `sed` before applying `patch`, so no need to modify patch file when upgrade OpenBLAS version.
I know I should fix all changes by `sed` without `patch`...

saudet:
> I know I should fix all changes by `sed` without `patch`...

Right, please try :)

n-kai-cj:
done!

opencv/cppbuild.sh
```diff
@@ -125,6 +125,73 @@ if [[ "$EXTENSION" == *gpu ]]; then
...
+# exclude openblas dependencies
+if [[ "${NOOPENBLAS:-no}" == "yes" ]]; then
+    # arguments: <file name> <exclude key word>
+    function del_keyword_line {
+        fname=$1
+        ex_word=$2
+        ln=`grep -n $ex_word $fname | sed -n -e 's/:.*//gp'`
+        cnt=0
+        for l in $ln; do
+            sed -i -e "`expr ${l} - ${cnt}`d" $fname
```
supergrecko, line 137: Should probably use the sedinplace function defined [here](https://github.com/bytedeco/javacpp-presets/blob/master/cppbuild.sh#L158) due to `sed -i`'s arguments not being consistent across implementations.

saudet:
Hum, since we need to modify the pom.xml files, it's too late to modify them once Maven has been launched. Of course, we can manually call the cppbuild.sh script, but this patch isn't about C++. What about putting the content in a separate script like noopenblas.sh or something like that? Users will need to execute that manually before the build itself though... It's not pretty, but I don't have a better idea right now. Would you? Do you feel that it's OK to leave this in cppbuild.sh?

n-kai-cj:
Oh, it seems github actions doesn't call `bash cppbuild.sh` before `mvn -Djavacpp.cppbuild.skip=true install`, but just call `mvn install`, then it's too late...
Actually I don't understand the reason why we should separate scripts, but I don't care separate or not. Please as you like.

PR #989, opened 2020-12-20
Support Gradle builds for GitHub Actions

supergrecko:
This patch implements support for Gradle-based presets for the CI workflows

### DIFF
--- a/.github/actions/deploy-macosx/action.yml
+++ b/.github/actions/deploy-macosx/action.yml
@@ -10,7 +10,7 @@ runs:
         export MACOSX_DEPLOYMENT_TARGET=10.13
         echo "MACOSX_DEPLOYMENT_TARGET=$MACOSX_DEPLOYMENT_TARGET" >> $GITHUB_ENV
 
-        brew install boost ccache gcc swig autoconf-archive automake cmake libomp libtool libusb ant maven nasm xz pkg-config sdl gpg1 bison flex perl ragel
+        brew install boost ccache gcc swig autoconf-archive automake cmake libomp libtool libusb ant maven nasm xz pkg-config sdl gpg1 bison flex perl ragel binutils
 
         if [[ -n ${CI_DEPLOY_NEED_GCC:-} ]]; then
           brew install gcc@7
@@ -44,6 +44,10 @@ runs:
           sudo bash bazel-0.25.3-installer-darwin-x86_64.sh
         fi
 
+        if [[ -n ${CI_DEPLOY_NEED_GRADLE:-} ]]; then
+          brew install gradle
+        fi
+
         if [[ "$CI_DEPLOY_MODULE" == "mkl" ]]; then
           echo Installing MKL
           curl -LO https://registrationcenter-download.intel.com/akdlm/irc_nas/tec/17172/m_mkl_2020.4.301.dmg
@@ -51,6 +55,10 @@ runs:
           cp /Volumes/m_mkl_2020.4.301/m_mkl_2020.4.301.app/Contents/MacOS/silent.cfg .
           sed -i -e 's/decline/accept/g' silent.cfg
           sudo /Volumes/m_mkl_2020.4.301/m_mkl_2020.4.301.app/Contents/MacOS/install.sh -s silent.cfg
+        fi
+
+        if [[ "$CI_DEPLOY_MODULE" == "gcc" ]]; then
+          brew install gmp isl libmpc mpfr
         fi
 
     - name: Build project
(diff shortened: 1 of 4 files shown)


supergrecko:
GitHub doesn't seen to want to run the CI for this 🤔

You can see it working in action over at https://github.com/supergrecko/gcc/pull/3 (the gcc builds are failing, but the action is capable of running the gradle build)

.github/actions/deploy-macosx/action.yml
```diff
@@ -53,6 +57,10 @@ runs:
           sudo /Volumes/m_mkl_2020.4.301/m_mkl_2020.4.301.app/Contents/MacOS/install.sh -s silent.cfg
         fi
 
+        if [[ "$CI_DEPLOY_MODULE" == "gcc" ]]; then
+          brew install gmp isl libmpc mpfr
+        fi
```
supergrecko, line 62: This is specific to the WIP gcc preset based on Gradle

.github/actions/deploy-windows/action.yml
```diff
@@ -138,30 +142,43 @@ runs:
...
-        echo Sleeping %SECONDS% seconds to help avoid race conditions with Maven
-        timeout %SECONDS% > nul
+        if not "%CI_DEPLOY_NEED_GRADLE"=="" (
+          setlocal enabledelayedexpansion
+          if "${{ github.event_name }}"=="pull_request" (set GRADLE_TASK=build) else (set GRADLE_TASK=publish)
+          echo Executing Gradle !GRADLE_TASK! on %MAKEJ% processors
+          gradle clean !GRADLE_TASK! --info && goto :done
+          exit /b !ERRORLEVEL!
+          :done
+          echo Done executing Gradle Tasks
```
supergrecko, line 154: I'm not that well versed in Batch and the build wouldn't work without the enabledelayedexpansion flag like we have for the Maven route. It would also fail to exit and fail the CI run if I didn't have this switch for the gradle call. Let me know if you know how we can work around this in a nicer way.

saudet: Batch files don't have an `set -e` mode like Bash, so we need to add explicit checks like that, that's alright, yes. We'll probably need to execute Gradle twice like Maven to work around issues with snapshots, but I'm not sure exactly how that's going to look like yet...

saudet (changes requested):

saudet:
> GitHub doesn't seen to want to run the CI for this thinking

There's no workflows associated with the files in `.github`. That's on purpose, no worries.

> You can see it working in action over at [supergrecko/gcc#3](https://github.com/supergrecko/gcc/pull/3) (the gcc builds are failing, but the action is capable of running the gradle build)

:+1:

saudet:
Ok, I've made a few updates and also fixed the build on Windows (if/else is really strange in batch files, it's easier to use gotos).
I've tested all that with the build for MXNet and everything appears to work well:
https://github.com/saudet/incubator-mxnet/actions/runs/437312011
Please test it with your build for GCC and let me know if you encounter any issues! Thanks

.github/actions/deploy-ubuntu/action.yml
```diff
@@ -113,36 +116,49 @@ runs:
...
-        export EXT=${{ matrix.ext }}
-        export EXT2=${EXT:1}
-        export MAVEN_OPTIONS="clean $MAVEN_PHASE -B -U -e -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.count=3 -Djavacpp.platform=$CI_DEPLOY_PLATFORM -Djavacpp.platform.extension=$EXT $MAVEN_OPTIONS"
-        for i in {1..5}; do STATUS=0; mvn $MAVEN_OPTIONS -f $CI_DEPLOY_MODULE && break || STATUS=$? && sleep 60; done; (exit $STATUS)
-        for i in {1..5}; do STATUS=0; mvn $MAVEN_OPTIONS -f $CI_DEPLOY_MODULE/platform/$EXT2/pom.xml && break || STATUS=$? && sleep 60; done; (exit $STATUS)
+        if [[ -n ${CI_DEPLOY_NEED_GRADLE:-} ]]; then
+          [[ "${{ github.event_name }}" == "pull_request" ]] && GRADLE_TASK=publishToMavenLocal || GRADLE_TASK=publish
+          echo Executing Gradle $GRADLE_TASK ${{ matrix.options }} on $MAKEJ processors
+          export GRADLE_OPTIONS="-Dorg.gradle.jvmargs=-Xmx2048m -PjavacppPlatform=$CI_DEPLOY_PLATFORM -PjavacppPlatformExtension=${{ matrix.ext }} ${{ matrix.options }} --info"
+          gradle clean build $GRADLE_OPTIONS -b $CI_DEPLOY_MODULE/build.gradle
```
supergrecko, line 140: It seems like the `-b $CI_DEPLOY_MODULE/build.gradle` bits are breaking the builds on presets where the build.gradle file is in the root directory of the repository. For example, there is no `gcc` directory inside the gcc preset which is why the build will fail for that preset.

Perhaps there should be an option/switch for this? eg `CI_DEPLOY_GRADLE_ROOT_DIRECTORY` or something similar to toggle this `-b` option?

saudet: Just set `CI_DEPLOY_MODULE: "."`?

supergrecko: That works, but wouldn't that defeat the purpose of module-specific conditions in the composite action file?

It'll work for now, but say we had a large dependency for gcc which we would install only for the gcc module, then the typical `if [[ $CI_DEPLOY_MODULE=="gcc" ]];` wouldn't work anymore.

saudet: Hum, I suppose we should use the name of the workflow `${{ github.workflow }}` for those conditions. Does that work?

supergrecko: You mean something like this? Replacing `$CI_DEPLOY_MODULE` with `${{ github.workflow}}` inside the conditionals?

https://gist.github.com/supergrecko/2632d11089b4f51f376f2c8f9a9f1b8f#file-action-ubunty-yml-L107

saudet: Yes, like that.

saudet: But it doesn't work if we use the workflows for something else and end up with names like `java` anyway, so I don't know. Let's leave that for another time.

supergrecko:
Looks like the gcc build runs as expected I also noticed there's a typo in the ubuntu workflow (https://github.com/bytedeco/javacpp-presets/blob/master/.github/actions/deploy-ubuntu/action.yml#L17).

I can take care of that, want it in this patch or another pr?

saudet:
That's not a typo, that's how that architecture is named under Ubuntu: https://wiki.ubuntu.com/ppc64el

So, looks good to merge?

supergrecko:
Cool, looks good to me!

saudet (approved):

Issue #475, opened 2017-10-08
Merging with JCuda and JOpenCL projects for better quality cuda interfaces


archenroot:
@saudet Hi buddy,
it just came to my head in last few weeks what about merging cuda and opencl stuff here with work of guys from Jcuda and Jopencl projects. I understand there are some fundamental differences, but having rather more quality devs on single project could provide project quality as well.

The guys from JCuda opened discussion on my request here:
https://forum.byte-welt.net/t/about-jcuda-and-javacpp/19538

So, if you think it could bring more value as well, you are free to join the discussion.

saudet:
That would be nice, but the problem is that people expect Oracle to come up with a better solution than JavaCPP, even though they are not working on anything at the moment. As far as I can tell, the developers of Project Panama have given up on any *generic* solution to C++, no one knows how to make something better than JavaCPP. Still, they hope and believe, and wait, mostly. If you could help convincing them that nothing better is going to happen, explaining and reexplaining over and over again how JavaCPP could get better, that would be the first thing that needs to be done.

archenroot:
Just for reference: https://github.com/jcuda/jcuda/issues/12#issuecomment-335010118

archenroot:
@saudet 
I am reading about it it actually started long time ago..(the panama project). What is that project based on, JNI or something new?

jcuda:
There is a *lot* happening in Panama (the project...) right now. Admittedly, although I'm registered to the mailing list, too much to follow it all in detail. However, if they manage to achieve the goals that they stated at the project site, http://openjdk.java.net/projects/panama/ , this would certainly compete with JavaCPP. 

Of course, development there happens at a different pace. We all know that a "single-developer project" often can be far more agile than a company-driven project, where specifications and sustainability play a completely different role. Panama also approaches topics that go **far** beyond what can be accomplished by JavaCPP or JNI in general. They are really going down to the guts, and the work there is interwoven with the topics of Value Types, Vectorization and other HotSpot internals. 

So I agree to saudet that it does not make sense to (inactively) "wait for a better solution". JavaCPP **is** an existing solution for (many of, but by no means all of) the goals that are addressed in Panama.

---

[...]

PR #995, opened 2021-01-12
Implement module-independent data retrieval operations for LLVMNamedMDNodeRef

supergrecko:
I'd like to expand the LLVM C API, more specifically the Modules/NamedMDNode API as it has a rather interesting way to retrieve data owned by a `LLVMNamedMDNodeRef`

In the C++ API, all the data associated with a `llvm::NamedMDNode` is owned by the node itself so it would only make sense to pull said data from the node itself. For unknown? reasons [1], the C API requires the owning module of a `LLVMNamedMDNodeRef` to retrieve the data. Looking closer at the implementation we can tell that we shouldn't need the module to retrieve parts of this data.

The functions I have added in this patch have the same functionality as the ones present in the C API. The only difference is that we don't need to provide a module to look up a LLVMNamedMDNodeRef inside of, instead we provide the instance returned by `LLVMGetNamedMetadata` or `LLVMGetOrInsertNamedMetadata`.

###### [1]: I imagine it was designed this way because in the C++ API, `llvm::NamedMDNode`s may be detached and not associated with a Module so providing the owning module was the natural choice, but this is not possible to do through the C API alone which means we can do better.

### DIFF
--- a/llvm/src/main/resources/org/bytedeco/llvm/include/NamedMetadataOperations.h
+++ b/llvm/src/main/resources/org/bytedeco/llvm/include/NamedMetadataOperations.h
@@ -55,7 +55,6 @@ extern "C" unsigned NamedMDNodeGetOperandCount(LLVMNamedMDNodeRef NodeRef) {
  */
 extern "C" void NamedMDNodeGetOperands(
     LLVMNamedMDNodeRef NodeRef,
-    const char* Name,
     LLVMValueRef *Dest,
     LLVMContextRef InContext
 ) {
@@ -77,7 +76,6 @@ extern "C" void NamedMDNodeGetOperands(
  */
 extern "C" void NamedMDNodeAddOperand(
     LLVMNamedMDNodeRef NodeRef,
-    const char* Name,
     LLVMValueRef Val
 ) {
     assert(Val && "Expected not-null value Val");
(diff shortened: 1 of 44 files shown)


llvm/src/main/resources/org/bytedeco/llvm/include/NamedMetadataOperations.h
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Exact re-implementation of LLVMGetNamedMetadataNumOperands without providing
+ * a LLVMModuleRef parameter
+ *
+ * Pass a LLVMNamedMDNodeRef instead, which is the true owner for the data
+ * anyways
+ *
+ * See /llvm/lib/IR/Core.cpp for original implementation
+ */
+extern "C" unsigned NamedMDNodeGetOperandCount(LLVMNamedMDNodeRef NodeRef) {
```
supergrecko, line 42: Function names are of course, up for debate. Let me know what you think (looking at the discussion in #869). We should probably establish a good convention for these names in case we decide to try the stuff we talked about in #475

saudet: I suppose `getNamedMDNodeNumOperands()`, `getNamedMDNodeOperand()`, and `addNamedMDNodeOperand()` would make most sense?

saudet:
BTW, would you be able to help this guy? https://stackoverflow.com/questions/65634369/llvm-ir-with-java-api

llvm/src/main/resources/org/bytedeco/llvm/include/NamedMetadataOperations.h
supergrecko: Did getNamedMDNodeOperands as its in plural.

PR #1002, opened 2021-01-24
OpenCV: avoiding memory corruption when using Mat constructors taking primitive arrays 

HGuillemet:
`org.bytedeco.opencv.presets.opencv_core.java` defines Mat constructors taking either a java array (var args) or a `Pointer`.

Concerning arrays, one can write :
```
Mat m = new Mat(-1, -1, -1, -1, 8, -1, -1, -1, -1);
```
and, since 1D matrices are not what is needed most of the time, one is tempted to do:
```
m = m.reshape(1, 3);
```
This looks practical and quite armless for anyone used to OpenCV patterns but leads to memory corruption since this Mat constructor actually allocates a buffer, copies the Java array to the buffer, and creates an OpenCV Mat header pointing to this (opencv-external) data. A pointer to the buffer is kept in the Mat object, but when the Mat is garbage collected, the buffer is collected to and the buffer is freed. The reshaped Mat then points to a deallocated buffer.

This PR replaces the Java-side allocation by the allocation of a normal OpenCV Mat, that would benefit from OpenCV reference counting feature. There is no performance loss since we copy the data in both cases.

Concerning Mat constructors taking a `Pointer` as argument, they suffer from the same dangerous limitation. This PR simply adds a warning comment since creating a Mat from a Pointer enables the option to apply opencv processing to any kind of data without memory copy.


opencv/src/main/java/org/bytedeco/opencv/presets/opencv_core.java
```diff
@@ -277,34 +277,34 @@ public void map(InfoMap infoMap) {
...
-                     + "public Mat(ShortPointer p) { this(p, false); }\n"
-                     + "public Mat(ShortPointer p, boolean signed) { this(1, Math.max(1, p.limit() - p.position()), signed ? CV_16SC1 : CV_16UC1, p); }\n"
-                     + "public Mat(IntPointer p) { this(1, Math.max(1, p.limit() - p.position()), CV_32SC1, p); }\n"
-                     + "public Mat(FloatPointer p) { this(1, Math.max(1, p.limit() - p.position()), CV_32FC1, p); }\n"
-                     + "public Mat(DoublePointer p) { this(1, Math.max(1, p.limit() - p.position()), CV_64FC1, p); }\n"))
+                     + "public Mat(short[] s, boolean signed) { this(1, s.length, signed ? CV_16SC1:CV_16UC1); ((ShortBuffer)createBuffer()).put(s); }\n"
+                     + "public Mat(int ... n) { this(1, n.length, CV_32SC1); ((IntBuffer)createBuffer()).put(n); }\n"
+                     + "public Mat(double ... d) { this(1, d.length, CV_64FC1); ((DoubleBuffer)createBuffer()).put(d); }\n"
+                     + "public Mat(float ... f) { this(1, f.length, CV_32FC1); ((FloatBuffer)createBuffer()).put(f); }\n"
```
saudet, line 299: Don't create buffers here. Cast data() to the appropriate Pointer type instead.

saudet (changes requested):

saudet:
Instead of only adding warnings like you did, could you instead add a `boolean copyData` parameter flag and add default overrides that sets it to `false`? That would mostly mimic the original constructors from OpenCV, while keeping backward compatibility for us:
https://docs.opencv.org/master/d3/d63/classcv_1_1Mat.html#a23b182c4ffd46abe38f460c1480ae887

HGuillemet:
Good idea. 
But maybe we shouldn't add overloads for the 18 concerned constructors.
What about passing true for constructors taking Scalar and Points ?

Another possible addition : constructors taking 2D Java arrays ?

HGuillemet:
Hmm, forget my remark, Points and Scalars can be arbitrarily long. So 18 overloads ?

saudet:
> Good idea.
> But maybe we shouldn't add overloads for the 18 concerned constructors.

Why not? :) It's just like `public Mat(Point points) { this(points, false); }` etc, could even be on the same line...

> What about passing true for constructors taking Scalar and Points ?

No, that would break backward compatibility and it doesn't work well in our case where everything is a Pointer and which constructor might get used isn't always clear.

> Another possible addition : constructors taking 2D Java arrays ?

Yes, of course, there's a lot we can do :)

saudet:
BTW, since this is just Java code, we can use for loops and what not to generate code more easily, for example, like this:
https://github.com/bytedeco/javacpp-presets/blob/master/caffe/src/main/java/org/bytedeco/caffe/presets/caffe.java#L148

HGuillemet:
In the constructors taking pointers, if the pointers'limit equals the position, the length is set to 1.
What is the rationale behind that ? Doesn't that causes obscure bugs ?

saudet:
> In the constructors taking pointers, if the pointers'limit equals the position, the length is set to 1.
> What is the rationale behind that ? Doesn't that causes obscure bugs ?

I suppose we could leave that to 0? That's probably going to throw some exception, which would be acceptable.

HGuillemet:
> > In the constructors taking pointers, if the pointers'limit equals the position, the length is set to 1.
> > What is the rationale behind that ? Doesn't that causes obscure bugs ?
> 
> I suppose we could leave that to 0? That's probably going to throw some exception, which would be acceptable.

Pointers may have a limit of 0 when it's unknown (that seems to be the case of `Mat.data()`).
But throwing an exception should be better than silently supposing the length is 1.

See new proposition `with copyData` in last commit.
Similarly to the case `position == limit`, I also added a check when casting a long to int, instead of clamping.

saudet:
Except for a couple of small mistakes in the comments, looks good! To be consistent with the rest of JavaCPP though, instead of throwing an exception, let's saturate sizes this way: `size < Integer.MAX_VALUE ? (int)size : Integer.MAX_VALUE`.

HGuillemet:
> To be consistent with the rest of JavaCPP though, instead of throwing an exception, let's saturate sizes this way: `size < Integer.MAX_VALUE ? (int)size : Integer.MAX_VALUE`.

I'm not aware of the consistency problem, but saturation would yield to something like:
```
FloatPointer p = new FloatPointer(0x90000000L);
Mat m = new Mat(p);
```
silently creating a mat of length 0x7FFFFFFF instead of throwing an exception. Is this really why is preferred ?
Fortunately, any `reshape` afterwards would yield to an exception because of non-divisibility of `0x7FFFFFFF` but other situations could lead to non obvious disaster.

saudet:
I like it better that way anyway, less exceptions to worry about :) It's handled like that everywhere else in JavaCPP and I never heard anyone having any problems with it. Containers like that don't support large enough indices, so the user will get into trouble trying to index those with `long` anyway, and that throws compiler errors. Much better than an exception at runtime.

opencv/src/gen/java/org/bytedeco/opencv/opencv_core/Mat.java
```diff
@@ -360,29 +360,49 @@ elements of an array one by one (or elements from multiple arrays that have the
     private Pointer pointer; // a reference to prevent deallocation
```
saudet, line 360: As can be seen from the generated code above, we don't need to add a Javadoc comment for the original function since the one from the header file of OpenCV is already good enough.

saudet (approved):

saudet:
@HGuillemet Please let me know if that looks good enough and all correct! Thanks

opencv/src/gen/java/org/bytedeco/opencv/opencv_core/Mat.java
HGuillemet: As you wish, but this OpenCV comment warns about not forgettting to deallocate manually. But that's not the main problem, it's when the data is deallocated too early by the GC.

opencv/src/main/java/org/bytedeco/opencv/presets/opencv_core.java
```diff
@@ -277,54 +277,57 @@ public void map(InfoMap infoMap) {
...
+                     + "/** Calls {@link #Mat(Point3f, boolean) Mat(points, false)}. */  public Mat(Point3f points)  { this(points, false); }\n"
+                     + "/** Calls {@link #Mat(Point3d, boolean) Mat(points, false)}. */  public Mat(Point3d points)  { this(points, false); }\n"
+                     + "/** Calls {@link #Mat(Scalar, boolean) Mat(scalar, false)}. */   public Mat(Scalar scalar)   { this(scalar, false); }\n"
+                     + "/** Calls {@link #Mat(Scalar4i, boolean) Mat(scalar, false)}. */ public Mat(Scalar4i scalar) { this(scalar, false); }\n"
 
                      + "public Mat(byte ... b) { this(b, false); }\n"
-                     + "public Mat(byte[] b, boolean signed) { this(1, b.length, signed ? CV_8SC1:CV_8UC1); data().put(b); }\n"
-                     + "public Mat(short ... s) { this(s, true); }\n"
-                     + "public Mat(short[] s, boolean signed) { this(1, s.length, signed ? CV_16SC1:CV_16UC1); new ShortPointer(data()).put(s); }\n"
+                     + "public Mat(short ... s) { this(s, false); }\n"
```
HGuillemet, line 313: You changed `true` to `false`, but this is the `signed`parameter, not `copyData`. It used to be `true` by default.

opencv/src/gen/java/org/bytedeco/opencv/opencv_core/Mat.java
saudet: Users of raw pointers will be the ones to know if their buffers are registered with the GC or not. OpenCV can't know about that. It's actually your comment that isn't entirely accurate. It should say that, for example, if the external buffer has a deallocator registered with the GC, the user needs to make sure to keep a reference somewhere to prevent premature deallocation.

opencv/src/main/java/org/bytedeco/opencv/presets/opencv_core.java
saudet: Yes, that was an error I believe. It was false in the case of ShortPointer...

opencv/src/gen/java/org/bytedeco/opencv/opencv_core/Mat.java
saudet: I think I've improved your warning. Let me know what you think! Thanks

opencv/src/main/java/org/bytedeco/opencv/presets/opencv_core.java
saudet: I mean, it's an error I made when I originally wrote this a few years ago, not with your update.

HGuillemet: OK for me

HGuillemet: Ok

PR #1007, opened 2021-02-11
Updating libpng version.

nagygr:
Leptonica was built with an earlier version of libpng due to problems on CentOS6. That system has reached EOL. At the moment the old version is not compatible with e.g. the version that JavaFX uses making it impossible to build applications using JavaFX and Leptonica or Tesseract. For details, please see: https://github.com/bytedeco/javacpp-presets/discussions/1006


saudet:
Thanks! You can try it out with the snapshots: http://bytedeco.org/builds/

PR #1012, opened 2021-02-22
Add JavaCPP Preset for ModSecurity library

artemMartynenko:


modsecurity/cppbuild.sh
```diff
@@ -0,0 +1,51 @@
...
+# This file is meant to be included by the parent cppbuild.sh script
+if [[ -z "$PLATFORM" ]]; then
+    pushd ..
+    bash cppbuild.sh "$@" modsecurity
+    popd
+    exit
+fi
+
+
+sudo apt-get install g++ flex bison curl doxygen libyajl-dev libgeoip-dev libtool dh-autoreconf libcurl4-gnutls-dev libxml2 libpcre++-dev libxml2-dev
```
supergrecko, line 11: The build script should probably not install system dependencies, maybe add a notice on the README mentioning which system dependencies are required?

modsecurity/platform/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+      <artifactId>${javacpp.moduleId}</artifactId>
+      <version>${project.version}</version>
+      <classifier>${javacpp.platform.linux-arm64}</classifier>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>${javacpp.moduleId}</artifactId>
+      <version>${project.version}</version>
+      <classifier>${javacpp.platform.linux-ppc64le}</classifier>
+    </dependency>
```
supergrecko, line 62: According to the cppbuild you're only building for linux-x86, linux-x86_64 and macosx-x86_64. These dependencies for the -platform dependency wouldn't exist.

modsecurity/platform/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+            </goals>
+            <configuration>
+              <modules>
+                <module>
+                  <file>${project.build.directory}/${project.artifactId}.jar</file>
+                  <moduleInfoSource>
+                    module org.bytedeco.${javacpp.moduleId}.platform {
+                      requires static org.bytedeco.${javacpp.moduleId}.linux.x86_64;
+                      requires static org.bytedeco.${javacpp.moduleId}.macosx.x86_64;
+                    }
```
supergrecko, line 128: This should requires static linux.x86

modsecurity/cppbuild.sh
artemMartynenko: Yep, I agree.

modsecurity/platform/pom.xml
artemMartynenko: Fixed

modsecurity/cppbuild.sh
artemMartynenko: Fixed

modsecurity/platform/pom.xml
artemMartynenko: I removed `linux.x86` configurations from the platform build because the library always builds for x86_64 arch.

supergrecko:
We should also set up automatic builds for the preset through GitHub Actions like we have for the other presets, see https://github.com/bytedeco/javacpp-presets/tree/master/.github/workflows

modsecurity/cppbuild.sh
saudet: Those should be moved to the scripts for GitHub Actions, if needed.

artemMartynenko:
> We should also set up automatic builds for the preset through GitHub Actions like we have for the other presets, see https://github.com/bytedeco/javacpp-presets/tree/master/.github/workflows

artemMartynenko:
I need some help. I added workflow and now everything works ok for Linux builds, but for macOS, I got a fail ```class ModSecurity is public, should be declared in a file named ModSecurity.java```. You can see it in the check result. What can be the reason?

saudet:
File names are case insensitive on Mac and Windows. We can't have both a file named "modsecurity.java" and "ModSecurity.java" in the same directory.

modsecurity/README.md
```diff
@@ -0,0 +1,126 @@
...
+  <build>
+    <sourceDirectory>.</sourceDirectory>
+  </build>
+</project>
+
+```
+
+### The `ModSecuritySimpleIntervention.java` source file
+```java
+import org.bytedeco.javacpp.*;import org.bytedeco.modsecurity.*;
```
supergrecko, line 69: Quickly noticed there's an extra `import org.bytedeco.javacpp.*;` here

saudet:
If you could also commit the files under `src/gen`, it would help to diagnose this issue. Thanks

modsecurity/README.md
artemMartynenko: Fixed

artemMartynenko:
Sorry for the delay. I was a little bit busy. Now everything works.

supergrecko (changes requested):
I believe modsecurity should be added to this list in the main cppbuild.sh script

https://github.com/bytedeco/javacpp-presets/blob/master/cppbuild.sh#L167

artemMartynenko:
Totally agree. Done.

modsecurity/cppbuild.sh
```diff
@@ -0,0 +1,43 @@
...
+cd $PLATFORM
+INSTALL_PATH=`pwd`
+
+if   [ ! -d "ModSecurity" ];
+then
+  git clone https://github.com/SpiderLabs/ModSecurity
+  cd ModSecurity
+  git checkout origin/v3/master
+  git submodule init
+  git submodule update
```
saudet, line 20: They seem to offer proper source code bundles for releases:
https://github.com/SpiderLabs/ModSecurity/releases
Say we replace these `git` calls with something like the following, does it still work?
```
download https://github.com/SpiderLabs/ModSecurity/releases/download/v3.0.4/modsecurity-v3.0.4.tar.gz modsecurity-v3.0.4.tar.gz
```

artemMartynenko: I need to check it. I will answer as soon as check it.

artemMartynenko: I checked it. Unfortunately, it not works. There were a lot of changes from the previous release and it seems that release rules will be changed. I think it will work after the release 3.1.0.  And also their current recommendation to build from the master branch,  maybe it associated with the release policy.

saudet: Hum, ok, let's wait and see for 3.1.0? How soon is that going to happen?

saudet: We could also set the version of the presets to `master`, temporarily, until 3.1.0 gets released. What do you think?

artemMartynenko: Yep, we can set the version of the preset to master, temporarily, and then change it to 3.0.5 after the release. "Migration from master and 3.0.5 to 3.1.0 will be much easier" - it`s a [comment](https://github.com/SpiderLabs/ModSecurity/issues/2534) which I got on the issue about next release. Thanks a lot :+1:

artemMartynenko:
I have seen that 1.5.5 was released, so it seems that I need to update the code in PR to 1.5.6-SNAPSHOT to make it acceptable?

supergrecko:
> I have seen that 1.5.5 was released, so it seems that I need to update the code in PR to 1.5.6-SNAPSHOT to make it acceptable?

Correct, the version should be changed to 1.5.6-SNAPSHOT

saudet:
Yes, but also the release version of ModSecurity itself. It's not "3.0.4", so "master" will do for now.

artemMartynenko:
Versions are fixed. Thanks!

modsecurity/README.md
```diff
@@ -0,0 +1,126 @@
...
+  <artifactId>modsecurity-sample</artifactId>
+  <version>1.5.5-SNAPSHOT</version>
+  <properties>
+    <exec.mainClass>ModSecuritySimpleIntervention</exec.mainClass>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>org.bytedeco</groupId>
+      <artifactId>modsecurity-platform</artifactId>
+      <version>3.0.4-1.5.5-SNAPSHOT</version>
```
saudet, line 57: The versions haven't been updated in the README.md file.

.github/actions/deploy-macosx/action.yml
```diff
@@ -10,7 +10,9 @@ runs:
         export MACOSX_DEPLOYMENT_TARGET=10.13
         echo "MACOSX_DEPLOYMENT_TARGET=$MACOSX_DEPLOYMENT_TARGET" >> $GITHUB_ENV
 
-        brew install boost ccache gcc swig autoconf-archive automake cmake libomp libtool libusb ant maven nasm xz pkg-config sdl gpg1 bison flex perl ragel binutils gradle gmp isl libmpc mpfr
+        brew install boost ccache gcc swig autoconf-archive automake cmake libomp libtool libusb ant maven nasm xz pkg-config sdl gpg1 bison flex perl ragel binutils gradle gmp isl libmpc mpfr zlib curl pcre libffi yajl ssdeep luarocks
+        brew install geoip --with-geoipupdate
+        brew install doxygen --with-llvm
```
saudet, line 15: Why is the set of dependencies different on Linux and Mac? It looks like it's building without any of those dependencies too. Which ones do you think we actually need?

modsecurity/README.md
saudet: I still see 3.0.4 in the README.md...

artemMartynenko: maybe you did not update? I fixed it with the last commit today.

.github/actions/deploy-macosx/action.yml
artemMartynenko: If you are talking about geoip and oxygen they are installing on Linux too. For Mac, they have separate lines just because they need additional args for installation. Or you meant something else?

saudet: Well, it doesn't look like Lua or libffi gets used for anything, while curl and zlib come with the system, so I'll probably just leave those out. I also don't think we need to geoipupdate or LLVM in there either. Sounds good?

modsecurity/README.md
saudet: No, still seeing it in the update: https://github.com/bytedeco/javacpp-presets/blob/045c7f5e114509fd442e0a34f5f5287bb33daa24/modsecurity/README.md#introduction

.github/actions/deploy-macosx/action.yml
supergrecko: I'm also not sure we need doxygen to build an installation of the project, see https://github.com/SpiderLabs/ModSecurity#Dependencies

Doxygen is used to create api documentation from the source code as well as other documentation pages. Shouldn't be necessary to build modsecurity.

artemMartynenko: Sorry for the late response. @saudet yes it sounds good. I see that you already fixed it. Thanks! Once again sorry for the late response, I was AFK last week.

modsecurity/README.md
artemMartynenko: Sorry, I focused on commented lines. Now it`s already fixed. Thanks!

saudet:
@artemMartynenko FYI, v2.9.4 has now been released. We may want to set the version of the presets to that release.

artemMartynenko:
Hello. You should set the version to 2.9.4 because it's supporting the release of the previous version. This preset that were merged it`s for version 3+. We set the version to master because there were intermediate changes between 3.0.4 and 3.0.5. I got the [message](https://github.com/SpiderLabs/ModSecurity/issues/2534#issuecomment-872928695) that version 3.0.5 will be released tomorrow, so after release, we can fix the version to 3.0.5.

saudet:
> Hello. You should not set the version to 2.9.4 because it's a supporting release of the previous version. This preset that were merged it`s for version 3+. We set the version to master because there were intermediate changes between 3.0.4 and 3.0.5. I got the [message](https://github.com/SpiderLabs/ModSecurity/issues/2534#issuecomment-872928695) that version 3.0.5 will be released tomorrow, so after release, we can fix the version to 3.0.5.

Thanks! Please send a pull request when it's released.

artemMartynenko:
No problem, I will do that.

PR #1029, opened 2021-03-25
[WIP] Upgrade LLVM Preset to 12.0.0

supergrecko:
This is a PR for the upcoming LLVM 12.0.0 release

This build is currently building from rc3 which is likely to be the last release candidate before release. The cppbuild.sh script currently has some temporary fixes which will be resolved before release.

## Significant changes to the C API

- Added X86_AMX Type: https://github.com/llvm/llvm-project/commit/981a0bd85811fe49379fdbef35528e2c2f3511a3
- Deletion of unimplemented header: https://github.com/llvm/llvm-project/commit/6785ca01248ce271dfff16e03cce82f649fa14f7
- Added Poison Value APIs: https://github.com/llvm/llvm-project/commit/75f50e15bf8fff6fba1d4678adedd33ef6a945e5
- Scalable Vectors APIs: https://github.com/llvm/llvm-project/commit/c3783847ae67d31aece461458d8510b8bd0469c3
- LLVMDIGenericSubrangeMetadataKind in DebugInfo.h: https://github.com/llvm/llvm-project/commit/a6dd01afa3d5902203d04a72e0b478078f796a35
- OrcJITv1 removed in favor of OrcJITv2
  - Deletion of OrcBindings.h: https://github.com/llvm/llvm-project/commit/6154c4115cd4b78d0171892aac21e340e72e32bd
  - Error Reporter setter for OrcJITv2: https://github.com/llvm/llvm-project/commit/14cb9b4e21de63f74439259bad8caa6fcb7136e1
  - Symbol String pool in OrcJITv2: https://github.com/llvm/llvm-project/commit/35e48d7b91735136e423f71c1d7406eb07a03e89 & https://github.com/llvm/llvm-project/commit/00369849e10943b3afd35505504368a8cdc77100
  - Renamed LLVMOrcJITDylibDefinitionGeneratorRef to LLVMOrcJITDefinitionGeneratorRef: https://github.com/llvm/llvm-project/commit/49c065ae70ecde9c2176da4ff801076b9c2f132b=
  - Added resource tracking to OrcJITv2: https://github.com/llvm/llvm-project/commit/91d1f417fd89cdbea095f820729df0eb9d465f9f
  - Added API for creating StringError: https://github.com/llvm/llvm-project/commit/19402ce79a896563823fff45b853f229df4c1261
  - Added support for custom generators for OrcJITv2: https://github.com/llvm/llvm-project/commit/b6ca0c7dd53528cece4e058eb12c281792893b2f
  - Added C API for defining absolute symbols: https://github.com/llvm/llvm-project/commit/24afffe63a7b0c545ab854f7955c9b4322e7e690
  - Split Orc.h header into Orc.h, OrcEE.h and LLJIT.h: https://github.com/llvm/llvm-project/commit/f35707047e216e99052c15b04fe508d075b6121b


llvm/cppbuild.sh
```diff
@@ -125,4 +130,6 @@ case $PLATFORM in
         ;;
 esac
 
+cp -r $INSTALL_PATH/include/polly/isl $INSTALL_PATH/include/
+
```
supergrecko, line 134: I had compilation problems with polly where their isl dependency headers were undiscoverable by the JavaCPP jnilib command. It looks to me like a regression from 11.0.0, but I will do more research. This is a temporary fix which works for now.

supergrecko:
Will take care of Arrow update to 12.0.0 once LLVM releases a final version

llvm/cppbuild.sh
saudet: @yukoba Here, it looks like Polly changed with LLVM 12. If you could help with that, please do! Nothing urgent though.

yukoba: @saudet
> Polly changed with LLVM 12

I see.

> Is there a way to add header search paths to the JavaCPP invocation without messing with CXX/CFLAGS?

I think you should answer this.
If I am not wrong, I think something related to ```@Platform```.

saudet: If it's just for `LinkAllPasses.h`, just updating `polly/LinkAllPasses.h` with `polly/isl/LinkAllPasses.h` should work...

supergrecko: I'm not sure its as simple as that... The file in question is the `$INSTALL_PATH/polly/isl/isl-noexceptions.h` which seems to be an entrypoint to libisl which doesn't use exceptions and this header is indirectly included through `polly/ScopInfo.h`. Rewriting `polly/ScopInfo.h` path to include polly would not work either because `polly/isl/isl-noexceptions.h` includes even more of isl's headers which would result in header search failure once that file is included.

Building without moving isl's headers yields the following compile error

```
In file included from /home/me/source/supergrecko/javacpp-presets/llvm/cppbuild/linux-x86_64/include/polly/ScopDetection.h:50,
                 from /home/me/source/supergrecko/javacpp-presets/llvm/cppbuild/linux-x86_64/include/polly/ScopInfo.h:20,
                 from /home/me/source/supergrecko/javacpp-presets/llvm/cppbuild/linux-x86_64/include/polly/ScopPass.h:20,
                 from /home/me/source/supergrecko/javacpp-presets/llvm/cppbuild/linux-x86_64/include/polly/Simplify.h:16,
                 from /home/me/source/supergrecko/javacpp-presets/llvm/cppbuild/linux-x86_64/include/polly/LinkAllPasses.h:20,
                 from /home/me/source/supergrecko/javacpp-presets/llvm/target/native/org/bytedeco/llvm/linux-x86_64/jniLLVM.cpp:134:
/home/me/source/supergrecko/javacpp-presets/llvm/cppbuild/linux-x86_64/include/polly/Support/ScopHelper.h:20:10: fatal error: isl/isl-noexceptions.h: No such file or directory
   20 | #include "isl/isl-noexceptions.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

I think we can solve this by adding it to the `@Platform` include path. Is it possible to use  https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/annotation/Platform.java#L62 relative from the install directory? We could alternatively rewrite every include directive in isl and polly's headers but that sounds very error prone.

saudet: Ah, I see. For these cases, we can add more includePath in the pom.xml file like this:
https://github.com/bytedeco/javacpp-presets/blob/master/cpython/pom.xml#L36

supergrecko:
It seems like the Windows builds are failing because of two undefined externals, `LLVMOrcDisposeObjectLayer` and `LLVMOrcExecutionSessionGetJITDylibByName`. Both of which should be defined in this source file https://github.com/llvm/llvm-project/blob/main/llvm/lib/ExecutionEngine/Orc/OrcV2CBindings.cpp

I am unsure of what is wrong with LLVMOrcExecutionSessionGetJITDylibByName, but in LLVMOrcDisposeObjectLayer's case, it seems like the function is mistyped with a small L in LLVM. I will report about this to upstream.

We should be able to just skip these symbols for now using `new Info().skip()` right?

saudet:
Right, when symbols like that are not exported, I just consider that no one cares about them, so skip :)

supergrecko:
That does mean we're going to lose LLVMs destructor for llvm::orc::ObjectLayer and the LLJITDylib getter for now. I'm still waiting for bugzilla + mailing list access for LLVM. I highly doubt it, but we may be able to land a patch before 12.0.0. In the meantime we can consider adding polyfills for these functions if people show interest..

supergrecko:
I've managed to land a patch for the linking issues into upstream here https://github.com/llvm/llvm-project/commit/666df2e2cbe9fc252d3b2d6cbb214c2c2f6afc65. I'm not sure, but I would assume it'll be merged into 12.0.0-final. Let's wait and see what happens.

supergrecko:
Re-built from LLVM 12.0.0-rc5 which includes the patch for LLVM OrcJITv2 API fixes. I'm also looking into fixing LLVMConstGEP2 and LLVMConstInBoundsGEP2

saudet:
We can probably just keep skipping these 2 functions since they don't actually exist anyway.

saudet:
Looks like it's running out of disk space on Windows. We'll probably need to set the build directory to something like `C:\build` where is a lot more space:
```
2021-04-15T12:36:13.3120523Z Filesystem      Size  Used Avail Use% Mounted on
2021-04-15T12:36:13.3121700Z C:/msys64       256G  167G   89G  66% /
2021-04-15T12:36:13.3122929Z D:               14G  2.0G   13G  14% /d
```
As for Arrow, we also need to tell it to accept LLVM 12.x by updating this line in the build script:
https://github.com/bytedeco/javacpp-presets/blob/master/arrow/cppbuild.sh#L39

supergrecko:
The following patch will have to be applied to the `actions` branch to update CMake to 3.16 before this is merged.

https://github.com/supergrecko/javacpp-presets/commit/36e44ffc7978ccd55fd17aa4b679e277dff287d6

.github/workflows/llvm.yml
```diff
@@ -48,11 +48,18 @@ jobs:
...
+    env:
+      BUILD_DIR: C:\build
     steps:
-      - uses: supergrecko/javacpp-presets/.github/actions/deploy-windows@cmake-3-16
+      - uses: al-cheb/configure-pagefile-action@v1.2
+        with:
+          minimum-size: 8GB
+          maximum-size: 16GB
+          disk-root: "C:"
+      - uses: bytedeco/javacpp-presets/.github/actions/deploy-windows@actions
```
saudet, line 62: Not the pagefile directory, the build directory of CMake.

arrow/cppbuild.sh
```diff
@@ -36,7 +36,7 @@ tar --totals -xzf ../openssl-$OPENSSL_VERSION.tar.gz
 tar --totals -xf ../llvm-$LLVM_VERSION.src.tar.xz || tar --totals -xf ../llvm-$LLVM_VERSION.src.tar.xz
 cd apache-arrow-$ARROW_VERSION
 patch -Np1 < ../../../arrow.patch
-sedinplace 's/ARROW_LLVM_VERSIONS "10"/ARROW_LLVM_VERSIONS "11" "10"/g' cpp/CMakeLists.txt
+sedinplace 's/ARROW_LLVM_VERSIONS "10"/ARROW_LLVM_VERSIONS "11" "10" "12"/g' cpp/CMakeLists.txt
```
saudet, line 39: You probably got the order wrong here.

.github/workflows/llvm.yml
supergrecko: Right, what would that look like? I just grabbed this from the Pytorch preset file which had CI setup like that.

arrow/cppbuild.sh
supergrecko: OSX wouldn't build with the LLVM 12 version because that sed doesn't match the file contents anymore, they've put each version on a new line. Should have a fix for this locally which I will push.

.github/workflows/llvm.yml
saudet: Usually with CMake we just need to `cd ...` to a different directory before calling `cmake ...`, and I don't think LLVM tries to do anything fancy here like we need to do for PyTorch.

supergrecko: So you suggest we just cd into the C:\ disk and build everything in there?

saudet:
Yes, but it should be configurable with an environmental variable like
BUILD_DIR.

supergrecko:
Did you mean something like the above?

The Arrow LLVM version problem should be fixed now. Works locally for me.

PR #1031, opened 2021-04-10
Upgrade librealsense2 to 2.44.0

n-kai-cj:
upgrade librealsense2 to 2.44.0.
also, add `Release` built option to all platforms, otherwise making some processes low performance.


supergrecko (commented):
Looks good. You also want to update the version numbers in both README files as well as the platform builds. That means the version should be updated in the following files:

- librealsense2/platform/pom.xml
- librealsense2/README.md
- README.md

PR #1032, opened 2021-04-16
add cuda build for librealsense2

n-kai-cj:
Add cuda build option `BUILD_WITH_CUDA` for librealsense2.
We can get benefits of cuda for some processes such as alignment.


saudet:
CUDA is only available for linux-arm64, linux-ppc64le, linux-x86_64, and windows-x86_64, but it won't work with linux-arm64 and linux-ppc64le because we would need first to get the cross-compiler working on GitHub Actions, so please adjust the workflows/librealsense2.yml and platform/gpu/pom.xml files with only linux-x86_64 and windows-x86_64, for now anyway. Thanks!

n-kai-cj:
I modified to add CUDA option for only linux-x86_64 and windows-x86_64.

BTW, CUDA option is originally for ARM platform because librealsense2 supports only SSE and CUDA backend, NEON is not implemented.
I'd like to use `linux-arm64-gpu`, could you make environments with cross-compiler for ARM? or are there anything I could help?

saudet:
If you would like to figure how to get the cross-compiler working for Arm,
please do!

n-kai-cj:
Actually I don't understand GitHub Actions at all, but I'll try to learn.

saudet:
No need to learn GitHub Actions. We first need to get it working,
somewhere, anywhere, it doesn't matter, for now. We'll worry about GitHub
Actions later.

n-kai-cj:
I did build success on pure arm platform with CUDA, but I'm not sure whether we can build CUDA with cross compiling.
GitHub Actions may have ARM actions function? If yes, we don't need to cross compile.
https://github.com/marketplace/actions/arm-action

saudet:
Maybe, I don't think that it works well though. Please try to make it work
either way.

n-kai-cj:
I found CUDA is not supported by ARM except for Jetson. So I removed ARM build from GitHub Actions workflow, but cuda build option is left so that we can get linux-arm64-gpu.jar if we build with physical arm platform such as Jetson.

saudet:
CUDA does support Arm since 11.0:
https://developer.nvidia.com/cuda-11.0-download-archive?target_os=Linux&target_arch=sbsa

saudet:
In any case, please fix the workflow and let's see if there are any build errors. Thanks!

PR #1039, opened 2021-04-30
Build Arrow using LLVM preset

junlarsen:
This is a follow up to #1029

Builds the Apache Arrow preset using the LLVM preset instead of compiling LLVM on its own.

This does not work in its current state (as discussed on Gitter)


saudet:
It appears that the CMake config files from LLVM that Arrow uses needs all the static libraries, so to do things correctly we would need to bundle all of them in the JAR files for all platforms. Right now, we're bundling them only for Windows because it doesn't support dynamic libraries for LLVM, and that's why the JAR files for that platform are so huge. We could hack the CMake build files to remove all traces of the static libraries, but Arrow itself was designed to link statically with LLVM anyways, so it may be more trouble than it's worth trying to figure out a way to make it all work with the dynamic libraries...

supergrecko:
I see, I think we should do what we did for the LLVM build here as well then right?

saudet:
We could, but it doesn't build as much of LLVM and it looks like we won't need it:
https://github.com/bytedeco/javacpp-presets/actions/runs/820400693

supergrecko:
Okay, I suppose I can close this then?

PR #1040, opened 2021-05-05
Upgrade Spinnaker to version 2.4.0.143

jpsacha:


jpsacha:
CI Test are failing, but it looks that the old version of the links for binaries are code is used. It is downloading old version of binaries 1.27.0.48 using old links. For instance see https://github.com/bytedeco/javacpp-presets/pull/1040/checks?check_run_id=2505944194#step:2:283
That link is not present in the committed code. Is this some code from cache?

saudet:
No, it's the location of the actions used in the workflow that is causing this:
https://github.com/bytedeco/javacpp-presets/blob/master/.github/workflows/spinnaker.yml
You can temporarily change those to this branch on your fork to this test the update.

jpsacha:
I do not see any reference to specific branches in https://github.com/bytedeco/javacpp-presets/blob/master/.github/workflows/spinnaker.yml 
It is not clear to me how downloads are going from a different branch than tested. May be some other some other file has them hardcoded?

By the way, Spinnaker dropped support for Ubuntu 16.04 (as it lost its LTS status this month), so that will need to be updated to 18.04 or 20.04. Spinnaker binaries used in CI are for 18.04 (20.04 are available too). I can update that separately. 16.04 dropping from LST will impact other tests, beside Spinnaker, too.

saudet:
> I do not see any reference to specific branches in https://github.com/bytedeco/javacpp-presets/blob/master/.github/workflows/spinnaker.yml
> It is not clear to me how downloads are going from a different branch than tested. May be some other some other file has them hardcoded?

Simply replace `bytedeco/javacpp-presets` with `jpsacha/javacpp-presets` and `@actions` with `@Spinnaker_2.4.0.143`.

> By the way, Spinnaker dropped support for Ubuntu 16.04 (as it lost its LTS status this month), so that will need to be updated to 18.04 or 20.04. Spinnaker binaries used in CI are for 18.04 (20.04 are available too). I can update that separately. 16.04 dropping from LST will impact other tests, beside Spinnaker, too.

It's still getting built from CentOS 6, and it's working fine, so I wouldn't worry about that.

saudet:
BTW, for debugging purposes, we can connect via SSH to the build machines with, for example, this action:
https://github.com/marketplace/actions/debugging-with-tmate

jpsacha:
All set now. CI build are passing. Enjoying new faster CI setup 👍

PR #1045, opened 2021-05-08
Remove JPMS dependencies from native modules

HGuillemet:
Native module currently depends on other native modules with same system, architecture and extension.
For instance `org.bytedeco.opencv.linux.x86_64` descriptor contains `requires org.bytedeco.openblas.linux.x86_64`. But when using a specific extension, like `gpu`, module graph initialization fails when the extension doesn't exist for some dependency, like `openblas` or `javacpp` itself.

This PR removes these dependencies between modules. The preferred way to bring the native modules into the module graph is to rely on Maven dependencies using the `javacpp.platform` and `javacpp.platform.extension` properties, and on `--add-modules ALL-MODULE-PATH` option.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -395,9 +395,6 @@
                   <file>${project.build.directory}/${project.artifactId}-${javacpp.platform}${javacpp.platform.extension}.jar</file>
                   <moduleInfoSource>
                     open module org.bytedeco.${javacpp.packageName}.${javacpp.platform.module} {
-                      requires transitive org.bytedeco.${javacpp.packageName};
-                      requires org.bytedeco.javacpp.${javacpp.platform.module};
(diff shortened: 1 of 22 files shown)


pom.xml
```diff
@@ -395,9 +395,6 @@
                   <file>${project.build.directory}/${project.artifactId}-${javacpp.platform}${javacpp.platform.extension}.jar</file>
                   <moduleInfoSource>
                     open module org.bytedeco.${javacpp.packageName}.${javacpp.platform.module} {
-                      requires transitive org.bytedeco.${javacpp.packageName};
```
saudet, line 398: I think we should leave the first one here though. And we should also remove `<javacpp.nativeRequires/>`

Issue #833, opened 2019-12-30
How to create a function pointer and call it?


oneengineer:
```cpp
int (*sum_func)(int, int) = (int (*)(int, int))LLVMGetFunctionAddress(engine, "sum");
```

I need to call `LLVMGetFunctionAddress` like the C code above in Java, however it returns a pointer and I need to convert it to `(int (*)(int, int))` function.

More specifically, I guess the code would be like:

```java
long funAddr = LLVMGetFunctionAddress(engine, "sum")

new Pointer {

   //someshow pass the funAddr to it

   public native void call( /*parameters here*/ );

}
```

saudet:
Are you saying that you would like to call that function using your native platform's C/C++ compiler?

oneengineer:
> Are you saying that you would like to call that function using your native platform's C/C++ compiler?

I updated the question. 
No I want to call the function pointer in Java. But I don't know how to do it.

saudet:
Since you're using LLVM, you might as well keep using that. There is an example of function calling here:
https://github.com/bytedeco/javacpp-presets/blob/master/llvm/samples/Fac.java

oneengineer:
I have read this article and tried to follow it to try this lib, however I have to (I don't know other ways) to use `LLVMGetFunctionAddress` instead of `LLVMRunFunction`. 

It seems `LLVMRunFunction` only support 1 argument?
In the stderr it print, it tells use `LLVMGetFunctionAddress` like the C code above instead of `LLVMRunFunction`.

I thought converting **pointer** to **a function pointer** is quite normal in C.
Can I do the same thing in javacpp?

saudet:
I see, you're talking about the limitation of LLVM discussed on this page?
https://www.owenstephens.co.uk/blog/2018/09/25/getting-started-with-the-newer-llvm-c-api.html
To do the same thing from Java with JavaCPP, we still need to use a C++ compiler, but if that's OK with you then yes we can define a new `FunctionPointer` that matches that signature. There's an example of that in the unit tests here:
https://github.com/bytedeco/javacpp/blob/master/src/test/java/org/bytedeco/javacpp/PointerTest.java

oneengineer:
I created function wrapper like this:

```
public class Fn extends FunctionPointer {
        /** Pointer cast constructor. Invokes {@link Pointer#Pointer(Pointer)}. */
        public    Fn(Pointer p) { super(p); }
        public    Fn(Long addr) {
            this.address = addr;
[...]

PR #1050, opened 2021-05-24
Update LLVM examples for LLVM 12 and libffi

supergrecko:
As mentioned in #833 I have updated the LLVM samples.

###### Changes in this patch:

1. Moved the different samples into their own directories. It quickly becomes confusing which files are related to each other, especially when the EmitBitcode example is 2 Java files.
2. Updated Factorial sample for LLVM 12 - one of the passes that sample used is now gone, so the old one wouldn't compile. 
    - I will also be updating this sample to use the bindings for the new LLVM pass manager which I recently landed upstream when LLVM 13 releases.
3. Added steps for each sample describing what each step does. It was previously a bit difficult to differentiate between what each line of code was used for.
4. Added a new sample using the new OrcJIT v2 API which was added in LLVM 12. This uses the new libffi preset to call the JIT'ed functions

I suppose we should also change the sample code shown in `llvm/README.md`.

I also thought about adding the samples to CI so we're sure the samples still run when we update code.

Let me know what you think.


saudet:
> As mentioned in #833 I have updated the LLVM samples.
> ###### Changes in this patch:
> 
> 1. Moved the different samples into their own directories. It quickly becomes confusing which files are related to each other, especially when the EmitBitcode example is 2 Java files.

I find that splitting everything up in multiple directories make things more complicated. @Neiko2002 Did it for LLVM vs Clang, and @yukoba did it for Polly, which is not too confusing because those correspond to different LLVM modules, but I'm not following what the benefit is for the changes proposed here. Isn't there a way to make each sample stand alone? They are not even 100 lines long. Why do they need to be related to each other? If you're trying to create a structure for some high-level API, that code could be better placed as part of another project like LLVM4J (whose repository you could move under this organization if you wished BTW).

> 2. Updated Factorial sample for LLVM 12 - one of the passes that sample used is now gone, so the old one wouldn't compile.
>        
>    * I will also be updating this sample to use the bindings for the new LLVM pass manager which I recently landed upstream when LLVM 13 releases.
> 
> 3. Added steps for each sample describing what each step does. It was previously a bit difficult to differentiate between what each line of code was used for.
> 
>  4. Added a new sample using the new OrcJIT v2 API which was added in LLVM 12. This uses the new libffi preset to call the JIT'ed functions
> 
> 
> I suppose we should also change the sample code shown in `llvm/README.md`.
> 
> I also thought about adding the samples to CI so we're sure the samples still run when we update code.
> 
> Let me know what you think.

The rest sounds good. :+1: FYI, running samples would only work for builds that do not use a cross compiler, in other words probably only linux-x86_64, macosx-x86_64, and windows-x86_64.

supergrecko:
Thank you for the feedback, I figured a module for each sample would be easier to follow, but if you prefer to keep it as is, then I'm good with that. I could also merge the Bitcode samples into a single file.

> The rest sounds good. :+1:  FYI, running samples would only work for builds that do not use a cross compiler, in other words probably only linux-x86_64, macosx-x86_64, and windows-x86_64.
Yeah that sounds about right. I'll add a ci build for those environments.

supergrecko:
Went ahead and implemented running the various samples with Maven profiles which will conditionally set the exec.mainClass

saudet:
Ah, yes, we need to hack the rpath with Autoconf on Mac. Try with this patch:
```diff
--- a/libffi/cppbuild.sh
+++ b/libffi/cppbuild.sh
@@ -84,6 +84,7 @@ case $PLATFORM in
         make install-strip
         ;;
     macosx-*)
+        sedinplace 's/\\\$rpath/@rpath/g' configure
         CC="clang" ./configure --prefix="$INSTALL_PATH" --disable-multi-os-directory
         make -j $MAKEJ
         make install-strip
```

supergrecko:
> Ah, yes, we need to hack the rpath with Autoconf on Mac. Try with this patch:
> 
> ```diff
> --- a/libffi/cppbuild.sh
> +++ b/libffi/cppbuild.sh
> @@ -84,6 +84,7 @@ case $PLATFORM in
>          make install-strip
>          ;;
>      macosx-*)
> +        sedinplace 's/\\\$rpath/@rpath/g' configure
>          CC="clang" ./configure --prefix="$INSTALL_PATH" --disable-multi-os-directory
>          make -j $MAKEJ
>          make install-strip
> ```

The samples run off of the libffi packages which are on Sonatype OSS, shouldn't that patch be merged upstream first (or is it not supposed to be)?

saudet:
Sure, I've applied the patch in commit https://github.com/bytedeco/javacpp-presets/commit/82b9be756f3a772fc9ba4115d6b1f602ac45a392 and updated snapshots are up.

supergrecko:
libffi seems to be working - getting a segfault in the sample on macos, will investigate.

supergrecko:
> libffi seems to be working - getting a segfault in the sample on macos, will investigate.

This should be fixed now. Let me know if there's anything else.

PR #1052, opened 2021-06-01
Annotate the LLVM preset with NoException

supergrecko:
The LLVM/clang project does not use C++ exceptions. We're also only mapping the C interface, so it's safe to assume there will be no exceptions thrown here.


saudet:
The C++ API used in the following files probably throws exceptions but maybe not. Could you confirm?
https://github.com/bytedeco/javacpp-presets/tree/master/llvm/src/main/resources/org/bytedeco/llvm/include

supergrecko:
> The C++ API used in the following files probably throws exceptions but maybe not. Could you confirm?
> https://github.com/bytedeco/javacpp-presets/tree/master/llvm/src/main/resources/org/bytedeco/llvm/include

You're right, the code in FullOptimization.h makes use of exceptions. Seeing as both functions return void, we could return an LLVMErrorRef instead and drop the exceptions, similar to how LLVM does error handling.

Another option would be to enable exception handling for these functions.

Thoughts? 
/cc @yukoba

saudet:
Handing exceptions and returning error codes sounds good to me.

yukoba:
@supergrecko 

> You're right, the code in FullOptimization.h makes use of exceptions. Seeing as both functions return void, we could return an LLVMErrorRef instead and drop the exceptions, similar to how LLVM does error handling.

Please change it.

PR #1056, opened 2021-06-16
Adding preset for leapmotion controller

seem-less:


leapmotion/platform/pom.xml
```diff
@@ -0,0 +1,158 @@
...
+      <artifactId>${javacpp.moduleId}</artifactId>
+      <version>${project.version}</version>
+      <classifier>${javacpp.platform.linux-x86_64}</classifier>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>${javacpp.moduleId}</artifactId>
+      <version>${project.version}</version>
+      <classifier>${javacpp.platform.macosx-x86_64}</classifier>
+    </dependency>
```
supergrecko, line 69: Judging from the current cppbuild.sh we're currently building `windows-x86_64` and `windows-x86`. The other platforms which are not supported yet should be removed from the dependency list.

leapmotion/platform/pom.xml
```diff
@@ -0,0 +1,158 @@
...
+                  <file>${project.build.directory}/${project.artifactId}.jar</file>
+                  <moduleInfoSource>
+                    module org.bytedeco.${javacpp.moduleId}.platform {
+                      requires static org.bytedeco.${javacpp.moduleId}.android.arm;
+                      requires static org.bytedeco.${javacpp.moduleId}.android.arm64;
+                      requires static org.bytedeco.${javacpp.moduleId}.android.x86;
+                      requires static org.bytedeco.${javacpp.moduleId}.android.x86_64;
+                      requires static org.bytedeco.${javacpp.moduleId}.linux.x86;
+                      requires static org.bytedeco.${javacpp.moduleId}.linux.x86_64;
+                      requires static org.bytedeco.${javacpp.moduleId}.macosx.x86_64;
```
supergrecko, line 145: Similarly to how we're only adding dependencies for existing platforms, we only want the available platforms listed here. These entries should be removed until more platforms are added.

leapmotion/README.md
```diff
@@ -0,0 +1,29 @@
...
+<sup>Build status for all platforms:</sup> [![leapmotion](https://github.com/bytedeco/javacpp-presets/workflows/leapmotion/badge.svg)](https://github.com/bytedeco/javacpp-presets/actions?query=workflow%3Aleapmotion)  <sup>Commercial support:</sup> [![xscode](https://img.shields.io/badge/Available%20on-xs%3Acode-blue?style=?style=plastic&logo=appveyor&logo=data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAEAAAABACAMAAACdt4HsAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAAZQTFRF////////VXz1bAAAAAJ0Uk5T/wDltzBKAAAAlUlEQVR42uzXSwqAMAwE0Mn9L+3Ggtgkk35QwcnSJo9S+yGwM9DCooCbgn4YrJ4CIPUcQF7/XSBbx2TEz4sAZ2q1RAECBAiYBlCtvwN+KiYAlG7UDGj59MViT9hOwEqAhYCtAsUZvL6I6W8c2wcbd+LIWSCHSTeSAAECngN4xxIDSK9f4B9t377Wd7H5Nt7/Xz8eAgwAvesLRjYYPuUAAAAASUVORK5CYII=)](https://xscode.com/bytedeco/javacpp-presets)
```
supergrecko, line 5: We should also add the workflow status badge to the main README.md as well as listing leapmotion as one of the available presets.

pom.xml
```diff
@@ -630,6 +630,7 @@
         <module>cpu_features</module>
         <module>modsecurity</module>
         <module>systems</module>
+        <module>leapmotion</module>
```
supergrecko, line 633: Leapmotion should also be registered as an available module for the platforms you're building. Each JavaCPP platform has its own profile in this pom and leapmotion should be registered in these.

junlarsen (commented):
Looks good so far. Just a few nits and a couple things to add:

1. JavaCPP is built and distributed in CI, meaning there should be a new `.github/workflows/leapmotion.yml` workflow. You can copy one of the existing workflows and go from there. It's usually just a matter of changing a few names.
    - Because the user has to download leapmotion themselves, we also need to do this in CI. You should download leapmotion in the action files found in `.github/actions` for the platforms your preset is available for.
    - Testing this is easily done by re-targeting the workflow's branch against your own fork.
2. The license for leapmotion should be copied into the preset's directory.

PR #1065, opened 2021-07-11
Upgrade LLVM preset to 12.0.1

supergrecko:
The LLVM 12.0.1 release adds Type metadata (https://reviews.llvm.org/D97763) to the LLVM-C API

I had troubles building TVM and Arrow using 12.0.1 locally so those presets have not been updated. It would be great if you could take a closer look at those.

- TVM fails because my linker does not support `-force_load`. I did some googling and it looks like it's an Xcode option. I'm not on macOS, so this sounds like a build system error.
- Arrow builds fine, but JavaCPP appears to produce invalid code. I've attached a log of my stderr here: https://gist.github.com/supergrecko/956d109bf7a1815ba286f6658d1cd8c3
  Let me know if you need more information to investigate. I'm building with Clang-10.0.0-4ubuntu1 and GCC/g++ 9.3.0 using LLD 10.0.0


saudet:
I had to fix a couple of things there actually. Could you rebase on master and try again?

supergrecko:
> I had to fix a couple of things there actually. Could you rebase on master and try again?

Arrow builds are working again. Still getting that linker error with TVM. It also fails to build 12.0.0, so it's probably not a regression.

saudet:
TVM builds fine on GitHub Actions, so I guess it's OK. Let's just update it!

BTW, instead of a separate workflow, let's try to get those examples running as part of the build of each platform, so they can succeed even from pull requests like this.

supergrecko:
> BTW, instead of a separate job, let's try to get those examples running as part of the build of each platform, so they can succeed even from pull requests like this.

Okay, I will look into this.

supergrecko:
The samples can run in CI by running it in the same job on the next stage. Got this working on a forked branch here https://github.com/supergrecko/javacpp-presets/blob/run-samples-in-ci/.github/workflows/llvm.yml#L42

My only concern is that the config will become huge if more samples are added like this (e.g. clang ones). Maybe we could think about integrating a sample execution step into the .github/actions scripts?

saudet:
My thinking here is that we should be able to use a Maven Plugin like Surefire, Failsafe, or Invoker, such that everything ends up in the pom.xml files somewhere instead of GitHub Actions, probably something like a mix of how JavaCPP and JavaCV do it:
https://github.com/bytedeco/javacpp/blob/master/pom.xml#L553
https://github.com/bytedeco/javacv/blob/master/platform/pom.xml#L194

PR #1071, opened 2021-08-20
Synchronize and prevent repeated package caching in all presets

HannesWell:
With this PR I suggest to store `pythonFile` computed in the `cpython.presets.python.cachePackage()` in a static variable and to return it if it is already computed to prevent caching of the python packages multiple times.
I assume the result should be the same in each call?

This method is called twice when using the embeddedpython.Python class (in a stripped variant) and each call takes at least one third of the 2 seconds initialization time of `embeddedpython.Python`. So also caching the path would speed up the initialization significantly.

Actually this method could be synchronized to prevent concurrent caching of the packages. If you agree I can update this PR accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Synchronize `cachePackage()` and prevent repeated package caching in all presets ([pull #1071](https://github.com/bytedeco/javacpp-presets/pull/1071))
  * Build FFmpeg with VA-API enabled and bundle its libraries to avoid loading issues ([issue bytedeco/javacv#1188](https://github.com/bytedeco/javacv/issues/1188))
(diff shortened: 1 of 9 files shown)


saudet:
Right, that and synchronizing it make it more consistent with how Loader.load() works. Sounds like a good idea! While we're at it, could you do the same for numpy, scipy, tensorflow, gym, opencv, llvm, and tvm as well?

HannesWell:
> Right, that and synchronizing it make it more consistent with how Loader.load() works. Sounds like a good idea! While we're at it, could you do the same for numpy, scipy, tensorflow, gym, opencv, llvm, and tvm as well?
Great.
Sure. I have applied it to the other mentioned presets and also searched for other cachePackage() methods but found none. So I think those are all.

Issue #1091, opened 2021-10-15
Shipping LLVM native libraries on mac/arm64


oxisto:
Great project! We are currently using it to natively access LLVM in our code analysis project. I was wondering.. Would it be possible to ship the LLVM native binaries also for mac/arm64? It seems that this particular combination is missing. Compiling and running it is no issue, so at least there should not be a technical blocker for it.

Keep up the good work!

saudet:
Everything is on the following page, but in this case we only need to add the entries for "macosx-arm64", obviously:
https://github.com/bytedeco/javacpp-presets/wiki/Create-New-Presets

PR #1092, opened 2021-10-16
Building LLVM on mac-arm64

oxisto:
First attempt to build LLVM on mac-arm64. Not sure, if I understood the profile system correctly.

Closes https://github.com/bytedeco/javacpp-presets/issues/1091


saudet:
Yes, the profiles need to be updated too. It's essentially the same as I did for JavaCPP itself in commit https://github.com/bytedeco/javacpp/commit/87a619482f8c4c967a340638931817f7cf8afbc5.

oxisto:
@saudet I might need your help on this one. I *think* I got the profiles correct now.

Do I need to change https://github.com/bytedeco/javacpp-presets/blob/fb4bba08164174377bb625bc9823b10114a6187c/llvm/cppbuild.sh#L90-L95 this as well to split it up and pass `-arch` or is this taken care of on another level?

oxisto:
> @saudet I might need your help on this one. I _think_ I got the profiles correct now.
> 
> Do I need to change
> 
> https://github.com/bytedeco/javacpp-presets/blob/fb4bba08164174377bb625bc9823b10114a6187c/llvm/cppbuild.sh#L90-L95
> 
> this as well to split it up and pass `-arch` or is this taken care of on another level?

I think I got now. Should hopefully run now.

oxisto:
> At first glance, that looks alright, thanks!
> 
> For consistency though, let's add the `macosx-arm64` lines before the ones for `macosx-x86_64`.

I can switch that, no problem.

There seems to be a build error on llvm though :(

```
/bin/sh: ../../../../bin/clang-tblgen: Bad CPU type in executable
7079
[  8%] Building arm_cde.h...
7080
make[2]: *** [tools/clang/lib/Headers/arm_bf16.h] Error 126
7081
make[2]: *** Waiting for unfinished jobs....
7082
[INFO] ------------------------------------------------------------------------
7083
/bin/sh: ../../../../bin/clang-tblgen: Bad CPU type in executable
```

saudet:
Since we're cross compiling, we need to prebuild those tools for x86_64 first, as with linux-arm64:
https://github.com/bytedeco/javacpp-presets/blob/master/llvm/cppbuild.sh#L62-L75

oxisto:
> After this gets merged, the snapshots will be available for download: http://bytedeco.org/builds/
> 
> Before we do that though, there are still a lot of places where arm64 is still after x86_64. Please update everything!

I was following the style of the individual files. For example in the cppbuild.sh, x86_64 was already placed before arm64 for Linux. Should I fix these already existing inconsistencies as well?

PR #1093, opened 2021-10-18
Add missing PassBuilder.h header for LLVM

junlarsen:
LLVM 13 landed a new header (Transforms/PassBuilder.h) for interfacing with LLVMs new Pass Manager

https://reviews.llvm.org/D102136


saudet:
Let's also update the src/gen directory! Thanks

PR #1096, opened 2021-10-28
quick update mkl-dnn unmaintained note

xwaeaewcrhomesysplug:


saudet:
Well, I don't think this is really necessary anyway. It's pretty clear from the version number that this is the old API, and besides the URL doesn't even appear on the main page: http://bytedeco.org/

If you can fix the formatting though, I'll be happy to merge this.

PR #1109, opened 2021-12-19
Minor upgrade of dependencies for arrow, ffmpeg, cpython

bradh:
Bumps openssl to 1.1.1m, plus minor upgrades to ALSA and freetype


saudet:
Looks good, thanks! BTW, do you happen to know if there are any issues with OpenSSL 3.0.x?

bradh:
> Looks good, thanks! BTW, do you happen to know if there are any issues with OpenSSL 3.0.x?

None that I know of, but the documentation suggests that there will be migration effort (i.e. it might compile, or it might not, but you'll definitely get deprecation warnings). Probably worth waiting until the using-projects declare support for it.

PR #1110, opened 2021-12-20
Update Maven Version For Triton Server README

dyastremsky:
The old version is no longer available, so this ticket updates the Maven version in the README.

### DIFF
--- a/.github/workflows/tritonserver.yml
+++ b/.github/workflows/tritonserver.yml
@@ -17,6 +17,6 @@ env:
 jobs:
   linux-x86_64:
     runs-on: ubuntu-18.04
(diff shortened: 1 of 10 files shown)


saudet:
Thanks! Let's use the archive instead to avoid this happening again:
https://archive.apache.org/dist/maven/maven-3/
/cc @jackyh

dyastremsky:
My pleasure. Makes sense! I switched to the archive link (3.8.4) and tested that it's working

saudet:
It looks like a new version of Triton has also just been released (21.12-py3), let's update that as well?
https://catalog.ngc.nvidia.com/orgs/nvidia/containers/tritonserver

dyastremsky:
Good eye. Done!

PR #1112, opened 2022-01-02
openvino: work in progress

bradh:


saudet:
@Grumpy141 @wureka @themarpe FYI, being able to call OpenVINO from Java may be useful to users of DepthAI.

themarpe:
Awesome @bradh - this would allow us to create `blobconverter` equivalent as we have it now in Python, but completely local.

Thanks saudet for the ping :)

saudet:
Duplicate of #1752

PR #1116, opened 2022-01-07
added the build part for generate uber jar for Tritonserver

jackyh:


tritonserver/samples/pom.xml
```diff
@@ -21,6 +21,28 @@
         </dependency>
     </dependencies>
     <build>
-        <sourceDirectory>.</sourceDirectory>
```
saudet, line 24: We need to leave that line because the source file isn't in the default directory.

jackyh: Yes, added back this line.

PR #1117, opened 2022-01-10
* Fix grammar and some typos

stweil:


saudet:
Thank you! The code in the src/gen and samples subdirectories though is copied from upstream as is, so those need to be fixed upstream.

stweil:
@saudet, I now removed those parts which need to be fixed upstream.

saudet:
The ones for videoInput all come from the same place upstream, so I've reverted that one too:
https://github.com/ofTheo/videoInput/blob/master/videoInputSrcAndDemos/libs/videoInput/videoInput.h

PR #1119, opened 2022-01-11
Replicate Triton Simple Preprocessor Flags

dyastremsky:
Update the tritonserver simple example to replicate the original example's preprocessor flag behavior.

### DIFF
--- a/tritonserver/samples/Simple.java
+++ b/tritonserver/samples/Simple.java
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -36,7 +36,6 @@ import static org.bytedeco.tritonserver.global.tritonserver.*;
 
 public class Simple {
     static final double TRITON_MIN_COMPUTE_CAPABILITY = 6.0;
-    private static final boolean TRITON_ENABLE_GPU = true;
 
     static void FAIL(String MSG) {
         System.err.println("Cuda failure: " + MSG);
@@ -144,7 +143,7 @@ public class Simple {
               actual_memory_type.put(0, requested_memory_type);
             }
 
-            switch (TRITON_ENABLE_GPU ? actual_memory_type.get(): TRITONSERVER_MEMORY_CPU) {
+            switch (actual_memory_type.get()) {
               case TRITONSERVER_MEMORY_CPU_PINNED: {
                 int err = cudaSetDevice((int)actual_memory_type_id.get());
                 if ((err != cudaSuccess) && (err != cudaErrorNoDevice) &&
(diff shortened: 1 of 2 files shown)


dyastremsky:
@jbkyang-nvi

saudet:
This can work in the sense that we ensure that no functions from CUDA are called, but it doesn't remove the dependency on CUDA. The calls to CUDA are still being compiled, they're just not getting called. Java does not have a preprocessor, so when we use some API directly, there is no easy way to make those dependencies optional. Triton would need to wrap those calls in a higher-level API, and then we could easily make the dependency on CUDA optional.

dyastremsky:
Thanks for the feedback, Samuel! We've opted to split the simple example into a CPU-only and GPU/CUDA-enabled one for simplicity (pardon the pun).

@jbkyang-nvi The PR is ready for review.

tritonserver/samples/Simple.java
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2021-2022, NVIDIA CORPORATION. All rights reserved.
```
jbkyang-nvi, line 1: `NVIDIA CORPORATION & AFFILIATES`

tritonserver/samples/SimpleCPUOnly.java
```diff
@@ -0,0 +1,754 @@
+// Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
```
jbkyang-nvi, line 1: `NVIDIA CORPORATION & AFFILIATE`

jbkyang-nvi (approved):
thanks. Looks good to me

tritonserver/samples/Simple.java
```diff
@@ -821,10 +818,10 @@ public void reset(Pointer p) {
               cudaMemcpy(dst, input1_data, input1_size, cudaMemcpyHostToHost),
               "setting INPUT1 data in pinned memory");
         }
-      }
 
-      input0_base = use_cuda_memory ? input0_gpu : input0_data;
-      input1_base = use_cuda_memory ? input1_gpu : input1_data;
+        input0_base = use_cuda_memory ? input0_gpu : input0_data;
+        input1_base = use_cuda_memory ? input1_gpu : input1_data;
```
saudet, line 823: We might as well also remove the redundant check for use_cuda_memory here?
BTW, the only reason I did it this way is because the original code is also like that:
https://github.com/triton-inference-server/server/blob/main/src/servers/simple.cc
So that should probably be "fixed" there as well...

tritonserver/samples/Simple.java
dyastremsky: Fixed!

tritonserver/samples/SimpleCPUOnly.java
dyastremsky: Fixed.

tritonserver/samples/Simple.java
dyastremsky: Nice find! Removed. I'll update in simple.cc as well, thanks for the heads up!

PR #1121, opened 2022-01-12
[WIP] Temporary Android branch build

themarpe:
Currently targets a temporary development branch of DepthAI with Android support.
Afterwards, when Android support is mainlined the changes will be able to be merged

### DIFF
--- a/.github/workflows/depthai.yml
+++ b/.github/workflows/depthai.yml
@@ -16,6 +16,26 @@ env:
   CI_DEPLOY_PASSWORD: ${{ secrets.CI_DEPLOY_PASSWORD }}
   STAGING_REPOSITORY: ${{ secrets.STAGING_REPOSITORY }}
 jobs:
+  android-arm:
+    runs-on: ubuntu-18.04
+    container: centos:7
+    steps:
+      - uses: bytedeco/javacpp-presets/.github/actions/deploy-centos@actions
+  android-arm64:
+    runs-on: ubuntu-18.04
+    container: centos:7
+    steps:
+      - uses: bytedeco/javacpp-presets/.github/actions/deploy-centos@actions
+  android-x86:
+    runs-on: ubuntu-18.04
+    container: centos:7
+    steps:
+      - uses: bytedeco/javacpp-presets/.github/actions/deploy-centos@actions
+  android-x86_64:
+    runs-on: ubuntu-18.04
+    container: centos:7
+    steps:
+      - uses: bytedeco/javacpp-presets/.github/actions/deploy-centos@actions
   linux-armhf:
     runs-on: ubuntu-18.04
     container: ubuntu:bionic
@@ -41,7 +61,7 @@ jobs:
     steps:
       - uses: bytedeco/javacpp-presets/.github/actions/deploy-macosx@actions
   redeploy:
-    needs: [linux-armhf, linux-arm64, linux-x86, linux-x86_64, macosx-x86_64]
+    needs: [android-arm, android-arm64, android-x86, android-x86_64, linux-armhf, linux-arm64, linux-x86, linux-x86_64, macosx-x86_64]
     runs-on: ubuntu-18.04
     steps:
       - uses: bytedeco/javacpp-presets/.github/actions/redeploy@actions
--- a/depthai/README.md
+++ b/depthai/README.md
@@ -46,7 +46,7 @@ We can use [Maven 3](http://maven.apache.org/) to download and install automatic
         <dependency>
             <groupId>org.bytedeco</groupId>
             <artifactId>depthai-platform</artifactId>
-            <version>2.13.3-1.5.7-SNAPSHOT</version>
+            <version>2.13.3-7abd507db16ea96b5f5a67d11f09f59b50d41827-1.5.7-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.bytedeco</groupId>
--- a/depthai/cppbuild.sh
+++ b/depthai/cppbuild.sh
@@ -7,21 +7,28 @@ if [[ -z "$PLATFORM" ]]; then
     exit
 fi
 
-DEPTHAI_VERSION=2.13.3
+DEPTHAI_VERSION_COMMIT=7abd507db16ea96b5f5a67d11f09f59b50d41827
+DEPTHAI_VERSION=2.13.3-7abd507db16ea96b5f5a67d11f09f59b50d41827
 LIBUSB_VERSION=1.0.22
-download https://github.com/luxonis/depthai-core/releases/download/v$DEPTHAI_VERSION/depthai-core-v$DEPTHAI_VERSION.tar.gz depthai-core-v$DEPTHAI_VERSION.tar.gz
-download http://sourceforge.net/projects/libusb/files/libusb-1.0/libusb-$LIBUSB_VERSION/libusb-$LIBUSB_VERSION.tar.bz2/download libusb-$LIBUSB_VERSION.tar.bz2
+
+# download https://github.com/luxonis/depthai-core/releases/download/v$DEPTHAI_VERSION/depthai-core-v$DEPTHAI_VERSION.tar.gz depthai-core-v$DEPTHAI_VERSION.tar.gz
+# download http://sourceforge.net/projects/libusb/files/libusb-1.0/libusb-$LIBUSB_VERSION/libusb-$LIBUSB_VERSION.tar.bz2/download libusb-$LIBUSB_VERSION.tar.bz2
+
+# download https://github.com/luxonis/depthai-core/archive/7abd507db16ea96b5f5a67d11f09f59b50d41827.tar.gz depthai-core-$DEPTHAI_VERSION_COMMIT.tar.gz
 
 mkdir -p $PLATFORM
 cd $PLATFORM
 mkdir -p include lib bin
 INSTALL_PATH=`pwd`
-echo "Decompressing archives..."
-tar --totals -xzf ../depthai-core-v$DEPTHAI_VERSION.tar.gz
-tar --totals -xjf ../libusb-$LIBUSB_VERSION.tar.bz2
 
-cd depthai-core-v$DEPTHAI_VERSION
-#patch -Np1 < ../../../depthai.patch
+# echo "Decompressing archives..."
+# tar --totals -xzf ../depthai-core-$DEPTHAI_VERSION_COMMIT.tar.gz
+
+git clone https://github.com/luxonis/depthai-core.git depthai-core-$DEPTHAI_VERSION_COMMIT --recursive
+cd depthai-core-$DEPTHAI_VERSION_COMMIT
+git checkout $DEPTHAI_VERSION_COMMIT
+git submodule update --init --recursive
+# patch
 sedinplace '/find_package(Git/d' CMakeLists.txt cmake/GitCommitHash.cmake shared/depthai-bootloader-shared.cmake shared/depthai-shared.cmake
 sedinplace '/protected:/d' include/depthai/pipeline/Node.hpp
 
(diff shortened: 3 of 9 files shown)


saudet:
BTW, instead of coming up with custom toolchains, try to use use what's in CMake since 3.21:
https://developer.android.com/ndk/guides/cmake

themarpe:
@saudet 
I guess the CI not triggering because it was based off an older commit / had conflicts 
`6f8889c3f3a7b91fe55d96195dd976b643da68f1` ( * Upgrade presets for Tesseract 5.0.0, DepthAI 2.13.3)

I've based it of and basically did a revert to target older depthai version. (and seems like it needs approval again, probably because of the force push - not sure / or was it always requiring approval, per commit)

saudet:
BTW, the workflows are already running from your fork, you don't need to make them work here:
https://github.com/luxonis/javacpp-presets/actions

saudet:
And the reason why we need to approve commits from external contributors is for security reasons:
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
This isn't needed when you make contributions to forks you own, of course.

themarpe:
Okay, seems like I got the build working. (https://github.com/luxonis/javacpp-presets/actions/runs/1696048933)
Where should I look for Android prebuilts? (or are those are only uploaded when CI runs on your end?)

saudet:
For security reasons again, the binaries from PR builds don't get uploaded anywhere. You can however upload the ones built from your fork anywhere you want, including Sonatype OSSRH: https://central.sonatype.org/publish/publish-guide/

That said, I could merge your PR in this repository and it would start uploading snapshots that way with my credentials. Now, can you rebase this on DepthAI 2.14.x? Or is this going to stay on 2.13.x for a while? If it's the latter, I'll create a special branch where to merge stuff for now until it's ready to be merged to the main branch.

themarpe:
All makes sense. Though that these get uploaded as "temporary builds" somewhere. 

Hmm, i dont want to pollute the reguar versioning - right now I added the comit after the version, would that do to create builds that arent version only?

Otherwise it might be better if I upload them manually somewhere... How can I get out the built artifacts? (Or I see that I might be able to just supply the relevant secrets and rerun? I assume usign the above mentioned Sonatype repo. Would that be the simplest repo to get some snapshots uploaded?)

 Yeah, regarding versioning that was the main issue - most likely won't be mainlined soon (I could expedite it a bit, bot it still requires a bit work around everything and testing...), so wanted just to create prebuilts based off this special branch that supports Android.

themarpe:
I checked and it seems that Artifactory has a Maven repository option - I could spin a repo there easily - would credentials to that be adequate to get the builds deployed to it?

saudet:
> Hmm, i dont want to pollute the reguar versioning - right now I added the comit after the version, would that do to create builds that arent version only?

You can give them a different version, by appending the commit hash or whatever, sure, that's fine.

> I checked and it seems that Artifactory has a Maven repository option - I could spin a repo there easily - would credentials to that be adequate to get the builds deployed to it?

You can do whatever you want from your fork, but if you're not already using something somewhere, the easiest option is probably to stay here on GitHub with GitHub Packages: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry

saudet:
In any case, I've created a branch named "depthai" that is set to the last commit before upgrading to 2.14.0, and rebased this PR on it. We can merge this there anytime you're ready, but please first revert unrelated changes. Thanks!

themarpe:
@saudet thanks a lot, I think this will be the easiest:)

I've force-pushed the changes, which should now be based on `depthai` branch.

.github/workflows/depthai.yml
```diff
@@ -40,8 +40,28 @@ jobs:
...
+  android-x86:
+    runs-on: ubuntu-18.04
+    container: centos:7
+    steps:
+      - uses: bytedeco/javacpp-presets/.github/actions/deploy-centos@actions
+  android-x86_64:
+    runs-on: ubuntu-18.04
+    container: centos:7
+    steps:
+      - uses: bytedeco/javacpp-presets/.github/actions/deploy-centos@actions
```
saudet, line 62: Please put everything for Android *before* the other platforms, in this file, cppbuild.sh, and pom.xml files as well.

depthai/README.md
```diff
@@ -1,7 +1,7 @@
 JavaCPP Presets for DepthAI
 ===========================
 
-[![Gitter](https://badges.gitter.im/bytedeco/javacpp.svg)](https://gitter.im/bytedeco/javacpp) [![Maven Central](https://maven-badges.herokuapp.com/maven-central/org.bytedeco/depthai/badge.svg)](https://maven-badges.herokuapp.com/maven-central/org.bytedeco/depthai) [![Sonatype Nexus (Snapshots)](https://img.shields.io/nexus/s/https/oss.sonatype.org/org.bytedeco/depthai.svg)](http://bytedeco.org/builds/)  
+[![Gitter](https://badges.gitter.im/bytedeco/javacpp.svg)](https://gitter.im/bytedeco/javacpp) [![Maven Central](https://maven-badges.herokuapp.com/maven-central/org.bytedeco/depthai/badge.svg)](https://maven-badges.herokuapp.com/maven-central/org.bytedeco/depthai) [![Sonatype Nexus (Snapshots)](https://img.shields.io/nexus/s/https/oss.sonatype.org/org.bytedeco/depthai.svg)](http://bytedeco.org/builds/)
```
saudet, line 4: Please also remove extra blank spaces added here and in a few other places.

pom.xml
```diff
@@ -765,6 +765,7 @@
       </activation>
       <modules>
         <module>opencv</module>
+        <module>depthai</module>
```
saudet, line 768: Please reinsert it in the same order as the platforms.

saudet (changes requested):
There's also missing entries for Android in these files:
https://github.com/luxonis/javacpp-presets/blob/depthai-android/depthai/platform/pom.xml
https://github.com/luxonis/javacpp-presets/blob/depthai-android/depthai/src/main/java/org/bytedeco/depthai/presets/depthai.java#L47

Normally I'd fix these nits myself, but by default we don't get permission to make changes to pull requests from repositories of organizations, only when it's from a personal repo:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork
Sorry for the trouble!

themarpe:
No issue @saudet - thanks so much for helping out with this!

I've force pushed again - let me know if I missed something or should modify something else:)

pom.xml
```diff
@@ -347,7 +347,7 @@
             <configuration>
               <classifier>${javacpp.platform}${javacpp.platform.extension}</classifier>
               <skipIfEmpty>true</skipIfEmpty>
-              <includes>
```
saudet, line 350: Let's also revert this, even if it shouldn't be there, to help with any merge conflicts later...

depthai/src/main/java/org/bytedeco/depthai/presets/depthai.java
```diff
@@ -324,6 +324,8 @@ public void map(InfoMap infoMap) {
                .put(new Info("std::function<void(std::string,std::shared_ptr<ADatatype>)>").valueTypes("NameMessageCallback"))
                .put(new Info("std::function<void(std::shared_ptr<ADatatype>)>").valueTypes("MessageCallback"))
                .put(new Info("std::function<void()>").valueTypes("Callback"))
+
+               .put(new Info("dai::DeviceInfo::toString").javaText("public native @StdString String toString();"))
```
saudet, line 328: BTW, `dai::DeviceInfo::toString()` doesn't exist in main branch. Is something new in the branch for Android?

.github/workflows/depthai.yml
```diff
@@ -41,7 +61,7 @@ jobs:
     steps:
       - uses: bytedeco/javacpp-presets/.github/actions/deploy-macosx@actions
   redeploy:
-    needs: [linux-armhf, linux-arm64, linux-x86, linux-x86_64, macosx-x86_64]
+    needs: [linux-armhf, linux-arm64, linux-x86, linux-x86_64, macosx-x86_64, android-arm, android-arm64, android-x86, android-x86_64]
```
saudet, line 64: Still need to change the order...

depthai/cppbuild.sh
```diff
@@ -107,6 +84,27 @@ case $PLATFORM in
...
+        $CMAKE -DCMAKE_TOOLCHAIN_FILE=${PLATFORM_ROOT}/build/cmake/android.toolchain.cmake -DANDROID_ABI=x86 -DANDROID_PLATFORM=24 -DCMAKE_INSTALL_PREFIX="$INSTALL_PATH" -DCMAKE_INSTALL_LIBDIR="lib" -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DDEPTHAI_OPENCV_SUPPORT=ON .
+        make -j $MAKEJ
+        make install/strip
+        ;;
+    android-x86_64)
+        $CMAKE -DCMAKE_TOOLCHAIN_FILE=${PLATFORM_ROOT}/build/cmake/android.toolchain.cmake -DANDROID_ABI=x86_64 -DANDROID_PLATFORM=24 -DCMAKE_INSTALL_PREFIX="$INSTALL_PATH" -DCMAKE_INSTALL_LIBDIR="lib" -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DDEPTHAI_OPENCV_SUPPORT=ON .
+        make -j $MAKEJ
+        make install/strip
+        ;;
+
```
saudet, line 107: ... and here ...

saudet (changes requested):
And missing entries here:

depthai/src/main/java/org/bytedeco/depthai/presets/depthai.java
themarpe: Yeah, some improvements on that branch, but they got left behind for now, until mainlined

themarpe: Added because it was failing otherwise with an overload between DeviceInfo toString and base class Pointer...

themarpe:
Force pushed - my bad about whitespace changes. I have diffs for those disabled and VSCode set to automatically trim trialing whitespace, so it kinda slips past my local review

themarpe:
Android not building anymore ... Doesn't seem to detect OpenCV library anymore, which causes it to fail later on.
Any ideas? (the differences were adding android entries to `depthai/platform/depthai.xml` and to preset `depthai.java`, those I think might be relevant)

saudet:
It looks like a problem with the build options. It says OpenCV support isn't enabled:
```
/__w/javacpp-presets/javacpp-presets/depthai/cppbuild/android-arm/include/depthai/pipeline/datatype/ImgFrame.hpp:189:9:
error: static_assert failed due to requirement 'dependent_false<bool>::value' "Library not configured with OpenCV support"
```

saudet:
OpenCV installs its CMake-related files a bit differently for Android.
You'll probably need to set its directory somewhere, for example, see:
https://github.com/bytedeco/javacpp-presets/blob/master/flandmark/cppbuild.sh

themarpe:
Thanks @saudet - fixed one extra issue on DepthAI side, but now the CI is passing fully:)

saudet:
Ok, Android builds are up here:
https://oss.sonatype.org/content/repositories/snapshots/org/bytedeco/depthai/
https://oss.sonatype.org/content/repositories/snapshots/org/bytedeco/depthai-platform/
You can use them in your Gradle builds as explained here: http://bytedeco.org/builds/

/cc @Grumpy141 @wureka

themarpe:
Thanks so much for leading me through this, and deploying this on your end:)

Looking forward to bringing Android to mainline, so this special branch can be merged as well:)

A couple of questions:
 - How should I update the library if there is something that needs to be fixed?
 - Any ideas on the following:
 ```
extern "C" JNIEXPORT jstring JNICALL
Java_com_luxonis_depthaiexampleapp_MainActivity_initXLink(
        JNIEnv* env,
        jobject /* this */) {

    JavaVM* jvm;
    env->GetJavaVM(&jvm);
    // ... use the JavaVM pointer to setup libusb underneath
``` 

Where would be the cleanest place to execute this kind of code? For libusb to work on Android, it requires getting JavaENV / JavaVM pointer so it can scan for devices and ask for permissions (from Android SDK).

I assume that'll be between depthai-core and javacpp - I saw a jnidepthai library/file? being created in the build, which has the required JNI code I assume. Thats a spot I think I'd be able to get the pointer from and call the init function.
Thoughts?

saudet:
Just keep sending pull requests against this branch to have updates merged and deployed!

For the init function, this looks like something that should be kept in depth-core, but if you export a JNI method, you'll need to provide a Java interface and everything to initialize that, which JavaCPP already does, so I'd just make it available as a normal C function, and then we could easily call it from Java with JavaCPP like this for FFmpeg:
https://github.com/bytedeco/javacv/blob/master/src/main/java/org/bytedeco/javacv/FFmpegFrameGrabber.java#L117

saudet:
Besides, there are a lot of developers for Android that do not code in Java, so having an init method that is only available from Java isn't something you want to do.

themarpe:
This is perfect! Just what I was looking for. Thanks!

themarpe:
I agree, just added the changes to `depthai-core` and opened the PR: https://github.com/bytedeco/javacpp-presets/pull/1126

Issue #1099, opened 2021-11-11
Enable ffmpeg with libwebp


minzhang-haus:
Trying to use javacv to generate animated `webp` from video. But it doesn't look like javacv supports it. In the cppbuild.sh, libwebp is not enabled either. 

Is there plan to support that? 

the cmd that i tried to achieve with javacv: https://gist.github.com/witmin/1edf926c2886d5c8d9b264d70baf7379

saudet:
We just need to add builds for it, yes. Contributions are welcome!

bradh:
@saudet Working this. Do you want to assign it to me?

saudet:
BTW, configure commands to build libwebp itself are already in the script for Leptonica:
https://github.com/bytedeco/javacpp-presets/blob/master/leptonica/cppbuild.sh

bradh:
> BTW, configure commands to build libwebp itself are already in the script for Leptonica: https://github.com/bytedeco/javacpp-presets/blob/master/leptonica/cppbuild.sh
Wish I'd seen this comment earlier. I spent most of the weekend working it...

saudet:
Thanks to @bradh, we now have support for WebP encoding!
@minzhang-haus Please give it a try with the snapshots: http://bytedeco.org/builds/

minzhang-haus:
@saudet @bradh thank you guys! will give a try 👍

PR #1125, opened 2022-01-15
update to ffmpeg 5.0

bradh:


saudet:
It looks like autoreconf is broken in the new Windows image of GitHub Actions.
Since the only thing using that is libmfx, let's move it to CMake...

ffmpeg/README.md
```diff
@@ -1,36 +1,38 @@
-# JavaCPP Presets for FFmpeg
+JavaCPP Presets for FFmpeg
+==========================
```
bradh, line 2: I would ask that these markdown changes not be reverted. They show up as red squiggle errors in my editor (i.e. fail markdown linting).

saudet: There's no reason to make it inconsistent with the other README.md files. If there's a good reason to change it though, we'll do it for all of them, and that's going to be for another pull request if you want to do that. However, maybe it's just your linting tool that's buggy. How do you know that's not the case?

bradh: I think the tool is reporting valid violations of these rules:

https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md014---dollar-signs-used-before-commands-without-showing-output

https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md006---consider-starting-bulleted-lists-at-the-beginning-of-the-line

https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md003---header-style

I accept that could be a different commit.

saudet: Yeah, let's do that in different commits, for the other repositories too!

BTW, it doesn't appear to violate setext_with_atx, that looks alright. The reason I do it that way is that it looks better with setext, but there aren't any for level 3. I guess I'm not the only one to feel that way.

saudet:
So, this looks good to merge? We can work on the markdown and issue #1099 later and open more pull requests for that.

PR #1126, opened 2022-01-17
DepthAI (Android experimental) - Initialize with JavaVM function

themarpe:
Added an initialize function in depthai, which enables passing JavaVM pointer to underlying libusb library for Android usecase

### DIFF
--- a/depthai/README.md
+++ b/depthai/README.md
@@ -46,7 +46,7 @@ We can use [Maven 3](http://maven.apache.org/) to download and install automatic
         <dependency>
             <groupId>org.bytedeco</groupId>
             <artifactId>depthai-platform</artifactId>
-            <version>2.13.3-7abd507db16ea96b5f5a67d11f09f59b50d41827-1.5.7-SNAPSHOT</version>
+            <version>2.13.3-b83eac2e5b623e357f1ef5a424f00fb1d5d0ac12-1.5.7-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.bytedeco</groupId>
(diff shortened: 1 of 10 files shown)


saudet:
Could you also update the files under src/gen to see what that looks like mapped to Java?

themarpe:
Updated - will `initialize(Pointer javavm)` and `initialize(@StdString String additionalInfo)` or `initialize(@StdString BytePointer additionalInfo)` overload correctly when passing `initialize(Loader.getJavaVM())`? Otherwise there is still the full form, so not too much of an issue
```
@Namespace("dai") public static native @Cast("bool") boolean initialize();
@Namespace("dai") public static native @Cast("bool") boolean initialize(@StdString BytePointer additionalInfo, @Cast("bool") boolean installSignalHandler/*=true*/, Pointer javavm/*=nullptr*/);
@Namespace("dai") public static native @Cast("bool") boolean initialize(@StdString BytePointer additionalInfo);
@Namespace("dai") public static native @Cast("bool") boolean initialize(@StdString String additionalInfo, @Cast("bool") boolean installSignalHandler/*=true*/, Pointer javavm/*=nullptr*/);
@Namespace("dai") public static native @Cast("bool") boolean initialize(@StdString String additionalInfo);
@Namespace("dai") public static native @Cast("bool") boolean initialize(Pointer javavm);
```

saudet:
Since `Loader.getJavaVM()` returns `Pointer` it can't match with any of the other overloads, so that looks alright, yes.

themarpe:
Great - then I think this PR should work as intended:)

PR #1132, opened 2022-01-22
Add IPC/SHM to systems

wmeddie:
Adds `shmget`  and related functions to linux and maxosx.


saudet:
Looks good to me! OK to merge?

wmeddie:
Do we need to add any tests?

saudet:
Nah, just that it compiles is good enough. I run the samples manually before releases to make sure loading works, and that's enough. That reminds me @matsjla is supposed to be working on automating that as part of the Maven build. Any updates?

wmeddie:
I see. If so then it is ready to go from my side.

PR #1150, opened 2022-02-17
Change the absolute path to a relative path about postbuild.sh execute in pom.xml

devjeonghwan:
I got error when building openblas using mingw on Windows.

Because, when i build on Windows, the maven variable(project.basedir)'s path separator is backslash, so bash can't find the shell file.

### DIFF
--- a/mkl-dnn/pom.xml
+++ b/mkl-dnn/pom.xml
@@ -47,7 +47,7 @@
               <skip>${javacpp.compiler.skip}</skip>
               <buildCommand>
                 <program>bash</program>
-                <argument>${project.basedir}/postbuild.sh</argument>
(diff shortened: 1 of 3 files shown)


saudet:
The cppbuild.sh script gets executed in exactly the same way though. How come that doesn't fail?
https://github.com/bytedeco/javacpp-presets/blob/master/pom.xml#L248
https://github.com/bytedeco/javacpp-presets/blob/master/pom.xml#L270

devjeonghwan:
I had manually run cppbuild.sh (from wiki). I'll test it soon and let you know. (probably the same error)

devjeonghwan:
image
Occurred same error.. xD

PR #1153, opened 2022-02-22
Preset for Bullet Physics SDK

### LINKED ISSUE #828: Presets for Bullet Phyics SDK
Would be nice to call Bullet from the JVM: https://github.com/bulletphysics/bullet3

mode89:
Should resolve #828 :)


mode89:
Should resolve #828 :)

saudet:
If you're ready to start testing the builds, please also add a workflow file here:
https://github.com/bytedeco/javacpp-presets/tree/master/.github/workflows

bullet/src/main/java/org/bytedeco/bullet/presets/BulletCollision.java
```diff
@@ -0,0 +1,152 @@
...
+                    "NO_VIRTUAL_INTERFACE",
+                    "PFX_USE_FREE_VECTORMATH",
+                    "__SPU__"
+                ).define(false))
+
+            .put(new Info("btAlignedObjectArray<btBvhSubtreeInfo>").pointerTypes("btAlignedObjectArray_btBvhSubtreeInfo"))
+            .put(new Info("btAlignedObjectArray<btCollisionObject*>").pointerTypes("btAlignedObjectArray_btCollisionObjectPointer"))
+            .put(new Info("btAlignedObjectArray<btIndexedMesh>").pointerTypes("btAlignedObjectArray_btIndexedMesh"))
+            .put(new Info("btAlignedObjectArray<btPersistentManifold*>").pointerTypes("btAlignedObjectArray_btPersistentManifoldPointer"))
+            .put(new Info("btAlignedObjectArray<btQuantizedBvhNode>").pointerTypes("btAlignedObjectArray_btQuantizedBvhNode"))
```
saudet: We don't need to map all the characters from the C++ type name. Here, for example, since it's obvious that btAlignedObjectArray is some sort of alternative to std::vector, we can name those instances more simply btBvhSubtreeInfoArray, btCollisionObjectArray (since there isn't another btCollisionObjectArray, I don't think we need "Pointer" in there), btIndexedMeshArray, etc.

saudet: We don't need to map all the characters from C++ type names. Here, for example, since it's obvious that btAlignedObjectArray is some sort of alternative to std::vector, we can name those instances more simply btBvhSubtreeInfoArray, btCollisionObjectArray (since there isn't another btCollisionObjectArray, I don't think we need "Pointer" in there), btIndexedMeshArray, etc.

saudet: There's some sort of bug in GitHub, this isn't a comment an outdated commit.

mode89: Hi Samuel :) Sounds good. I will update the mappings.

saudet:
This looks to merge! Please let me know if there is anything you'd like to change though. Thanks!

mode89:
Thank you Samuel for helping with the builds :)

I have discovered that there are bunch of headers that are referenced in the official [examples](https://github.com/bulletphysics/bullet3/tree/master/examples), but haven't been parsed yet. If you don't mind, I would like to add them through this PR, otherwise I'm fine with making a separate PR. I should be done with it by the end of this week.

saudet:
Sounds good, we can wait! Thanks for letting me know

mode89:
Hi @saudet :)

Do you happen to know how I can link bullet against opencl preset? Is there any guide or an example?

saudet:
As reference you can look at the presets for DNNL: https://github.com/bytedeco/javacpp-presets/tree/master/dnnl

mode89:
Hi Samuel :)

I think, it is ready. I have mapped almost all of the available headers.

saudet:
Thanks! Now why isn't the GitHub Actions workflow running...

saudet:
Could you try to run the workflow from your fork?

bullet/src/main/resources/org/bytedeco/bullet/include/clew_stubs.h
```diff
@@ -0,0 +1,120 @@
...
+ * limitations under the License.
+ */
+
+
+
+// This #ifdef disables these stubs during native code compilation,
+// as the stubs should be used for generation of the Java-side
+// code only, and the native code should use the original definitions
+// provided by clew's header.
+#ifdef XXXXXXXXXX
```
saudet, line 29: Are you saying JavaCPP isn't able to parse that header file for clew? What's wrong with the output? We should try to fix that first before adding some custom script. It looks like it's basically just OpenCL though. We can't link with the presets for OpenCL and use those functions from there?

saudet:
Huh, it looks like GitHub Actions won't run anymore for commits made by contributors not part of the organization.

bullet/src/main/resources/org/bytedeco/bullet/include/clew_stubs.h
saudet: This clew stuff also doesn't seem to work on Mac and 32-bit Windows. What's the reason we can't use standard OpenCL?

mode89: Sure, let's try to fix it :)

The reason why I did this is that each of those function names is actually a [macro definition](https://github.com/bulletphysics/bullet3/blob/22d9eec148d3a7dec9d7b10a4344808bf766f73a/src/clew/clew.h#L2628) that is substituted by a [global variable](https://github.com/bulletphysics/bullet3/blob/22d9eec148d3a7dec9d7b10a4344808bf766f73a/src/clew/clew.h#L2526) of a [function type](https://github.com/bulletphysics/bullet3/blob/22d9eec148d3a7dec9d7b10a4344808bf766f73a/src/clew/clew.h#L1972). Sorry, I just couldn't figure out a better way.

saudet: Ok, but why do we need any of this? What's missing from the presets for OpenCL? OpenCV is another library that uses OpenCL but doesn't need to link with any external version of OpenCL. It works just fine with the presets for OpenCL at runtime though.

mode89: > This clew stuff also doesn't seem to work on Mac and 32-bit Windows. What's the reason we can't use standard OpenCL?

Yeah, initially I thought about plugin in the javacpp's OpenCL, but than I found, that `clew` support is [hardcoded](https://github.com/bulletphysics/bullet3/blob/22d9eec148d3a7dec9d7b10a4344808bf766f73a/src/Bullet3OpenCL/CMakeLists.txt#L3) into the build script, and `clew`'s source code itself is [included](https://github.com/bulletphysics/bullet3/tree/master/src/clew) into the source code of `bullet`. So, I thought that maybe there are some special reasons behind that, and finally decided to just map `clew.h`.

I think, it might be possible to override it by patching the build script and maybe a bunch of cpp files. I haven't tried it though. Should I go that way?

mode89: Actually, I didn't find a working OpenCL example. They aren't even included into the [cmake build](https://github.com/bulletphysics/bullet3/blob/22d9eec148d3a7dec9d7b10a4344808bf766f73a/examples/ExampleBrowser/CMakeLists.txt#L151). I tried to add them manually, but there were compilation errors in the example's source code. The [comments](https://github.com/bulletphysics/bullet3/blob/22d9eec148d3a7dec9d7b10a4344808bf766f73a/examples/ExampleBrowser/OpenGLExampleBrowser.cpp#L855) in the ExampleBrowser say that it is still experimental.

mode89: There is also [this](https://github.com/bulletphysics/bullet3/issues/1401). And the git log says that there were only 6 commits made into Bullet3OpenCL since that time. And less than 20 commits into any of the `Bullet3` folders. Maybe we should just drop it.

mode89:
> Could you try to run the workflow from your fork?

Sorry, I haven't seen the comment yesterday. Would like me to do it now?

bullet/src/main/resources/org/bytedeco/bullet/include/clew_stubs.h
mode89: > Ok, but why do we need any of this? What's missing from the presets for OpenCL? OpenCV is another library that uses OpenCL but doesn't need to link with any external version of OpenCL. It works just fine with the presets for OpenCL at runtime though.

Yeah, I don't see any reason why it shouldn't work with the standard OpenCL. As I understand, `clew` is just an API loader for OpenCL - like `glew` for OpenGL. I expect the API would be identical to the standard one. As I said, if we want to switch to the standard OpenCL, we are gonna need to patch some files.

mode89: I will have a look at the 32-bit Windows build later.

saudet: I don't mean to disable Bullet3OpenCL or to somehow make it link with another version of OpenCL, I just mean to remove the mappings for clew. We should be able to pass objects around with any other version of OpenCL, so it should be fine. I've pushed that in the latest commit. The only issue is that we get opaque Pointer for OpenCL objects, but that's not a big issue. Let me know if that looks alright.

saudet:
> > Could you try to run the workflow from your fork?
> 
> Sorry, I haven't seen the comment yesterday. Would like me to do it now?

No, it's fine, it works for my commits.

bullet/src/main/resources/org/bytedeco/bullet/include/clew_stubs.h
mode89: Looks good to me :) Thank you for helping out with this!

PR #1154, opened 2022-02-24
Update Tritonserver/samples instructions and dependencies

jbkyang-nvi:
Change samples running documentation and also use release CUDA and TensorRT dependencies instead of asking users to build cuda/platform and tensorrt/platform


tritonserver/README.md
```diff
@@ -69,9 +69,27 @@ Now, this `models` directory will be our model repository.
  $ export PATH=/opt/tritonserver/apache-maven-3.8.4/bin:$PATH
  $ git clone https://github.com/bytedeco/javacpp-presets.git
  $ cd javacpp-presets
+ ```
+3. There are two examples in the samples folder, `Simple.java` and `SimpleCPUOnly.java`. 
+`Simple.java` requires additional `CUDA` and `TensorRT` bindings (as well as `CUDA` and `TensorRT` libraries themselves)
```
saudet, line 79: This isn't entirely correct. We can add dependencies to the -redist artifacts, allowing us to use everything from CUDA, but without having to install on the system anything at all related to CUDA (other than the video driver). This is why I was asking about whether it makes sense to use Triton outside the container, without anything installed on the system. Does it? /cc @jackyh

saudet: Actually, that sounds alright. It doesn't make any assumptions about where the libraries come from.

PR #1157, opened 2022-03-03
Fix Memory Leak issue for Tritonserver Sample

jackyh:
This is fixed according to:
http://bytedeco.org/news/2018/07/17/bytedeco-as-distribution/

cause the following code can not be GCed without a pointerscope:

        // We make a copy of the data here... which we could avoid for
        // performance reasons but ok for this simple example.
        BytePointer odata = new BytePointer(byte_size.get());
        output_data.put(name, odata);
        System.out.println(name + " is stored in system memory");
        odata.put(base.limit(byte_size.get()));


saudet:
How about something like this instead?
```patch
diff --git a/tritonserver/samples/SimpleCPUOnly.java b/tritonserver/samples/SimpleCPUOnly.java
index ec95546403..f9594aaad2 100644
--- a/tritonserver/samples/SimpleCPUOnly.java
+++ b/tritonserver/samples/SimpleCPUOnly.java
@@ -385,6 +385,14 @@ public class SimpleCPUOnly {
         Usage("-r must be used to specify model repository path");
       }
 
+      try (PointerScope scope = new PointerScope()) {
+        Run(model_repository_path, verbose_level);
+      }
+    }
+
+    public static void
+    Run(String model_repository_path, int verbose_level) throws Exception
+    {
       // Check API version.
       int[] api_version_major = {0}, api_version_minor = {0};
       FAIL_IF_ERR(
```

tritonserver/samples/pom.xml
```diff
@@ -26,6 +26,12 @@
 	    <classifier>shaded</classifier>
         </dependency>
     </dependencies>
+    <repositories>
+        <repository>
+            <id>sonatype-nexus-snapshots</id>
+            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+        </repository>
+    </repositories>
```
saudet, line 34: @jbkyang-nvi removed that previously because I think she doesn't want to have users rely on external snapshots, so let's not add it again?

saudet:
A single PointerScope should be able to capture all allocations happening on the same thread, and deallocate everything on return. Adding more intermediary ones for Check() can help to free memory earlier, but I don't see how it could prevent memory leaks. Are you sure a single overall PointerScope doesn't work?

jackyh:
> A single PointerScope should be able to capture all allocations happening on the same thread, and deallocate everything on return. Adding more intermediary ones for Check() can help to free memory earlier, but I don't see how it could prevent memory leaks. Are you sure a single overall PointerScope doesn't work?

yes.

tritonserver/samples/pom.xml
jackyh: > @jbkyang-nvi removed that previously because I think she doesn't want to have users rely on external snapshots, so let's not add it again?

yes

jackyh:
Samuel:

We have a new version of single-threaded RunInference memory growth test as attached. I think there's still memory leak. Throuth JMC, I think it's probably because of "ResponseAlloc.call". Could you have some check?

thanks.
@saudet 
[SimpleCPUOnlyMem.zip](https://github.com/bytedeco/javacpp-presets/files/8315448/SimpleCPUOnlyMem.zip)

PR #1178, opened 2022-05-17
Make Triton Example CPU-Only

dyastremsky:
Removing Triton GPU example to remove reliance on other bindings. Making the CPU example the primary, only example.


dyastremsky:
@jbkyang-nvi for awareness.

dyastremsky:
@saudet Heads up that the CI check failure seems to be un unrelated to this PR (looks like the CUDA repo GPG signature error). We're removing the GPU example to remove any dependency on other bindings for stability.

dyastremsky:
@saudet Thanks for the suggestion! Our goal is to provide the Triton bindings, without reliance on other bindings. Including a GPU example is ideal, but we've had issues with the stability during repo updates to the included CUDA/TensorRT bindings not building or being found. We want to release official support, and we need the build to be stable to do so.

saudet:
Right, but that doesn't happen with releases, correct? So there shouldn't
be any problems. Can you elaborate on the errors you're getting?

jbkyang-nvi:
@saudet the issues are related to when Triton releases, the CUDA container we use might
1. Not be released yet to the public or 
2. Not updated yet on JavaCPP-presets as a release. 

This is common for when we are doing pre-release for Tritonserver (monthly cadence). In this case, Javacpp-presets/CUDA will need to be built in our scripts and we don't want to be maintaining that alongside for our Javacpp-presets/Tritonserver bindings and tests. 

I'm not sure I like `samples/gpu` because it is hinting that this is the GPU enabled version of the samples which is not accurate. The Tritonserver library is capable of handling models that require the GPU using CUDA just the sampled data can be prepared with or without GPU.
We could potentially keep a `samples/unofficial` or similar folder but I think we would need very clear wording that  might not always work with existing Triton containers (which is what Triton asks the users to use). 

Does this reasoning make sense Samuel?

Sorry, I don't have the error log anymore...

saudet:
I'm not sure I understand your concerns. Right now, you're satisfied with installing the presets for Triton in a container this way:
```bash
 $ mvn clean install --projects .,tritonserver
 $ mvn clean install -f platform --projects ../tritonserver/platform -Djavacpp.platform=linux-x86_64
```

We can extend that a bit by running the following to install also the presets for CUDA and TensorRT:
```bash
 $ mvn clean install --projects .,cuda,tensorrt,tritonserver
 $ mvn clean install -f platform --projects ../cuda/platform,../tensorrt/platform,../tritonserver/platform -Djavacpp.platform=linux-x86_64
```
That should be working pretty well with a large enough range of versions of CUDA and TensorRT. When you get errors though, just let me know the exact versions of CUDA and TensorRT and I'll fix them when it happens. I don't need to have access to unrelased versions of your containers to do that.

jbkyang-nvi:
@saudet the problem is pre-release, we will be using _private_ versions of CUDA that we cannot share with you. In the case that TRT and CUDA bindings have problems, we won't be able to ask you for help in resolving them

jbkyang-nvi:
@saudet the problem is that we, being part of Nvidia have pre-release versions of CUDA and TensorRT.

saudet:
I see, so a `samples/unsupported` subdirectory seems like a reasonable compromise for now. Does that sound good?

jbkyang-nvi:
Sounds good. Thanks for understanding! @dyastremsky lets put the examples that need extra bindings into a separate folder and test with the CPU version. 

@saudet happy to chat more (I see the confused face)

dyastremsky:
Thank you both. Based on the conversation above, I've moved the GPU example and its pom into a new unsupported folder. I was not able to test it due to a CUDA build error, but it looks correct.

> In file included from /opt/tritonserver/qa/L0_java_simple_example/javacpp-presets/cuda/target/native/org/bytedeco/cuda/linux-x86_64/jnicublas.cpp:118:
> /usr/local/cuda-11.6/targets/x86_64-linux/include/cudaGL.h:57:10: fatal error: GL/gl.h: No such file or directory
>    57 | #include <GL/gl.h>

saudet:
That's odd, `/usr/include/GL/gl.h` used to be available in older versions of the container. We can easily work around that by removing `"<cudaGL.h>", "<cuda_gl_interop.h>"` from `cuda/src/main/java/org/bytedeco/cuda/presets/cudart.java`, but I can come up with something cleaner that does that automatically as part of the Maven build if it becomes desirable. Just let me know when that happens!

saudet:
@dyastremsky Could you also try to update to [Release 2.21.0 corresponding to NGC container 22.04](https://github.com/triton-inference-server/server/releases/tag/v2.21.0)? Maybe that's going to fix that build error that happens on `apt-get update`:
```
W: GPG error: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64  InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A4B469963BF863CC
E: The repository 'https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64  InRelease' is not signed.
```

dyastremsky:
> @dyastremsky Could you also try to update to [Release 2.21.0 corresponding to NGC container 22.04](https://github.com/triton-inference-server/server/releases/tag/v2.21.0)? Maybe that's going to fix that build error that happens on `apt-get update`:
> 
> ```
> W: GPG error: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64  InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A4B469963BF863CC
> E: The repository 'https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64  InRelease' is not signed.
> ```

Is there something specific I need to do for the above? (E.g. upload the files modified by mvn clean install).

saudet:
What do you mean? It's just updating the versions for Triton here and there in the repository, like I did last in commit b34e0cb942fa7fcce0e104b2bfb95bbff696eb89.

saudet:
There might be changes in `src/gen`, so you can push those as well, but it's not necessary. It's only for convenience so that we can browse the source code generated more easily online since it's the API.

dyastremsky:
Thanks for clarifying, Samuel. The build worked with the changes. I've also updated the server version. I'll update the tags and rest of the references to 2.20 tomorrow morning.

Once it's passing, I think we should be good to review/merge.

saudet:
It doesn't look like `apt` works out of the box in the new version of the container either though :( I guess we can wait for its next release for the build to pass, but anyway, we can merge this PR for now, that's fine.

dyastremsky:
Interesting. Thanks, Samuel. I made the changes. Please feel free to merge, assuming you're happy with them. I included the src/gen files to make browsing the source code easier.

saudet:
Thanks! One last thing, I've pushed a commit to update the version of GSON. It seems to work fine for me, but please confirm!

dyastremsky:
Thanks for the updates, Samuel! I thought the build succeeded, but I must have built the CUDA bindings earlier. Without them, the build is failing on my end (with the current instructions) due to it trying to build from the pom in the unsupported subdirectory.

Unless Maven has an easy way to exclude that folder when building in samples, I’ll move the CPU-only version into a separate supported folder on Monday. Will update you then.

saudet:
Right, they're not related projects, so they should be both in
subdirectories...

dyastremsky:
Perfect, thanks for the updates! All looks good from my end. Feel free to merge if all looks good to you.

PR #1192, opened 2022-06-26
kvazaar: initial work

bradh:


saudet:
It would be nice to have this non-GPL alternative to libx265 in FFmpeg anyway. What's the reason you don't want to add this as part of the presets for FFmpeg?

bradh:
I would like to have this as standalone because I need to have fine control over the container boxes ( mainly around linking to some extra info using saio and saiz). 

I agree it would be good to have it in ffmpeg preset too. However it won't work for me unless I can use the encoder API directly for my use case. I see it as similar to the way we have nvcodec support in ffmpeg and also standalone.

Is there a way to borrow the cppbuild parts from kvazaar and make that a dependency for ffmpeg? Or would it be better to generate another output from the ffmpeg build ? For development it is quick to have something standalone,  but would like to centralise the maintenance work.

saudet:
The reason we need a separate nvcodec presets is because it depends on CUDA. There is no such dependencies for other codecs. And we can use the raw API of libraries linked with FFmpeg. There are no restrictions in that sense. That would be easier to maintain since all other codecs are built that way right now.

bradh:
We don't currently surface those codec APIs, right? So you can't call x265 from java independently of the avcodec api. This would be a slightly different concept - basically merging in the parse of kvazaar.h into the existing structure. 

The only downside is that it would be a lot of ffmpeg code for just one codec,  if that was the only part I needed.

I still need to get the windows build running and complete the example,  but will look to migrate across if that is the requirement. I can run the preset as a project dep if ffmpeg turns out to be a problem.

saudet:
Yes, it would make FFmpeg a dependency even if we only use a single codec, but if that's the only downside, I'd be tempted to just do that. Anything more I'd classify it under "optimizations" that users either do by themselves or get someone to do it for them. Let me know if there's some issue other than the large size of the binaries for FFmpeg though.

bradh:
> Yes, it would make FFmpeg a dependency even if we only use a single codec, but if that's the only downside, I'd be tempted to just do that. 

Also, because its only a coder (not a decoder), it may not meet everyone's needs for a non-GPL solution. It would probably do for what I need though, and it is a start.

saudet:
FFmpeg comes with an HEVC decoder that is under LGPL. It doesn't need an external library.

bradh:
> FFmpeg comes with an HEVC decoder that is under LGPL. It doesn't need an external library.

Thanks. I didn't realise that.

PR #1200, opened 2022-07-18
Bullet with double-precision floats, mutex locking and OpenMP

mode89:


mode89:
Hi @saudet :)

Sorry to bother you with all that fixing. I spent the whole evening trying to bring MacOS in vagrant to reproduce the builds - no luck. Where did you get the MacOS server?

saudet:
I use https://github.com/mxschmitt/action-tmate to debug workflows, including Mac, or more precisely, especially Mac. It doesn't work too well for Windows, but I can *usually* reproduce build issues locally in that case...

Anyway, this time I didn't bother launching debug builds, since I knew more or less what to do to fix it. That's why I ended up doing a few more commits than I wanted to :)

mode89:
Thank you for help :)

Issue #1163, opened 2022-03-30
Issue: missing ./configure file while updating libjpeg-turbo dependency of leptonica from v1.5.3 to >v2.0.4


vibhasg:
Hi, 

I wanted to rebuild the leptonica preset with the LIBJPEG v2.0.4 or higher. I am currently on a Windows 10 environment.

I tried to change the version here: 
https://github.com/bytedeco/javacpp-presets/blob/78540be41638b6b19bcc12982152d5beba464984/leptonica/cppbuild.sh#L13

and also here: https://github.com/bytedeco/javacpp-presets/blob/78540be41638b6b19bcc12982152d5beba464984/leptonica/cppbuild.sh#L21

However, unlike LIBJPEG v1.5.x, there is no ./configure file in LIBJPEG v2.x. Can someone please help with making the changes to cppbuild.sh script?

vibhasg:
Hi @saudet 
Apologies for the delay. I was able to confirm that the updates you made are indeed working. May I know if and when it is possible to see your changes in master?

saudet:
They already are in master.

vibhasg:
Thanks for that update @saudet ! Do you have a rough timeline on when v1.5.8 will be released?

saudet:
Well, 1.5.7 was just released, so not too soon! You need to tell me about this before making releases, not after. That's why there are snapshots.

jd3nn1s:
I'm curious when there might be a release to resolve this? Thanks.

saudet:
Probably sometime this fall

saudet:
This has been released with version 1.5.8. Enjoy!

PR #1201, opened 2022-07-18
OpenMP disable for Windows

AramMu:
Using OpenMP on Windows makes it run around 4x slower than without OpenMP, the slowness stands out from Linux and MacOS versions, consider better to disable until the issue is handled.


saudet:
I don't think this is a general problem with Windows, but only with your
specific machine. Please set the OMP_NUM_THREADS environment variable to 1
to disable it only on your machine.

AramMu:
This has also happened to all Windows machines that we tested not only mine, kindly please take a deeper look on it, it's not much reasonable to ask all customers to do additional things unless if it's necessary.

saudet:
Since we've moved away from Autotools to build Leptonica and Tesseract, see issue #1163, it should now be possible to build them with Visual Studio. Could you try to do that instead?

AramMu:
Ok later I'll try with Visual Studio

saudet:
I think I understand why it doesn't perform well on Windows. It looks like it's hard coding the num_threads(), whose values are most likely not optimal for any OS, especially not Windows, so removing them should solve this issue without disabling OpenMP: https://github.com/tesseract-ocr/tesseract/search?q=num_threads

@stweil What do you think?

stweil:
I don't expect that would improve the performance on Windows or any other OS unless when running on a host which does not support at least four threads. But of course you can try removing the hard coded values and test the performance of the resulting code.

saudet:
The problem on Windows is that thread scheduling isn't as efficient like as on Linux or Mac, so when we start like 100 threads, everything slows down to a crawl. It's recommend to limit the number of threads to the number of available processing cores, which OpenMP does by default anyway, so we don't (usually) need to hard code num_threads().

saudet:
@AramMu I've pushed changes to that effect to your branch. Could you give that a try?

AramMu:
Ok I'll try the changes

AramMu:
Hi, I compiled and tried with the change and it crashes when running, can later investigate further or wait

saudet:
Could you give me some sample code that crashes like that and that I can try to run here?

saudet:
It looks like Tesseract doesn't always work right with more than 4 threads, so I've updated it to force it to always use 4 threads. This way it doesn't crash and performance should be good enough on all platforms, including Windows which should be about as fast as Linux now. Please give it a try!

saudet:
Nope, unfortunately, it becomes slower on Windows even with just 2 threads. 🤔

PR #1213, opened 2022-08-21
ci: update to ubuntu 20.04

bradh:
The ubuntu 18.04 runner is deprecated. This checks how we go on building with 20.04.


bradh (commented):
Ouch. Good pick-up.

PR #1229, opened 2022-09-15
add ffmpeg-avstream-param.patch

nanguantong:
@saudet  please feel free to merge this patch~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Add support for HEVC and Opus to FLV format in presets for FFmpeg ([pull #1228](https://github.com/bytedeco/javacpp-presets/pull/1228))
  * Build Leptonica with OpenJPEG for JPEG 2000 support
  * Introduce `linux-arm64` and `macosx-arm64` builds to presets for libpostal ([pull #1199](https://github.com/bytedeco/javacpp-presets/pull/1199) and [pull #1205](https://github.com/bytedeco/javacpp-presets/pull/1205))
  * Map missing factory functions in `torch::` namespace using `torch_` as prefix in presets for PyTorch ([issue #1197](https://github.com/bytedeco/javacpp-presets/issues/1197))
--- a/ffmpeg/cppbuild.sh
+++ b/ffmpeg/cppbuild.sh
@@ -106,6 +106,7 @@ export PKG_CONFIG_PATH=$INSTALL_PATH/lib/pkgconfig/
 patch -Np1 -d $LAME < ../../lame.patch
 patch -Np1 -d $OPENSSL < ../../openssl-android.patch
 patch -Np1 -d ffmpeg-$FFMPEG_VERSION < ../../ffmpeg.patch
+patch -Np1 -d ffmpeg-$FFMPEG_VERSION < ../../ffmpeg-flv-support-hevc-opus.patch
 sedinplace 's/bool bEnableavx512/bool bEnableavx512 = false/g' x265-*/source/common/param.h
 sedinplace 's/detect512()/false/g' x265-*/source/common/quant.cpp
 
--- a/ffmpeg/cppbuild.sh
+++ b/ffmpeg/cppbuild.sh
@@ -107,6 +107,7 @@ patch -Np1 -d $LAME < ../../lame.patch
 patch -Np1 -d $OPENSSL < ../../openssl-android.patch
 patch -Np1 -d ffmpeg-$FFMPEG_VERSION < ../../ffmpeg.patch
 patch -Np1 -d ffmpeg-$FFMPEG_VERSION < ../../ffmpeg-flv-support-hevc-opus.patch
+patch -Np1 -d ffmpeg-$FFMPEG_VERSION < ../../ffmpeg-avstream-param.patch
 sedinplace 's/bool bEnableavx512/bool bEnableavx512 = false/g' x265-*/source/common/param.h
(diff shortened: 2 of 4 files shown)


saudet:
Can you try to contribute this upstream first and see what they say? http://ffmpeg.org/developer.html#Contributing

nanguantong:
`1、int64_t av_stream_get_cur_dts(const AVStream *st);`
    In order to set input AVPackt pts/dts better that avoid "non monotonically increasing dts to muxer in stream" while muxing stream.

`2、void av_stream_set_probe_packets(AVStream *st, int probe_packets);`
    In order to avoid taking too long time for `avformat_find_stream_info` to probe the known codec info (e.g. mpeg audio pcma).

nanguantong:
> 

The AVStream partial parameters had been moved to internal struct FFStream since 5.0 that caused previous call failed in my code.

saudet:
Oh, I see, so they're not going to add this back... Do you have a reference that explains why they moved these parameters internally?

nanguantong:
> Oh, I see, so they're not going to add this back... Do you have a reference that explains why they moved these parameters internally?

Not found detail reasons, only https://github.com/FFmpeg/FFmpeg/commit/591b88e6787c4e678237f02a50421d101abd25c2
https://github.com/FFmpeg/FFmpeg/commit/fab2ed47042d4cc2a4cd69bb97738024c01300c7

**"It's a private fields, no reason to have it exposed in a public header."**

saudet:
It sounds like there's supposed to be another way to set those values, for example, there is a max_probe_packets option:
https://github.com/FFmpeg/FFmpeg/blob/n5.1/libavformat/options_table.h#L108
That doesn't do what you need?

nanguantong:
> 

Option max_probe_packets is another one that different effect, option probe_packets can be set  to interrupt the probe known format at any time that reducing read and write latency.

nanguantong:
When the input stream bitrate is so small (e.g. 256k), the probe processing spends nearly 10 seconds finding the stream info, so long... so that need set the probe_packets = 0 actively.

This is so important issue, please feel free to merge this.

nanguantong:
`2、void av_stream_set_probe_packets(AVStream *st, int probe_packets);`
We found the replaced method to avoid the too long time to probe the stream info. So ignore this issue.

saudet:
> We found the replaced method to avoid the too long time to probe the stream info. So ignore this issue.

Ok, good! What method is that?

And to avoid "non monotonically increasing dts to muxer in stream" I'm pretty sure that setting correct pts/dts values in frames and/or packets should fix that.

nanguantong:
https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/demux.c#L448 set s->probesize = 0 or other smaller value (e.g. 1024* 4).

You'are right, but sometimes stream self pts and dts has some out of order or drop packets, the program do not immediately detect it, so need set 
```
 if (stream.cur_dts() >= outPkt.dts()) {
    outPkt.dts(stream.cur_dts() + 1);
    if (outPkt.pts() < outPkt.dts()) {
        outPkt.pts(outPkt.dts());
    }
 }
ret = av_interleaved_write_frame(ofmtCtx, outPkt);
```

saudet:
There must be some other way to get that information from a stream...

nanguantong:
> There must be some other way to get that information from a stream...

Not found yet.

saudet:
That sounds like a question they could answer upstream: https://ffmpeg.org/contact.html

nanguantong:
https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/mux.c#L543
https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/mux.c#L562

So may use `av_stream_get_end_pts` to replace the stream.cur_pts(), fixed this issue.

                        ```
long endPts = av_stream_get_end_pts(stream);
                        if (endPts >= outPkt.dts()) {
                            outPkt.dts(endPts + 1);
                            if (outPkt.pts() < outPkt.dts()) {
                                outPkt.pts(outPkt.dts());
                            }
                        }
```

saudet:
It looks like cur_pts is simply set to the last pkt->dts, so you could simply do the same in your application:
https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/mux.c#L561

nanguantong:
> It looks like cur_dts is simply set to the last pkt->dts, so you could simply do the same in your application: https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/mux.c#L561

Yes, got it.

nanguantong:
> It looks like cur_dts is simply set to the last pkt->dts, so you could simply do the same in your application: https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/mux.c#L561

Yes, got it.

nanguantong:
> It looks like cur_dts is simply set to the last pkt->dts, so you could simply do the same in your application: https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/mux.c#L561

Yes, got it.

> 

Not simply, ffmpeg changes the input pkt->pts and pkt->dts internally, so not simply save the last input pkt->dts as the cur_dts.

nanguantong:
Okay, I'll try to find other methods to replace the stream.cur_dts, this pr may be closed.

saudet:
Didn't you just agree that pkt.dts, after it gets written, contains the
same information?

nanguantong:
May use the pkt.dts as the last dts, same effect as stream.cur_dts normally, but sometimes not equal, so need to enhance the the processing logic.

PR #1235, opened 2022-09-21
Added build support for librealsense2 on Apple Silicon arm64

cansik:
This PR adds build support for Apple Silicon arm64 using the XCode build system as recommended [here](https://lightbuzz.com/realsense-macos/) and [here](https://github.com/cansik/pyrealsense2-macosx).

### DIFF
--- a/.github/workflows/librealsense2.yml
+++ b/.github/workflows/librealsense2.yml
@@ -42,7 +42,7 @@ jobs:
     steps:
       - uses: bytedeco/javacpp-presets/.github/actions/deploy-macosx@actions
   macosx-arm64:
-    runs-on: macos-12
+    runs-on: macos-11
     steps:
       - uses: bytedeco/javacpp-presets/.github/actions/deploy-macosx@actions
   windows-x86:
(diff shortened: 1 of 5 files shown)


saudet:
Looks good, but we'll also need to update the workflow file for GitHub Actions to get this tested:
https://github.com/bytedeco/javacpp-presets/blob/master/.github/workflows/librealsense2.yml
As well as the platform/pom.xml file:
https://github.com/bytedeco/javacpp-presets/blob/master/librealsense2/platform/pom.xml

saudet:
It's probably easier to use just the normal make to build that since we know that works.

cansik:
The standard make command did not work on my system, even for the python library. Others are working with XCode as well because of that.

Intel dropped MacOS suppport lately, so it could be related to that.

But let's try it on the Github Actions environment. I am going to adapt it.

saudet:
It looks like we'll need to build libusb from source as well?

saudet:
Looks like GitHub Actions now has free M1 runners. Want to give this a try again with that? https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

PR #1243, opened 2022-10-10
Add macosx-arm64 preset for lz4

gab1one:
This PR enables the build of the macosx-arm64 preset for lz4.

Build is working, please let me know if anything else is missing.

(I created the PR with the wrong account earlier, that is why I closed that one and opened this one)

### DIFF
--- a/.github/workflows/lz4.yml
+++ b/.github/workflows/lz4.yml
@@ -27,11 +27,11 @@ jobs:
     container: centos:7
     steps:
       - uses: bytedeco/javacpp-presets/.github/actions/deploy-centos@actions
-  macosx-x86_64:
+  macosx-arm64:
     runs-on: macos-11
     steps:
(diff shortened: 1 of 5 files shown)


saudet:
For consistency, everything "macosx-arm64" should appear before "macosx-x86_64", and we'll need to cross compile in cppbuild.sh, but the rest looks fine.

gab1one:
- I added a commit to fix the ordering.
- Looking into setting the cross-compile settings

saudet:
I don't think the target flag is going to work. Anyway, it's easier to use just "-arch arm64" and that's been working for all other builds, for example: https://github.com/bytedeco/javacpp-presets/blob/master/libffi/cppbuild.sh#L90

gab1one:
Thanks for the tip, that seems to be working when running clang on mac.

gab1one:
Is there anything else missing for this to get merged?

gab1one:
Thank you very much for the support and merging 👍🏾

PR #1296, opened 2022-12-26
Use latest VAAPI for building ffmpeg on linux (x86, x86_64)

### LINKED ISSUE #1289: [ffmpeg] Linux natives use an old version of VAAPI
When using VAAPI with the ffmpeg libraries, opening a VAAPI hwdevice fails because the old version of VAAPI can't call the new vendor libraries on the system.

Error messages:
```
libva info: VA-API version 0.40.0
libva info: va_getDriverName() returns 0
libva info: User requested driver 'iHD'
libva info: Trying to open /usr/lib64/dri/iHD_drv_video.so
libva error: /usr/lib64/dri/iHD_drv_video.so has no function __vaDriverInit_0_32
libva info: va_openDriver() returns -1
[AVHWDeviceContext @ 0x7f937c7662c0] Failed to initialise VAAPI connection: -1 (unknown libva error).
```

Currently used version is 0.40.0, but the most recent version is 1.16.0

silenium-dev:
This allows the ffmpeg build to use the latest version of libva. Fixes #1289.


saudet:
This doesn't work on CentOS 7 though so we can't never this.

silenium-dev:
Why is it required to build on CentOS 7? It just results in having to build every dependency from source instead of just installing the packages.

saudet:
> Why is it required to build on CentOS 7? It just results in having to build every dependency from source instead of just installing the packages to have the latest version.

Yes, if you're willing to give that a try, that's what we would need to do.

silenium-dev:
I'm going to give it a try, could someone maybe take a look on why the build is failing?

silenium-dev:
I'm now building the libva package from source by taking the old source rpm and replacing the source archive and version and then building it with [`mock`](https://github.com/rpm-software-management/mock) for either i686 or x86_64.

saudet:
It's probably going to be easier to build from source from the original repository: https://github.com/intel/libva

silenium-dev:
It is build from this original source, I'm downloading the latest release source tarball, but it happens in the context of rpmbuild, which makes it much easier to build for 32-bit on a 64-bit system.

silenium-dev:
> @silenium-dev just curious what's the actual status of this pull request? I encountered some problems in regards of the VAAPI hw decoder on my Linux test machine and I think this is related to the same issue.

The actions run successful in my repository, but the checks are failing, because the fetch of the action is broken (https://github.com/silenium-dev/javacpp-presets/actions/runs/4872749240/jobs/8691383577#step:4:5852). I don't know how to fix this, maybe you have an idea?

streamingdv:
Also no idea unfortunately, pretty new to that awesome repo here to be honest. But maybe @saudet has some inputs about how you can fix that?

saudet:
This looks like a lot of hacking that I don't think we need... Instead of trying to figure this out though, it looks like there are already RPMs available for that here: https://negativo17.org/repos/multimedia/epel-7/x86_64/
Did you try to just use those?

silenium-dev:
> This looks like a lot of hacking that I don't think we need... Instead of trying to figure this out though, it looks like there are already RPMs available for that here: https://negativo17.org/repos/multimedia/epel-7/x86_64/ Did you try to just use those?

I didn't, but it seems that there are only 64 bit builds, but we need 32 bit ones, too.

saudet:
32-bit builds seem available to me. Did you give it a try and found that they don't work? Just saying that they're not available without actually looking at the files isn't helpful information.
```
[   ] [libva-2.18.0-1.el7.i686.rpm](https://negativo17.org/repos/multimedia/epel-7/x86_64/libva-2.18.0-1.el7.i686.rpm)                                2023-04-13 20:54   94K  
[   ] [libva-2.18.0-1.el7.x86_64.rpm](https://negativo17.org/repos/multimedia/epel-7/x86_64/libva-2.18.0-1.el7.x86_64.rpm)                              2023-04-13 20:54   95K  
[   ] [libva-devel-2.18.0-1.el7.i686.rpm](https://negativo17.org/repos/multimedia/epel-7/x86_64/libva-devel-2.18.0-1.el7.i686.rpm)                          2023-04-13 20:54  683K  
[   ] [libva-devel-2.18.0-1.el7.x86_64.rpm](https://negativo17.org/repos/multimedia/epel-7/x86_64/libva-devel-2.18.0-1.el7.x86_64.rpm)                        2023-04-13 20:54  683K  
[   ] [libva-utils-2.18.2-1.el7.x86_64.rpm](https://negativo17.org/repos/multimedia/epel-7/x86_64/libva-utils-2.18.2-1.el7.x86_64.rpm)                        2023-04-20 12:18  537K
```

saudet:
In any case, we won't be able to merge this until **all** the builds pass, and they are currently **all** failing. Either get this working the simple way, or please figure out how to make **all*** your hacks actually work.

silenium-dev:
The problem isn't the libva build, this part is working correctly, but the pipeline doesn't work on pull requests, for some reason it tries to fetch the code again in the pipeline instead of using the already checked out one by GitHub actions.

silenium-dev:
Ok, I think I solved the issue that was preventing the pipeline from succeeding.
The workflow always used the actions from **this** repository, so there didn't have to be `actions/checkout` before. Now, if I want to modify the pipeline and change the action path to be in the repository the pipeline runs in and not a hard-coded path, it requires a `actions/checkout` before.
This resulted in the workspace directory not being empty and `git fetch` + `git checkout` reporting an error. I've replaced the `checkout` with `reset --hard` and now it works.
I'm curious if this fixed the issue with the failing pull request checks as well, or if this is another issue.

P.S: I've tried the epel-multimedia repo as suggested, it seems to work, but the pipeline is still running, so we'll see.

.github/actions/deploy-centos/action.yml
```diff
@@ -26,7 +27,9 @@ runs:
         if [[ "$CENTOS_VERSION" == "6" ]]; then
           sed -i 's/mirror\.centos\.org\/centos/vault.centos.org/g;s/6\/sclo/6.10\/sclo/g;s/mirrorlist/#mirrorlist/g;s/#baseurl/baseurl/g' /etc/yum.repos.d/CentOS-SCLo-scl-rh.repo
         fi
-        yum -y install $SCL_ENABLE rh-java-common-ant boost-devel ccache clang gcc-c++ gcc-gfortran java-1.8.0-openjdk-devel ant python python3-devel python3-pip swig file which wget unzip tar bzip2 gzip xz patch autoconf-archive automake make libtool bison flex perl-core nasm alsa-lib-devel freeglut-devel gtk2-devel libusb-devel libusb1-devel curl-devel expat-devel gettext-devel openssl-devel bzip2-devel zlib-devel SDL2-devel libva-devel libxkbcommon-devel libxkbcommon-x11-devel xcb-util* fontconfig-devel libffi-devel ragel ocl-icd-devel GeoIP-devel pcre-devel ssdeep-devel yajl-devel
+        curl -L "https://negativo17.org/repos/epel-multimedia.repo" -o /etc/yum.repos.d/epel-multimedia.repo
```
saudet, line 30: Does that make YUM pick up libraries from that repo in priority? Is there a way to not make it do that? If not, since we're just interested in libva here, what about downloading the RPMs manually and installing only those?

saudet: Then again, it doesn't look like there's any other package from there than libva that overlaps with EPEL anyway? What do you think?

silenium-dev: We should lower the priority, just to be safe, because we don't know if the owner adds any packages in the future.

silenium-dev: I'm now setting the priority to 99 (lowest) via sed

saudet: I've reverted that last one since it's failing with linux-x86. It's fine like that.

PR #1177, opened 2022-05-13
add av1 codec to ffmpeg

n-kai-cj:
add `libaom-av1` and `libsvtav1` codec to ffmpeg.
I just checked working on windows-x86_64.

PR #1303, opened 2023-01-06
ffmpeg: AV1 integration

bradh:
Obsoletes #1177


bradh:
@n-kai-cj Can you check this works for your use case?

n-kai-cj:
@bradh Thank you! This works.

ffmpeg/cppbuild.sh
```diff
@@ -43,6 +45,10 @@ XML2=libxml2-2.9.12
 LIBSRT_VERSION=1.5.0
 WEBP_VERSION=1.2.4
 FFMPEG_VERSION=5.1.2
+LIBSRT_VERSION=1.4.4
+WEBP_VERSION=1.2.2
```
saudet, line 49: Do we need to use old versions of libsrt and webp to make this work?

bradh: I think just bad merge on my part. Will try cleaning that up.

bradh: Looks like its still OK.

PR #1307, opened 2023-01-08
Bundle libcupti with libtorch to make it work with cuda-platform-redist

sbrunk:
I've managed to bundle libcupti by adding it to the link list. Is this the right approach in general?

One issue is that it adds `libcupti.so` without the version information, but `libcuda` links with `libcupti.so.11.8`.

### DIFF
--- a/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
+++ b/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
@@ -1735,17 +1735,18 @@ import org.bytedeco.openblas.presets.openblas;
         ),
         @Platform(
             value = {"linux", "macosx", "windows"},
-            linkpath = {
-                    "/usr/local/cuda-11.8/lib64/",
-                    "/usr/local/cuda-11.8/extras/CUPTI/lib64/",
-                    "/usr/local/cuda/lib64/",
-                    "/usr/local/cuda/extras/CUPTI/lib64/",
-                    "/usr/lib64/",
-                    "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/lib/x64/",
-                    "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/extras/CUPTI/lib64/"
+            link = {"c10", "c10_cuda", "torch_cpu", "torch_cuda", "torch"},
+            preload = {"gomp@.1", "iomp5", "omp", "tbb@.2", "asmjit", "fbgemm", "cupti@.11.8"},
+            preloadpath = {
+                "/usr/local/cuda-11.8/lib64/",
+                "/usr/local/cuda-11.8/extras/CUPTI/lib64/",
+                "/usr/local/cuda/lib64/",
+                "/usr/local/cuda/extras/CUPTI/lib64/",
+                "/usr/lib64/",
+                "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/lib/x64/",
+                "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/extras/CUPTI/lib64/",
+                "C:/Program Files/NVIDIA Corporation/NvToolsExt/bin/x64/",
             },
-            link = {"c10", "c10_cuda", "torch_cpu", "torch_cuda", "torch", "cupti@.11.8"},
-            preloadpath = "C:/Program Files/NVIDIA Corporation/NvToolsExt/bin/x64/",
             extension = "-gpu"
         ),
     },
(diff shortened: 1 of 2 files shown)


saudet:
> I've managed to bundle libcupti by adding it to the link list. Is this the right approach in general?

I think that's OK for this case, yes.

> One issue is that it adds `libcupti.so` without the version information, but `libcuda` links with `libcupti.so.11.8`.

There's a dot missing, it needs to be `cupti@.11.8`.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
```diff
@@ -1735,7 +1735,8 @@
         ),
         @Platform(
             value = {"linux", "macosx", "windows"},
-            link = {"c10", "c10_cuda", "torch_cpu", "torch_cuda", "torch"},
+            linkpath = {"/usr/local/cuda/lib64/", "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/lib/x64/"},
```
saudet, line 1738: For consistency, we should add the same paths as the presets for CUDA here: https://github.com/bytedeco/javacpp-presets/blob/master/cuda/src/main/java/org/bytedeco/cuda/presets/cudart.java#L46

saudet: With a default installation on Linux though, it looks like that particular library is found in "/usr/local/cuda-11.8/extras/CUPTI/lib64/" on Linux and in "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8//extras/CUPTI/lib64/" on Windows, no?

saudet:
> But since the Linux CI-build did not fail, it should have been able to find the library on centos as well right?
So I've just changed the Windows path for now.

Seems to be located there on the CI server as well, yes, but let's add that extras path on Linux as well since it sometimes gets placed there by default, for example, see https://github.com/tensorflow/tensorflow/issues/39526.

sbrunk:
> Seems to be located there on the CI server as well, yes, but let's add that extras path for Linux as well since it sometimes gets placed there by default, for example, see [tensorflow/tensorflow#39526](https://github.com/tensorflow/tensorflow/issues/39526).

Done.

saudet:
Actually, we should try to get this done with "preload" since we know that works well for the core libraries from CUDA. I've pushed a commit in that regards to your branch. Could you test this out to make sure it works?

sbrunk:
Yes works for me (tested only on Linux so far after building locally).

Do you know if there's a way to download the CI artifacts from GH actions? I've seen repos where they are cached and are available for download for a few days but perhaps there are size restrictions.

saudet:
> Yes works for me (tested only on Linux so far after building locally).

It looks like the profiler is disabled on Windows by default anyway, so no need to worry about that for now I guess:
https://github.com/pytorch/pytorch/blob/master/cmake/Dependencies.cmake#L1898
 
> Do you know if there's a way to download the CI artifacts from GH actions? I've seen repos where they are cached and are available for download for a few days but perhaps there are size restrictions.

Yes, all builds get deployed as snapshots: http://bytedeco.org/builds/

sbrunk:
But only builds on master right? At least the [sonatype snapshot](https://oss.sonatype.org/#nexus-search;gav~org.bytedeco~pytorch~~~) seems to correspond to the latest pytorch related commit on master (before this PR) on Jan 3.

What I meant was the ability to download artifacts from PRs for testing things. I think what I've seen was a result of the upload-artifact action: https://github.com/actions/upload-artifact#where-does-the-upload-go

I could try to add that if you think it could make sense.

saudet:
> But only builds on master right? At least the [sonatype snapshot](https://oss.sonatype.org/#nexus-search;gav~org.bytedeco~pytorch~~~) seems to correspond to the latest pytorch related commit on master (before this PR) on Jan 3.

Not just the master branch, any branch. You'll need to set "CI_DEPLOY_*" secrets in your fork to values corresponding to your account on Sonatype OSSRH, and I can give you access to the "org.bytedeco" groupId to have your artifacts pushed there as well, with different versions probably to avoid conflicts, but still it works.

> What I meant was the ability to download artifacts from PRs for testing things. I think what I've seen was a result of the upload-artifact action: https://github.com/actions/upload-artifact#where-does-the-upload-go
> 
> I could try to add that if you think it could make sense.

Possibly, but it's usually easier to just use a Maven repository than a manual system like that. If we could wire this to GitHub Packages somehow, then being able to use our GitHub Account would be a clear advantage over Sonatype OSSRH: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry

Issue #1315, opened 2023-01-28
[libffi] remove constant global.ffi.FFI_DEFAUL_ABI and friends and point to platform-specific preset.ffi.FFI_DEFAULT_ABI


drcicero:
Hi, I am using `"org.bytedeco" % "libffi-platform" % "3.4.4-1.5.8"` (via Scala) on Linux.

I think the enum values:
https://github.com/bytedeco/javacpp-presets/blob/86a78a6bca5cadc01d33933ce7a77fa6a29f1219/libffi/src/gen/java/org/bytedeco/libffi/global/ffi.java#L85-L103 should be removed, as there platform specific, e.g., the comments still show that in C they were platform-specifically defined via `#ifdef`.

It seems the correct way to go is to use the already defined `presets.ffi.FFI_DEFAULT_ABI()` methods and friends, which are defined as native and supplied from somewhere else:
https://github.com/bytedeco/javacpp-presets/blob/86a78a6bca5cadc01d33933ce7a77fa6a29f1219/libffi/src/main/java/org/bytedeco/libffi/presets/ffi.java#L85-L103
I'm not sure where the native things get their value from, but those have the right value for me.

(I noticed this, because in the current state they are easy to confuse, e.g, when your call the method `FFI_DEFAULT_ABI()` you get the right value, but when you access the field `FFI_DEFAULT_ABI` you get the wrong value, and an error later when you try to pass it to prepare. They are even easier to confuse in Scala, where syntactic sugar often allows one to shorten `x.f()` to `x.f`.)

saudet:
We can probably get that done with a couple of Info.skip.
Please open a pull request to that effect, contributions are welcome!

drcicero:
Instead of skipping lines by regex i would prefer to skip these by variable name; however it seems that's only possible with fields/methods; not enum fields, is it?

saudet:
I don't think it's possible to skip individual enumerators at the moment, but it should be possible to skip whole enumerations. If you've tried something specific that doesn't work, please show me what you've tried exactly.

saudet:
Fix released with JavaCPP 1.5.9! Thanks for the contribution

PR #1318, opened 2023-01-30
libffi: hide platform-ignorant FFI ABI enum values

drcicero:
This change ignores processing the whole `ffitarget.h` file, which If i understood correctly contains platform-specific values, but only for one platform, and therefore is on most platforms wrong. I believe the platform-specific values can already be accessed `native` methods handwritten in the preset. I added the otherwise missing `FFI_TRAMPOLINE_SIZE()` method.

I think only two other variables remain in the `ffitarget.h` file, which are `FFI_CLOSURES = 1; FFI_GO_CLOSURES = 1;`. I'm not super sure, but i believe those are only there to enable/disable definitions in the other header, it shouldn't hurt not to have them.

I tried `mvn clean test --projects .,libffi` and it succeeds. I have not figured out if there are more tests to run?

See https://github.com/bytedeco/javacpp-presets/issues/1315


drcicero:
Oh, the example https://github.com/bytedeco/javacpp/wiki/Mapping-Recipes#skipping-lines-from-header-files sounded like one can only skip lines of the form
```
// START COMPLEX DECLARATIONS
// ...
// END COMPLEX DECLARATIONS
```
but by studying the other InfoMap's I found it is actually possible to individual skip definition as well.

I suppose a simpler change would be to add the following:
```
.put(new Info("FFI_FIRST_ABI","FFI_SYSV","FFI_THISCALL","FFI_FASTCALL","FFI_STDCALL","FFI_PASCAL","FFI_REGISTER","FFI_MS_CDECL","FFI_LAST_ABI","FFI_DEFAULT_ABI").skip())
```

drcicero:
Now, the change is only the small addition of a few skips (`.put(new Info("FFI_FIRST_ABI", "FFI_SYSV", "FFI_THISCALL", "FFI_FASTCALL", "FFI_STDCALL", "FFI_PASCAL", "FFI_REGISTER", "FFI_MS_CDECL", "FFI_LAST_ABI", "FFI_DEFAULT_ABI").skip())`), because those ones already are defined in the preset. And as this leads to having the correct `FFI_DEFAULT_ABI` definition autogenerated, I removed the manual definition of `FFI_DEFAULT_ABI` from the preset.

libffi/src/main/java/org/bytedeco/libffi/presets/ffi.java
```diff
@@ -79,6 +79,9 @@ public void map(InfoMap infoMap) {
                .put(new Info("_ffi_type").pointerTypes("ffi_type"))
                .put(new Info("void (*)(void)").cast().pointerTypes("Pointer"))
                .put(new Info("__attribute__((deprecated))").annotations("@Deprecated"))
+               .put(new Info("FFI_FIRST_ABI", "FFI_SYSV", "FFI_THISCALL", "FFI_FASTCALL",
+                             "FFI_STDCALL", "FFI_PASCAL", "FFI_REGISTER", "FFI_MS_CDECL",
+                             "FFI_LAST_ABI", "FFI_DEFAULT_ABI").skip())
```
drcicero, line 84: Actually, I did tell it to skip `FFI_DEFAULT_ABI`. But that didn't work, and I currently don't see why...

saudet: Ah, ok, probably just a bug in JavaCPP, but it doesn't cause any problems, so whatever. :)

PR #1322, opened 2023-02-10
Add mamba 2023.02.09

mkitti:
Greetings from the ["Global BioImaging hackathon to advance deep learning image analysis"](https://globalbioimaging.org/news/gbi-hackathon-italy-2023).

Here we have built bindings for `libmamba`. This allows us to replicate the functionality of micromamba to manage conda environments from Java.


saudet:
Thanks! But what's the advantage of using the C API over just using the `micromamba` program?

mkitti:
Using the micromamba static executable is definitely an option, but this does not provide an API in and of itself. It would be nice to be able to usa mamba programmatically. Thus, I could either build a new API or just use the existing one.

Also, note that Python has an API to mamba:
https://mamba.readthedocs.io/en/latest/python_api.html

I plan on further exposing the C++ API in the next week after I address the the CI issues.

The application here is that we need to setup machine learning models provided in a model zoo. These predefined models may require specific versions of software to operate.

saudet:
We can use the micromamba program from Java programmatically too. What can we do with the C++ API that we can't do with the command line interface?

mkitti:
Sure, but if we used the micromamba CLI, then we need to format the command line arguments and parse the output manually.

Why would I build an API around the CLI when a C interface already exists? Why would I scrape data structures from stdout, when I could just use the native data structures of the library?

mkitti:
Specifically, from the micromamba CLI tool, we cannot access the internal API in libmamba as shown here in via Python:  https://mamba.readthedocs.io/en/latest/python_api.html#using-internal-mamba-apis

mkitti:
To move this forward, I need some of the dependencies in place, which actually may be more interesting than libmamba itself.

I'm going to start working on https://github.com/openSUSE/libsolv . This could be really useful to SciJava / FIJI 's updater.
https://imagej.net/plugins/updater

mkitti:
conda-forge recipe: https://github.com/conda-forge/libsolv-feedstock/blob/main/recipe/build.sh

```
{% set name = "libsolv" %}
{% set version = "0.7.23" %}
{% set sha256 = "0286155964373c6cc3802d025750786c3ee79608d5cb884598e110e3918bb2fe" %}

package:
  name: {{ name|lower }}
  version: {{ version }}

source:
  fn: {{ name }}-{{ version }}.tar.gz
  url: https://github.com/openSUSE/libsolv/archive/{{ version }}.tar.gz
```

Issue #1185, opened 2022-06-03
OpenCV Fisheye Camera Functions Missing


calvertdw:
The "undistort" function is missing:

https://docs.opencv.org/3.4/da/d54/group__imgproc__transform.html#ga69f2545a8b62a6b0fc2ee060dc30559d

and the fisheye functions:
https://docs.opencv.org/4.5.5/db/d58/group__calib3d__fisheye.html

Edit: I renamed this issue to be more general about needing functions relating to fisheye cameras. We are trying to accelerate our new Fisheye camera on our robot with OpenCV :slightly_smiling_face:

saudet:
I see all of them:
http://bytedeco.org/javacpp-presets/opencv/apidocs/org/bytedeco/opencv/global/opencv_calib3d.html#undistort-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-
http://bytedeco.org/javacpp-presets/opencv/apidocs/org/bytedeco/opencv/global/opencv_calib3d.html#undistortImage-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Size-
http://bytedeco.org/javacpp-presets/opencv/apidocs/org/bytedeco/opencv/global/opencv_calib3d.html#undistortPoints-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-org.bytedeco.opencv.opencv_core.Mat-

calvertdw:
Oh well that's great news! Thank you. I was expecting them to be in imgproc.

calvertdw:
[...]

PR #1324, opened 2023-02-15
Address OpenCV fisheye symbol collision

ds58:
This PR should address #1185 

OpenCV has `cv:::initUndistortRectifyMap` and `cv::fisheye::initUndistortRectifyMap` in the calib3d module. The Java for the fisheye version wasn't getting generated because of the symbol collision. I added a `javaNames` for the fisheye version so it maps to `fisheyeInitUndistortRectifyMap`

### DIFF
--- a/opencv/src/gen/java/org/bytedeco/opencv/global/opencv_calib3d.java
+++ b/opencv/src/gen/java/org/bytedeco/opencv/global/opencv_calib3d.java
@@ -4988,26 +4988,32 @@ CV_64FC2) (or vector\<Point2f\> ).
     image points coordinates (as functions of all the input parameters) with respect to the particular
     parameters, intrinsic and/or extrinsic.
      */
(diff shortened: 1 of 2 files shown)


saudet:
Looks good, thanks! While we're it, let's also prefix with "fisheye" all functions from the`cv::fisheye` namespace?
If I'm not overlooking anything, the complete list is:
cv::fisheye::projectPoints
cv::fisheye::distortPoints
cv::fisheye::undistortPoints
cv::fisheye::initUndistortRectifyMap
cv::fisheye::undistortImage
cv::fisheye::estimateNewCameraMatrixForUndistortRectify
cv::fisheye::calibrate
cv::fisheye::stereoRectify
cv::fisheye::stereoCalibrate

ds58:
> While we're it, let's also prefix with "fisheye" all functions from the`cv::fisheye` namespace?

Yes, I think that's a good idea. I've done that.

PR #1332, opened 2023-03-13
[HDF5] Use cmake on all platforms

mkitti:


mkitti:
Closing this. This is not going to work until 1.14.1. There are too many issues with cmake that are now fixed upstream, but not released yet.

PR #1334, opened 2023-03-14
Enable native builds of szip via libaec

mkitti:
szip is a compressor that can be built into HDF5. A free implementation of szip is available via libaec.


mkitti:
After sorting out Windows, I may know how to convert everything to cmake now. That might be another pull request.

saudet:
Why do we need to build libaec just for szip? Let's remove that dependency if we're not using it.

mkitti:
How else do you propose building szip? Basically libaec provides a compatability layer that provides a szip interface.

szip itself is otherwise provided via a non-free license.

saudet:
Ah, yes, I remember. So it doesn't matter if we use it via libaec, it's still not free and it's just libaec not being compliant isn't?

mkitti:
According to NASA, there are no known patents on this standard compression either.

https://public.ccsds.org/Pubs/121x0b3.pdf

> At time of publication, the specifications of this Recommended Standard are not known to be
the subject of patent rights. There is currently no known active patent for this standard.2

hdf5/cppbuild.sh
```diff
@@ -9,21 +9,45 @@ fi
...
+    cd ../../hdf5-$HDF5_VERSION
+}
+build_aec_szip_windows() {
+    mkdir -p ../libaec-$AEC_VERSION/build
+    cd ../libaec-$AEC_VERSION/build
+    "$CMAKE" $@ -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$INSTALL_PATH ..
+    ninja -j $MAKEJ
+    ninja install
+    cd ../../hdf5-$HDF5_VERSION
+}
```
saudet, line 48: Let's not create functions like this. It gets ugly real quick. Place anything in common for all platforms in variables only, and anything else in the switch below.

mkitti: I removed the functions in [aaf057d](https://github.com/bytedeco/javacpp-presets/pull/1334/commits/aaf057dccf347ddd85082e45c21b7d80ce9d7774). I increased use of `pushd` / `popd` to help keep the directories straight.

mkitti:
Thank you for the edits and the merge!

Issue #1051, opened 2021-05-27
How to deal with template taking std::function as arguments?


egolearner:
```cpp
    template <typename T>
    void set_filter(T &&func) {
       // ...
    }
```
T's receiver is `std::function<bool(uint64_t)>`.

We tried to add the following mapping info
```java
        infoMap.put(new Info("std::function<bool(uint64_t)>").pointerTypes("FilterFunction"));
        infoMap.put(new Info("SomeClass::set_filter<std::function<bool(uint64_t)> >&&").javaNames("setFiler"));

```
`mvn install` works but the function is omitted from generated java file.

egolearner:
```java
        infoMap.put(new Info("std::function<bool(uint64_t)>").pointerTypes("FilterFunction"));

    public static class FilterFunction extends FunctionPointer {
        static { Loader.load(); }
        /** Pointer cast constructor. Invokes {@link Pointer#Pointer(Pointer)}. */
        public    FilterFunction(Pointer p) { super(p); }
        protected FilterFunction() { allocate(); }
        private native void allocate();
        public native boolean call(@Cast("uint64_t") long key);
    }
```
Hi @saudet, I added `FilterFunction` definition according to your comments. It worked if I add another non template function.
```cpp
  void set_filter2 (std::function<bool(uint64_t)> &&func) {
    filter_ = std::move(func);
  }
```
However, the original function is a template function. I still cannot get it work. I have tried adding the following mapping info.
```java
//        infoMap.put(new Info("SomeClass::set_filter<std::function<bool(uint64_t)> >&&").javaNames("set_filter"));


//        infoMap.put(new Info("SomeClass::set_filter<std::function<bool(uint64_t)> >&&")
//                .javaText("public native void set_filter(@ByVal FilterFunction func);"));

```

saudet:
Try without the `&&`, it's not being processed as part of type signature, although it probably should...

egolearner:
Remove `&&` works. Thanks a lot @saudet 

```java
        infoMap.put(new Info("SomeClass::set_filter<std::function<bool(uint64_t)> >").javaNames("set_filter"));
```

saudet:
BTW, I've added support for `std::function` as a "basic container" in commit https://github.com/bytedeco/javacpp/commit/587ed7a4e704927eb9238f0fae80ff9a5d546ed7, so this should work now. Please give it a try with the snapshots: http://bytedeco.org/builds/

egolearner:
[...]

PR #1360, opened 2023-05-22
[Pytorch] New version of the presets

HGuillemet:
Reorganization of the Pytorch presets.

Main improvements are
* Handling of the virtual inheritance between `torch::nn::Module` and its subclasses. Avoiding segmentation faults when a function accepting `torch::nn::Module` or a `Module` instance method is called from a subclass instance.
* Transparent and more rigorous handling of `shared_ptr`. This concerns `Module` and its subclasses, `Tensor` and a few less important classes. A consequence is that modules are now deallocated normally (they were never deallocated in previous version of the presets).
* Module holders that are the C++ counterpart of the transparent handling of `shared_ptr` are not mapped any more.
* Friend functions that were ignored are now mapped, giving access to additional operator overloads, like on iterators.
* `Module` is now virtualized, meaning that if you override in Java its virtual member functions (`train`, `is_training`, `to`, `zero_grad`, `save`, `load`, `pretty_print`, `is_serializable`), your Java code will get called by libtorch.
* Added support for `Sequential`, `AnyModule`, `AnyValue`.
* Hopefully a better coverage of the API. Almost all what is available in C++ with:
```c++
#include <torch/torch.h>
#include <ATen/native/TensorShape.h>
#include <torch/csrc/jit/runtime/custom_operator.h>
#include <torch/csrc/jit/serialization/storage_context.h>
#include <torch/csrc/jit/serialization/import.h>
#include <ATen/cudnn/Descriptors.h>
#include <ATen/cudnn/Types.h>
#include <c10/cuda/CUDAGuard.h>
```
should be mapped, ... baring a lot of exceptions and a lot of missing template instantiations.

The CUDA versions of the library now includes PTX code for compute capability 7.0 and binary codes for 5.0 and 6.0. This means you need a Maxwell card minimum, and that at first launch, if you have a Volta or more recent card, the library will compile the PTX code of all libtorch kernels into binaries for your card. This can take 10 to 20 minutes. The binaries are stored in a cache for future launches. If your application still lag at startup after first launch, this means your cache size is too low. Try to increase it by setting environment variable CUDA_CACHE_MAXSIZE to 2000000.
If this is not acceptable for your needs and you cannot afford the delay at first launch, or if you want to benefit from newer compute capabilities, you can install libtorch from pytorch.org and set the `org.bytedeco.javacpp.pathsFirst` system property to `true`.

Please give it a test and report here any problem, comment, missing API or other RFE.


saudet:
Could you undo unnecessary changes in indenting?

saudet:
> > Could you undo unnecessary changes in indenting?
> 
> I reindented using rules closer, I think, that the ones you use.
> 
> Also added 2 missing includes.

I mean, could you please not change the lines that you haven't touched? It makes it harder to review.

HGuillemet:
I removed the includes depending on CUDA includes.
But, if needed, I guess there is a way to patch the include list depending on the availability of cuda, like you did for libraries.

saudet:
The build is still failing on Windows. Please fix this!

HGuillemet:
Pytorch tries a dlopen on `nvfuser_codegen` on my machine at startup and issues a warning because it does not found it.
I added this library to the list of preloads.  No warning anymore. Was it the right thing to do ?

saudet:
I had already added that library. The question should be, why did you
remove it? Please don't remove stuff for no reason.

HGuillemet:
The answer is that I did't remove anything. I have started this before your update for Pytorch 2.0 and I didn't merge your changes.

HGuillemet:
About CUDA architectures: the presets currently builds libtorch with `TORCH_CUDA_ARCH_LIST="5.0+PTX"`. This has the advantage to support almost all currently existing architectures and all future ones, but the following disavantages:
* the CUDA compute capabilities are limited to 5.0, which is quite old
* the PTX must be compiled by the JIT at the first launch of an application using the presets. This can takes more than 10 minutes since all existing pytorch kernels are compiled, not only the one used by the application. The result of the JIT compilation is cached for next launches. However the default cache is limited to 1Gb. And the kernel binaries need 1,2Gb (at least for my Ampere GPU). So even the next launches takes a couple of minutes since 200M of binaries are missing. The solution for the user is to rise the cache limit with the CUDA_CACHE_MAXSIZE environment variable.

For these 2 reasons, I wonder if we cannot set something better for `TORCH_CUDA_ARCH_LIST`.
This is about a tradeoff between the size of `libtorch_cuda` and both the efficiency of code and launch time.
Here are some figures I tested:

|TORCH_CUDA_ARCH_LIST | Size of libtorch_cuda|
|--|--|
|5.0+PTX | 414754504|
|All (something like 3.5;5.0;8.0;8.9;9.0) | 690895008|
|5.0;5.2;5.3;6.0;6.1;6.2;7.0;7.2;7.5;8.0;8.6;8.7;8.9;9.0 | 1561317488|

Any opinion about this ?
How much do we care about libtorch_cuda size ? (I'd say little).
Can we drop support for oldest architectures ?
Can we drop forward compatibility ?

saudet:
The problem is GitHub Actions. We only have 6 hours of build time. If you can figure out how to make this work with GitHub Actions, go ahead.

HGuillemet:
Is using the binaries of libtorch provided by pytorch.org an option ?

saudet:
They don't provide binaries for CUDA 12

HGuillemet:
They will. And in the meantime, is it a problem to lag a bit and still link pytorch with the previous version of Cuda ?

I don't understand the time took by the action. When there is no build error, it only takes 155min max for one platform, but can time out when there is an error. Do you know why ?
Doesn't this mean we have a big margin if we spawn one job per platform and per presets ?

saudet:
> They will. And in the meantime, is it a problem to lag a bit and still link pytorch with the previous version of Cuda ?

We could maintain that in a branch, sure. If you want to do that, I'll give you write access.

> I don't understand the time took by the action. When there is no build error, it only takes 155min max for one platform, but can time out when there is an error. Do you know why ?

The actions retry 5 times in a row by default. You can disable that temporarily on your fork to speed things up when you expect things to fail.

> Doesn't this mean we have a big margin if we spawn one job per platform and per presets ?

That's how it's already doing it

saudet:
Please resolve the conflicts.

pytorch/cppbuild.sh
```diff
@@ -22,7 +22,7 @@ if [[ "$EXTENSION" == *gpu ]]; then
     export USE_CUDNN=1
     export USE_FAST_NVCC=0
     export CUDA_SEPARABLE_COMPILATION=OFF
-    export TORCH_CUDA_ARCH_LIST="5.0+PTX"
+    export TORCH_CUDA_ARCH_LIST="7.5+PTX"
```
saudet, line 25: Please revert that back to at least 5.0+PTX. This won't work on older cards

HGuillemet: We must make a tradeoff between support of older cards and support for new computation features. For instance support for half floats won't work (I think) with 5.0+PTX. I thought in 2023 not much people will try to use these presets with pre-turing cards, but I may be wrong.

saudet: Like I keep telling you, users still have the choice to use LibTorch. 

In any case, like I also told you many times now, please stop trying to do everything at the same time. Revert that and anything else I tell you to, and let's keep discussing that and the rest of your laundry list in future pull requests. Please keep them as short as possible!

HGuillemet: I will revert. Anyway any serious use of the presets will probably need to use libtorch or custom recompiling.

But this PR is like a major release or rework of the presets. if you don't like that or don't trust me for maintaining it, feel free to not merge it, or merge some features one by one, and I will keep this version on my fork.

saudet: No, that's not the point, the point is to do one thing at a time. If we want to change the default CUDA archs, it's not only about PyTorch, it's also about OpenCV, ONNX Runtime, and TVM. These projects don't care about interoperability of their libraries, but I do. If you also don't care about using OpenCV, ONNX Runtime, TVM, and PyTorch in the same application, then you should do like @Craigacp who works upstream in the ONNX Runtime repository and publishes binaries that don't work well with OpenCV, etc, and work upstream in the PyTorch repository to work on Java bindings there, and do the same, yes.

saudet:
Ok, I've finally got the release out of the way and updated the dependencies to build on Ubuntu. Please update the version to 2.0.1-1.5.10-SNAPSHOT, remove centos from the workflow, and we can start publishing snapshots under that version, for now. Thanks!

HGuillemet:
> Ok, I've finally got the release out of the way and updated the dependencies to build on Ubuntu. Please update the version to 2.0.1-1.5.10-SNAPSHOT, remove centos from the workflow, and we can start publishing snapshots under that version, for now. 

CMake cannot find NCCL. Not sure it's related but is it on purpose that you use rhel rpm instead of ubuntu packages from nvidia ?

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
```diff
@@ -1522,235 +1203,1643 @@
...
-                "torch/data/dataloader_options.h",
-                "torch/data/dataloader/stateful.h",
-                "torch/data/dataloader/stateless.h",
-                "torch/data/datasets.h",
-                "torch/data/datasets/base.h",
+            },
+            exclude = {
+                // We only want to include torch.h in JNI file and the
+                // additional .h listed above
+                // Listing all others for exclusion
```
saudet, line 1479: Why do you need to do that?

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
```diff
@@ -0,0 +1,216 @@
...
+import org.bytedeco.javacpp.tools.InfoMap;
+import org.bytedeco.javacpp.tools.InfoMapper;
+
+import java.util.List;
+
+/**
+ * @author Samuel Audet, Hervé Guillemet
+ */
+@Properties(
+    //inherit = torch.class,
```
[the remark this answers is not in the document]
HGuillemet: If I remember well, if we inherit, all headers from the main presets will be reparsed, while we need only a small subset here.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
HGuillemet: I didn't find a way to specify distinct lists for parsing, and for inclusion from JNI. 
In JNI we only need `#include <torch/torch.h>` (and a small number of others that are not included automatically from torch.h). 
If we manually include the big list of parsed .h in addition of torch.h, we get some compilation errors like the one triggered by : [this check](https://github.com/pytorch/pytorch/blob/37359c36fdb413df3b02996eb0ea2433c147db34/c10/util/complex_math.h#L1-L4). Or errors related to order (the order needed for JavaCPP parsing might not be the same than the one for inclusion).

If you can think of a way to avoid this big copy/paste, that would be nicer. We could use templating from maven but I don't think it's worth. Maybe adding some JavaCPP property to allow explicit specification of the header list for JNI ?

saudet: I know *some* files don't like to be reincluded, but it's a very small number of them. Why do you want to exclude all of them?

HGuillemet: We know that the JNI will compile correctly if we include the headers the way they are supposed to be included. So I preferred this systematic approach and reset the whole list instead of determining by try and error a minimal exclusion list that works. This was simpler at the cost of this copy/paste in the presets file.

Don't you think it could be a good idea to be able to decorrelate the two lists by adding a property like `jninclude` ?

saudet: That sounds reasonable. It would default to the values of include/cinclude, but users could override it, sure.

HGuillemet: Do you mean the reordering of includes such as `Half.h` or `intrusive_ptr.h` ? Those were necessary for the parsing to work, so I don't think it's related. 
If we really want to stick to the order given by `g++ -M`, to ease maintenance, they are maybe some Info that could be added that would have the same effect that reordering.

HGuillemet: About non-important headers that were commented out manually because non parseable: I have some patch to the parser that would make them parse, but as this is not blocking, I'm not sure you want them as PR.

saudet: > If we really want to stick to the order given by g++ -H, to ease maintenance, they are maybe some Info that could be added that would have the same effect as that reordering.

Right, ok... It'd be nice to automate this a bit more using clang from the LLVM presets, but anyway let's start by getting rid of the duplicate exclude list.

HGuillemet: Do you want me to propose a PR for this jniinclude property or are you doing it ?

saudet: If you can yes please open a pull request. Thanks!

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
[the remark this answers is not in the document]
HGuillemet: Yes, Parser parses [`allIncludes` which contains inherited includes](https://github.com/bytedeco/javacpp/blob/b9f91d7f41e99f0f209fbc5ef318ef2217c89dc8/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L4343-L4356). What is confusing ? Isn't it meant to work like that ?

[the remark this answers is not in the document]
saudet: No, that's fine, it's just that the only reason you're not using is because it's slow, and what you're doing is adding technical debt by maintaining more lines of code.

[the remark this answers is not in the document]
HGuillemet: What about defining a common abstract parent ?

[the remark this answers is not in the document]
saudet: No need to do that, see https://github.com/bytedeco/javacpp/pull/689#issuecomment-1589240613.

pytorch/cppbuild.sh
saudet: It looks like it's able to build within 6 hours with "5.0;6.0;7.0+PTX":
https://github.com/bytedeco/javacpp-presets/actions/runs/5255778236/jobs/9496144498
This means any GPU with CC 5.x, 6.x, or 7.x can load the binaries, while newer ones with CC 8.x and 9.x need to compile from PTX. Is that good enough for you?

pytorch/src/main/resources/org/bytedeco/pytorch/presets/torch_parselist
```diff
@@ -0,0 +1,1415 @@
...
+torch/csrc/jit/serialization/pickler.h
+
+# Parsed and for inclusion in JNI
+# See also https://github.com/pytorch/pytorch/blob/main/docs/cpp/source/Doxyfile
+# for an approximation of what should be in API in addition to torch.h
+# torch/csrc/jit/runtime/custom_operator.h:  Name conflict with torch::RegisterOperator + little chance to have any use
+torch/torch.h
+ATen/native/TensorShape.h
+torch/csrc/jit/serialization/storage_context.h
+torch/csrc/jit/serialization/import.h
```
saudet, line 1415: Instead of trying to come up with some sort of new propriety format just for fun, please use an existing format. I'd go with a subset of C++ that can be interpreted as a normal header like `torch_include.h` like this:
```cpp
// All files included by #include <torch/torch.h>
// as listed by g++ -M torch/csrc/api/include/torch/all.h (or -H)
// Excluding:
// - the ones that fill at::meta at::native and at::_ops namespaces
//   (ATen/ops/*_native.h ATen/ops/*_meta.h ATen/ops/*_ops.h)
// - ATen/ops/_*
// - and some exceptions commented below
#include "torch/csrc/utils/python_stub.h"
#include "c10/macros/cmake_macros.h"
#include "c10/macros/Export.h"
...
```

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
```diff
@@ -2909,13 +100,50 @@ public class torch implements LoadEnabled, InfoMapper {
         Loader.checkVersion("org.bytedeco", "pytorch");
     }
 
+    static void initIncludes(Class thisClass, ClassProperties properties) {
+        // If we are called from Parser, fetch the list of headers to parse from resources
+        if (!Loader.isLoadLibraries()) {
+            for (int i = 3; i < 10; i++) {
+                Class caller = Loader.getCallerClass(i);
```
saudet, line 107: You don't need to have a loop here. Just use 3 if that works, and if it changes later, which it probably wont, but if it does we can always easily fix that.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
```diff
@@ -2909,13 +100,50 @@ public class torch implements LoadEnabled, InfoMapper {
...
+                if (caller == null) return;
+                if (caller == Parser.class) {
+                    properties.put("platform.include", new ArrayList<String>());
+                    Class presets = properties.getEffectiveClasses().get(0);
+                    // We don't want to fill the include list of torch when we are processing torch_cuda
```
saudet, line 112: So, why are you calling it from torch_cuda.class?

pytorch/cppbuild.sh
HGuillemet: I'm using custom build for my needs anyway. The question is what's the best for users, not for me.

It's a tradeoff between available CC, startup time, binary size, and old hardware support .
You also talked about interoperability, I'm not sure to understand how it's related. 

I'd say binary size is not important so "5.0;6.0;7.0+PTX" is better than "5.0+PTX". 7.5 introduces support for Half (FP16) which is important nowadays so I'd try to include 7.5. Maybe with "5.0+PTX;7.5+PTX" ?

pytorch/src/main/resources/org/bytedeco/pytorch/presets/torch_parselist
HGuillemet: I used what was the easiest to parse while allowing comments.
So that we introduce the minimum and simplest code to maintain, which usually is the way you prefer.
Why not a .h style format... Is there an easy way to bulk-insert in into the JNI without parsing ?
Else I guess I can parse that with a regexp without much more line of codes.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
HGuillemet: I do because the index of Parser in the stack trace is not the same when coming from torch parsing or torch_cuda parsing.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
HGuillemet: To read torch_cuda_parselist ? 
Because of the property inheritance I added in the commit, both torch.init() and torch_cuda.init() are called when processing torch_cuda, and we don't want to populate the include property of torch when processing torch_cuda or we will reparse reparse the whole stuff.

pytorch/src/main/resources/org/bytedeco/pytorch/presets/torch_parselist
HGuillemet: Forget what I said about bulk-inserting. We are talking about the list of headers to parse, not to include in JNI.

pytorch/cppbuild.sh
saudet: CC 7.0 looks like it supports FP16, just not INT8:
https://docs.nvidia.com/deeplearning/tensorrt/support-matrix/index.html#hardware-precision-matrix

HGuillemet: Ok. 7.5 adds some FP16 features but maybe they are not that relevant. What we can also do is support old hardware with 5.0+PTX and add binaries for the cards we believe are most common at the day of the release, so that those cards get both fast startup and full features. Something like "5.0+PTX;8.6;8.9".

pytorch/src/main/resources/org/bytedeco/pytorch/presets/torch_parselist
HGuillemet: Done. It's better like this indeed.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
saudet: Right, but you're also saying you don't want the same list anyway for torch and torch_cuda, so why do you want to  catch both here when in the line right after that you only proceed if it's not torch_cuda?

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
saudet: Please don't do that. Do the inheritance, but modify platform.include differently in the case of torch and torch_cuda.

pytorch/cppbuild.sh
saudet: If you want to do research for that, please do and open an issue with your findings.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
```diff
@@ -48,6 +48,12 @@
                 "ATen/cuda/CUDAGeneratorImpl.h"
             }
         ),
+        @Platform(
+                value = "windows",
+                extension = "-gpu",
+                link = { "cudart" },
+                linkpath    = { "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.1/lib/x64/" }
+        ),
```
saudet, line 56: This is going to be necessary on Linux as well, it just won't complain until you try to call a function.

pytorch/cppbuild.sh
saudet: We probably just need to refresh the cache. BTW, it looks like NVCC is supported by ccache on Windows as well these days: https://ccache.dev/platform-compiler-language-support.html

If you want to give that a try, we could compile for any number of CUDA archs with build times of 10+ hours and offer 2+ GB binaries to users who want them.

HGuillemet: By simply relaunching the job until it's complete ? Do you know a way to automate this ?

And how would you handle "to users who want them" ? Shall we create multiple extensions ? "-gpu-ptx", "-gpu-cubins" ?

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
HGuillemet: The linux JNI was already linked with cudart, probably because the linker finds out the needed symbols and where to find them according to transitive dependencies. But it doesn't hurt to explicitly specify it.
Anyway this didn't work on windows and the job you relaunched failed again on undefined symbol. I don't know why. Another attempt running.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
HGuillemet: When we parse, init is called at lease 3 times:
1. torch.init() when processing torch presets
2. torch_cuda.init() when processing torch_cuda presets, 
3. torch.init() when processing torch_cuda presets (because of properties inheritance)

We need at least to empty the include property in 3 cases, so we need to check if we are called from Parser in 3 cases. And parser position is 5 for 1. and 2. and 6 for 3.. Then we want to fill the property with the resource file in cases 1. and 2. only

I don't think we can do much simpler.
The `isLoadLilbraries()` test at the start of `initIncludes` is useless though.

HGuillemet: However this code loads uselessly the big Parser  class when the users load the presets. Maybe we'd better test for the class name instead of `== Parser.class`, or add something to parser to avoid querying the stack track, like using a distinct initParsing() method.

saudet: > I don't think we can do much simpler.

I'm pretty sure Loader.getCallerClass(3) is going to return the Parser class when we want it, in both torch.init() and torch_cuda.init(), and return something else when we don't want it anyway, so I don't think we need a loop. Just please try it!

HGuillemet: Here are the stack traces.
```
CALLED FROM torch with effective class=torch
java.lang.Exception
        at org.bytedeco.pytorch.presets.torch.initIncludes(torch.java:107)
        at org.bytedeco.pytorch.presets.torch.init(torch.java:144)
        at org.bytedeco.javacpp.ClassProperties.load(ClassProperties.java:326)
        at org.bytedeco.javacpp.Loader.loadProperties(Loader.java:386)
        at org.bytedeco.javacpp.tools.Parser.parse(Parser.java:4343)
```
```
CALLED FROM torch with effective class=torch_cuda
java.lang.Exception
        at org.bytedeco.pytorch.presets.torch.initIncludes(torch.java:107)
        at org.bytedeco.pytorch.presets.torch.init(torch.java:144)
        at org.bytedeco.javacpp.ClassProperties.load(ClassProperties.java:326)
        at org.bytedeco.javacpp.ClassProperties.load(ClassProperties.java:162)
        at org.bytedeco.javacpp.Loader.loadProperties(Loader.java:386)
        at org.bytedeco.javacpp.tools.Parser.parse(Parser.java:4343)
```
```
CALLED FROM torch_cuda with effective class=torch_cuda
java.lang.Exception
        at org.bytedeco.pytorch.presets.torch.initIncludes(torch.java:107)
        at org.bytedeco.pytorch.presets.torch_cuda.init(torch_cuda.java:68)
        at org.bytedeco.javacpp.ClassProperties.load(ClassProperties.java:326)
        at org.bytedeco.javacpp.Loader.loadProperties(Loader.java:386)
        at org.bytedeco.javacpp.tools.Parser.parse(Parser.java:4344)
```
We can write 
```Java
if (Loader.getCallerClass(5).getName().equals("org.bytedeco.javacpp.tools.Parser") || Loader.getCallerClass(6).getName().equals("org.bytedeco.javacpp.tools.Parser")) 
```
instead if you prefer. 
But what bothers you exactly with this loop ? It seems more resilient in the long run to small modifications of javacpp.

HGuillemet: Ah, I think I understand what you mean : you would like to use `Loader.getCallerClass(5).getName().equals("org.bytedeco.javacpp.tools.Parser")` to test both that we are called from Parser and that we are not in the case `properties.getEffectiveClasses().get(0) != thisClass`.
The problem (besides code resiliance) is that, as said above, we must reset the include property (`properties.put("platform.include", new ArrayList<String>())`) when we are called by the parser in all cases.

saudet: > 

> The problem (besides code resilience) is that, as said above, we must reset the include property (`properties.put("platform.include", new ArrayList<String>())`) when we are called by the parser in all cases.

No we don't. What happens if you don't do that?

HGuillemet: Ok, I see: resetting the properties only once when processing torch_cuda is enough.
So we can replace the loop and the check for effective class by a check for depth 5 only. It's a bit cryptic but there will be less lines of code.

saudet:
JavaCPP now supports template instantiation for std::function, with Info.define(), see issue #1051, just like for std::vector, etc so we shouldn't need the "functions" packages you created. Please give that a try!

HGuillemet:
> JavaCPP now supports template instantiation for std::function, with Info.define(), see issue #1051, just like for std::vector, etc so we shouldn't need the "functions" packages you created. Please give that a try!

I did. For example for `ArchiveWriter`, two classes are created: `ArchiveWriter`, mapping `std::function<size_t(const void*,size_t)>`, and a subclass of `FunctionPointer`, called `ArchiveWriter_Pointer_long`. Why do we need two classes while only one is necessary when we define the type manually ?

saudet:
> I did. For example for `ArchiveWriter`, two classes are created: `ArchiveWriter`, mapping `std::function<size_t(const void*,size_t)>`, and a subclass of `FunctionPointer`, called `ArchiveWriter_Pointer_long`. Why do we need two classes while only one is necessary when we define the type manually ?

That's the point, so we don't need to define those classes manually. You should be able to remove all the manually written classes in the functions package, and rely entirely on JavaCPP to generate them at build time, based on the information in the header files.

saudet:
> I did. For example for ArchiveWriter, two classes are created: ArchiveWriter, mapping std::function<size_t(const void*,size_t)>, and a subclass of FunctionPointer, called ArchiveWriter_Pointer_long. Why do we need two classes while only one is necessary when we define the type manually ?

Ah, sorry, I didn't catch your question correctly.  The manually defined classes only implement a FunctionPointer, but it's not possible to convert a FunctionPointer to std::function, so when you have a function that returns a std::function, we can't call it without std::function. If what you're really asking is why we can't inherit or extend std::function somehow, I do not know. Like I always try to make you understand, I do not have all the answers! I suggest you ask that kind of question to the C++ committee.

HGuillemet:
Thanks for the explanation.
Another question: why the default class name for the function pointer is named after the function signature ? We need a different pointerType for each `std::function<XXX>` anyway.
Why not using something like an inner class `ArchiveWriter.FunctionPointer`  ?
Anyway I see we can customize it by adding an info like:
`new Info("size_t (*)(const void*, size_t)").pointerTypes("ArchiveWriterFunctionPointer")`
(but we cannot create an inner class this way).

saudet:
> Why not using something like an inner class `ArchiveWriter.FunctionPointer` ? 

Because no one has spent time on this yet? I keep telling it's just not possible for me to do everything on my own!

So anyway, if that's the reason why you want to keep the `functions` package, I guess that's fine.

HGuillemet:
> > Why not using something like an inner class `ArchiveWriter.FunctionPointer` ?
> 
> Because no one has spent time on this yet? I keep telling it's just not possible for me to do everything on my own!

Why do you take my question as a reproach or a criticism ? I just thought there is a reason for it to be like this and asked what it is.
If you don't remember a reason, and think that using an inner class is a good idea, I'm wiling to take some time and post a PR.

> 
> So anyway, if that's the reason why you want to keep the `functions` package, I guess that's fine.

Writing each function by hand is error-prone and not so straightforward, so having something to automate this is a good thing. I'd prefer to replace that with the define().

saudet:
@frankfliu @lanking520 @nguyenvietyen @zachgk @hal-314 @hulk-us @guangster @jbaron @jxtps @mehmetcalikus @napps2017 @scott01272001 @tgg9527 @carlosuc3m @doofin @wumo @zaleslaw @HichemMaiza @luyulei-lyl @oOTWK @doudouyzgf @lzmchina @mullerhai @sbrunk

@HGuillemet has been busy expanding and improving the presets for PyTorch quite a bit, and notwithstanding the mappings for std::function, this pull request looks good to merge so that we can start getting snapshots and test some more. However, those changes do break some things, so if there's anything you'd like to get fixed before we do that, please let us know now! Thank you

sbrunk:
It's not going to break the published version of Storch, since we have the stable `2.0.1-1.5.9` artifacts and the current Storch version depends on that.

So while it probably requires changes on my side, it's actually good to have snapshot builds to work against, test the improvements and see what needs to be changed.

jxtps:
Improving the presets sounds great, the active surface area I use is pretty small so I’m sure it won’t be too big of a deal to upgrade.  

My only concern is with the CUDA compute capability versions, where I’m both very interested in small size and prompt startup (no first time compile) - it already takes “forever” (= 7-10 minutes) to start one of my inference instances (g5.2xlarge on AWS), probably mostly due to needing to load gigabytes of libtorch (haven’t profiled, but my non-DL instances boot in <2 minutes and only substantive difference is torch). 

If there’s a way to prune the compute capability list or perhaps do per-capability jars that would be greatly appreciated! 

Happy to help out here of course, though I don’t really know where I’d start…

HGuillemet:
@jxtps That's weird that with libtorch the startup time is so important, since it includes CUBINs for almost all compute capabilities. Have you checked that the library loaded is actually libtorch ? (you can check `/proc/<process id>/maps` if you run on linux)

Another thing to check is the `~/.nv` directory that contains the result of JIT compilation at startup, when you use PTX and not CUBINs. Is it filled ? If yes, you can try to increase its size of this cache as explained in the first post of this thread.

jxtps:
@HGuillemet checking /proc/nnn/maps shows the /usr/lib/jni/libtorch_* that I put there being loaded. 

~/.nv/ only has ComputeCache/ in it, and that directory is empty, so it doesn't seem to do any CUBIN/PTX compilation at startup. 

I think the explanation lies in how AWS starts EBS backed instances - the AMI is under the hood stored on S3 (not visible to me as a user), then when the instance starts it's restored to EBS, but for starters just with the meta information, then the actual files are transferred over time & as they're needed. Libtorch requires almost 4gb of data, and S3 -> EBS can be as slow as 10 megs/sec, so that can be >6 minutes right there. 

If it's possible to reduce the Torch size (which is how I read your cuda compute capability comments above), then under this model that would linearly reduce my startup time. 

(I'm separately looking into not having libtorch baked into the AMI and instead transferred on startup, which could then be done in parallel etc, but that requires significant modifications to my deploy system and would also benefit from any reductions in libtorch size)

HGuillemet:
@jxtps  You can always clone the javacpp-presets repo and build your own version of the pytorch presets, after changing  the `TORCH_CUDA_ARCH_LIST` line in `cppbuild.sh` to only include the compute capability you are needing.

We probably also could distribute one org.bytedeco.pytorch artifact per compute capability, with different classifiers. That would take several days of github actions to build all of them. I'm not sure it's realistic.

doofin:
thanks for the contribution. I previously encounter memory issues while implement transformers. there seems to be thousands of files changed in this commit, breaking it into smaller ones maybe better for review(next time)

jxtps:
@HGuillemet maybe do two versions, one for “latest” or “modern” (8 & 9?) and one for “compatible” or “most” (whatever is in there right now)?

HGuillemet:
> > So anyway, if that's the reason why you want to keep the `functions` package, I guess that's fine.
> 
> Writing each function by hand is error-prone and not so straightforward, so having something to automate this is a good thing. I'd prefer to replace that with the define().

After spending some time on this, it turns out to be a bit difficult. I found some strange behaviors that looks like bugs. A couple of PR to JavaCPP coming. One of the problem is that function parameters with const and adapters (like a `const std::string`) is not translated into a `@Const` annotation in the class generated by an info like `new Info("std::function<void(const std::string&)>").pointerTypes("F").define()`, causing C++ compilation errors. @saudet, if you have an idea why it's like this, this could save me some time.

So I'm not sure the usage of std::function containers will make it in this PR of the pytorch presets. I haven't found any native function returning a `std::function` anyway.

HGuillemet:
After spending some more time on `std::function` container, I have some patches that improve things, but a problem remains for which I can't find a solution:
```Java
new Info("std::function<void(const std::shared_ptr<torch::nn::Module>&)>").pointerTypes("SharedModuleApplyFunction").define();
```
The `FunctionPointer` created by parser must contain:
```
    public native void call(@SharedPtr @ByRef @Cast({"", "std::shared_ptr<torch::nn::Module>"}) Module m);
```
But there is no way for the parser to get the information that would make it generate the `@Cast`, with the precast argument, since it doesn't know about adapters.
Maybe the generator could generate a cast from the `@SharedPtr` annotation alone, but then  it doesn't know the cpp type to cast to.
@saudet any idea or shall I give up function containers ?

HGuillemet:
Can we proceed with snapshots ?
Or shall we first discuss further about compute capabilities and the option to publish multiple GPU classifiers like '-ptx-5.0', '-cubin-8.6', etc... ?

saudet:
If you're not planning on pushing more changes, I think we can merge this,
yes.

HGuillemet:
No more changes for now.

saudet:
Please update for https://github.com/bytedeco/javacpp/pull/696

sbrunk:
I just built this locally to get a better estimate how much I'll have to change when upgrading.

One thing I noticed is that sqrt for `Tensor` in `global.torch` is missing:
```diff
// aten::sqrt(Tensor self) -> Tensor
- @Namespace("at") public static native @ByVal Tensor sqrt(@Const @ByRef Tensor self);
```
Instead we have the complex variants:
```java
@Namespace("c10_complex_math") public static native @ByVal @Name("sqrt<float>") FloatComplex sqrt(@Const @ByRef FloatComplex x);

@Namespace("c10_complex_math") public static native @ByVal @Name("sqrt<double>") DoubleComplex sqrt(@Const @ByRef DoubleComplex x);
```

Probably due to this mapping:
```java
        infoMap.put(new Info("c10_complex_math::sqrt<float>").javaNames("sqrt"))
               .put(new Info("c10_complex_math::sqrt<double>").javaNames("sqrt"))
```
Is it possible to keep the original `sqrt` mapping somehow?

saudet:
I don't think it's due to those, but if it works by adding one more for at::sqrt() as well, let's just do that?

HGuillemet:
`complex_math.h` is parsed before `ATen/ops/sqrt.h`, where `at::sqrt` is defined.
So when `ATen/ops/sqrt.h` is parsed, because of [this loop](https://github.com/bytedeco/javacpp/blob/8646e971c46a168cb9d04d86fe05f07d44da0dd3/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L1575-L1584), sqrt qualified name is set to the first matching cppName it finds a info for. So `c10_complex_math::sqrt` in this case.

I added an explicit info for `at::sqrt` to prevent this. And I added the other missing complex math operators for good measure.

sbrunk:
`sqrt` is working again. Thanks! I got Storch compiling now with minimal changes (just a few type names, additional parameters needed etc.).

I'm seeing a bunch of crashes now when running our tests, especially in [IndexingSlicingJoiningOpsSuite](
https://github.com/sbrunk/storch/blob/e8d6b3fffb0d38a733ed2e9f2f3b0ef8619ce7b3/core/src/test/scala/torch/ops/IndexingSlicingJoiningOpsSuite.scala).I think they are starting to become a nice regression test suite for the native bindings. :)

I've tried to isolate the crashing ops. Most seem to take lists of tensors in some form, like `stack` or `cat`:

```java
import org.bytedeco.pytorch.*;
global.torch.stack(
  new TensorArrayRef(
    new TensorVector(
      AbstractTensor.create(0),
      AbstractTensor.create(1)
    )
  ),
  0
).print();
```

```
[info] # C  [libtorch_cpu.so+0x1de6e4f]  at::_ops::stack::call(c10::ArrayRef<at::Tensor>, long)+0x9f
```

The same as above, `cat` instead of `stack`:
```
[info] # C  [libtorch_cpu.so+0x1cc0a2c]  c10::detail::MultiDispatchKeySet::operator()(c10::IListRef<at::Tensor>)+0xcc
```

saudet:
Should we wait after that is fixed to merge this pull request?

sbrunk:
From my side it's fine merging now and then fixing this in a separate PR.

I'd actually like to try with the CI snapshots to rule out that the issue is caused by me building locally.

HGuillemet:
I can reproduce your crash, so it's not related to your local build.
@saudet, please give me some minutes before merging in case I quickly find out the reason for this.

HGuillemet:
The ArrayRef constructor taking a Vector isn't mapped any more. So it's the "pointer-cast" constructor that is called instead in your case, causing the crash.
I'll try to re-add it but you can also use the constructor taking a Tensor and a length instead.

sbrunk:
So you mean s.th. like this, relying on the fact that vectors are stored contiguously?

```java
var vector = new TensorVector(AbstractTensor.create(0), AbstractTensor.create(1));
global.torch.stack(
  new TensorArrayRef(vector.front(), vector.size()),
  0
).print();
```

sbrunk:
So you mean s.th. like this, relying on the fact that vectors are stored contiguously?

```java
var vector = new TensorVector(AbstractTensor.create(0), AbstractTensor.create(1));
global.torch.stack(
  new TensorArrayRef(vector.front(), vector.size()),
  0
).print();
```

HGuillemet:
Or something like:
```
Tensor t = new Tensor(2);
t.put(AbstractTensor.create(0));
t.position(1).put(AbstractTensor.create(1));
torch.stack(new TensorArrayRef(t.position(0), 2)).print();
```

pytorch/src/gen/java/org/bytedeco/pytorch/BlockArrayRef.java
```diff
@@ -41,8 +41,7 @@ public class BlockArrayRef extends Pointer {
 
   /** Construct an ArrayRef from a single element. */
   // TODO Make this explicit
-  public BlockArrayRef(@ByPtrRef Block OneElt) { super((Pointer)null); allocate(OneElt); }
-  private native void allocate(@ByPtrRef Block OneElt);
+
```
saudet, line 44: This seems like an odd thing to remove, but maybe not something that anyone actually uses? @sbrunk ?

sbrunk: No. Things in `torch::jit` are for TorchScript support and we're not there yet. Besides, the constructor isn't available in the current bindings either.

saudet (approved):

sbrunk:
This is great! Thanks for these improvements @HGuillemet!

PR #1361, opened 2023-05-25
Add higher level tritonserver API

jbkyang-nvi:
With https://github.com/triton-inference-server/developer_tools/tree/main/server, 
Triton added a C++ API that is higher level API for the original API.
This PR adds `tritonserverwrapper` project that makes use of that API. It is higher than the 
Javacpp/tritonserverwrapper project


saudet:
Thanks! Instead of adding a new presets though, I think we could simply update the original tritonserver one with the new "tritonserverwrapper"? BTW, we typically these modules after the library names, which is in this case "tritondevelopertoolsserver"...

jbkyang-nvi:
> Thanks! Instead of adding a new presets though, I think we could simply update the original tritonserver one with the new "tritonserverwrapper"? BTW, we typically these modules after the library names, which is in this case "tritondevelopertoolsserver"...

`tritonserver` wraps around the CAPI and the new `tritondevelopertoolsserver` API wraps around a higher level version of that. I think both work a little differently and it would be nice to have them as different projects, if that makes sense.

saudet:
It makes it harder to maintain and for users to use as well. Let's please
not do that. I'm not going to maintain them in that form myself and they
are currently not functional because they are missing too many
configurations.

How about this, if you're not going to maintain the C interface, let's nuke
it?

jbkyang-nvi:
I see your point that it will be harder to maintain however I do have some concerns with respect to totally nuking the current tritonserver project.

The pros of having CAPI-Javacpp bindings are 1. tritonserver.h is the core piece of server code and it's relatively stable and well tested. It is fast. 2. There are currently users who are using the CAPI-Javacpp presets and they will need to migrate their CAPI-Javacpp code to the C++-Javacpp code (the one I'm trying to merge)

The developer_tools repo is adding sugar to simplify the interface but has not a lot of performance guarantees. I don't think nuking the original CAPI-Javacpp bindings would be super beneficial. 

Going forward, I have a couple of suggestions, let me know your thoughts on these:
1. Keep one tritonserver project but include both developer_tools/generic_server_wrapper.h and tritonserver.h. This would be a bigger project to maintain, but will be less confusing for the user. This is because the user can use either the CAPI or the C++API bindings with Java.
2. Keep 1 of the 2 projects inside the Triton repository. I am not sure this makes more sense than option 1. We can do it if that's better.
 
Not sure about what you mean by
> that form myself and they
are currently not functional because they are missing too many
configurations.

Both projects are usable as is with the sample code. What do you think is missing/can be improved?
Triton has existing tests around the CAPI-Javacpp and C++API-Javacpp API. Is there anything special about either presets that makes them harder than normal projects to maintain?

saudet:
> Going forward, I have a couple of suggestions, let me know your thoughts on these:
> 
> 1. Keep one tritonserver project but include both developer_tools/generic_server_wrapper.h and tritonserver.h. This would be a bigger project to maintain, but will be less confusing for the user. This is because the user can use either the CAPI or the C++API bindings with Java.

Right, that's what I proposed, but you said you didn't want to do that. Are you saying this is fine now?

> > It makes it harder to maintain and for users to use as well. Let's please
> > not do that. I'm not going to maintain them in that form myself and they
> > are currently not functional because they are missing too many
> > configurations.
> 
> Both projects are usable as is with the sample code. What do you think is missing/can be improved? Triton has existing tests around the CAPI-Javacpp and C++API-Javacpp API. Is there anything special about either presets that makes them harder than normal projects to maintain?

There's no README.md file and no configuration for the CI, but I didn't check in detail. I'm sure there are a couple of other things missing.

If what you have here is good enough for the Triton team though, then by all means, please do maintain it on a repository over there, with all the tests and everything. I'll still be able to help you out with stuff that don't work on request. If you want me to maintain this here though, I'm not going to increase the effort required to do that, especially since I also believe it will make it harder for users to use because they would have to choose between 2 different sets of dependencies to use essentially the same library.

jbkyang-nvi:
> > Going forward, I have a couple of suggestions, let me know your thoughts on these:
> > 
> > 1. Keep one tritonserver project but include both developer_tools/generic_server_wrapper.h and tritonserver.h. This would be a bigger project to maintain, but will be less confusing for the user. This is because the user can use either the CAPI or the C++API bindings with Java.
> 
> Right, that's what I proposed, but you said you didn't want to do that. Are you saying this is fine now?

I think there was some misunderstanding in what "nuking tritonserver" means. I thought you _only_ wanted tritonserverdevelopertools folder.

jbkyang-nvi:
@saudet can you please take a look?

saudet:
Thanks! This is looking good. Let's also update the versions in .github/workflows/tritonserver.yml, tritonserver/README.md, platform/pom.xml, the other pom.xml files under tritonserver, and the main README.md file.

saudet:
I've fixed a few things, but now I'm hitting an error that indicates an `#include <string>` is missing from `/opt/tritonserver/include/triton/developer_tools/common.h`. Could you please add a line like this?
```
2023-06-07T01:53:35.2152590Z In file included from /__w/javacpp-presets/javacpp-presets/tritonserver/target/native/org/bytedeco/tritonserver/linux-x86_64/jnitritonserver.cpp:105:
2023-06-07T01:53:35.2235872Z /opt/tritonserver/include/triton/developer_tools/common.h:65:15: error: field 'message_' has incomplete type 'std::string' {aka 'std::__cxx11::basic_string<char>'}
2023-06-07T01:53:35.2236794Z    65 |   std::string message_;
```

jbkyang-nvi:
@saudet sorry I did a force push without your changes 😓. Can you push them back?
 I made the changes on `common.h`

saudet:
Thank you! It's almost passing, but now it seems to be failing because of something different with the version of GCC in Ubuntu 22.04 that NGC has been upgraded to. :thinking:

saudet:
I see what's going on, it's not because of GCC. The mapping you've added for String doesn't work for the C API, so when we remove that it should go back to normal.

What I had in mind from the beginning was to put the new C++ API in another presets that match the library name like presets/tritondevelopertoolsserver.java. This way also it won't affect anything for presets/tritonserver.java. Sounds good?

jbkyang-nvi:
> What I had in mind from the beginning was to put the new C++ API in another presets that match the library name like presets/tritondevelopertoolsserver.java. This way also it won't affect anything for presets/tritonserver.java. Sounds good?

so @saudet would that mean we will have 2 tritonserver related bindings then?

saudet:
We can have multiple presets per module, for example:
https://github.com/bytedeco/javacpp-presets/tree/master/opencv/src/main/java/org/bytedeco/opencv/presets
We can do the same in this case:
https://github.com/bytedeco/javacpp-presets/tree/master/tritonserver/src/main/java/org/bytedeco/tritonserver/presets

jbkyang-nvi:
Updated the PR. I noticed there's some issues caused by updating the tritonserver.h from 23.04 to 23.05 though [comparison link](https://github.com/triton-inference-server/core/compare/r23.04...r23.05). Through this PR: https://github.com/triton-inference-server/core/pull/192 we made tritonserver.h conform to proper C standard (vs C++ standard).

tritonserver.java created seemed to be confused and ommited a few functions that used defined variables.: https://github.com/bytedeco/javacpp-presets/pull/1361/commits/63431c6b96b5a60e784a85c44a75811a396e9021#diff-0b23a69d7a0bd76ec9fff99224228855a84574ccaf9185df0c5d774ba00dc933

I tried to use cpptext and it also did not solve the problem locally. Is there anything else I should be doing? I checked, and 23.04 seems to be compiling fine. 
https://github.com/bytedeco/javacpp-presets/commit/e90149b0d58d92353cd3b32666391a6bedb5e3ae

saudet:
Right now it's not failing because, but because of this:
```
2023-06-10T02:51:14.7189856Z ../cppbuild.sh: line 10: INCLUDE_DEVELOPER_TOOLS_SERVER: unbound variable
```
Let's start by fixing this

saudet:
I've fixed the parser issue with function pointers in commit https://github.com/bytedeco/javacpp/commit/b9f91d7f41e99f0f209fbc5ef318ef2217c89dc8 so it should work fine now.

jbkyang-nvi:
@saudet Thanks for your help! It looks like samples/Simple.java is running fine but the example for samples/SimpleCPP.java does not work. Is this something with how platform/pom.xml is written?

saudet:
What do you mean by does not work? What's the error message that you get?

jbkyang-nvi:
> What do you mean by does not work? What's the error message that you get?

It simply does not find the C++ binding files. I checked in the compiled jar and it seems like all the generated java files are in there

saudet:
Could you run the sample with `mvn clean compile exec:java -Dorg.bytedeco.java.logger.debug` and copy/paste here what that outputs?

jbkyang-nvi:
> Could you run the sample with `mvn clean compile exec:java -Dorg.bytedeco.java.logger.debug` and copy/paste here what that outputs?

I think I found the reason of the failure. Thanks. Let me know if there's anything else that needs to be change

tritonserver/src/main/java/org/bytedeco/tritonserver/presets/tritonserver.java
```diff
@@ -65,6 +65,18 @@ public void map(InfoMap infoMap) {
...
+               .put(new Info("struct TRITONSERVER_InferenceResponse").cppText("TRITONSERVER_InferenceResponse"))
+               .put(new Info("struct TRITONSERVER_InferenceTrace").cppText("TRITONSERVER_InferenceTrace"))
+               .put(new Info("struct TRITONSERVER_Message").cppText("TRITONSERVER_Message"))
+               .put(new Info("struct TRITONSERVER_Metrics").cppText("TRITONSERVER_Metrics"))
+               .put(new Info("struct TRITONSERVER_Parameter").cppText("TRITONSERVER_Parameter"))
+               .put(new Info("struct TRITONSERVER_ResponseAllocator").cppText("TRITONSERVER_ResponseAllocator"))
+               .put(new Info("struct TRITONSERVER_Server").cppText("TRITONSERVER_Server"))
+               .put(new Info("struct TRITONSERVER_ServerOptions").cppText("TRITONSERVER_ServerOptions"))
+               .put(new Info("struct TRITONSERVER_Metric").cppText("TRITONSERVER_Metric"))
+               .put(new Info("struct TRITONSERVER_MetricFamily").cppText("TRITONSERVER_MetricFamily"));
```
saudet, line 80: I don't think we need those changes. Could you try to revert this and see if it breaks anything?

saudet:
Looks good! Thanks

saudet (approved):

PR #1373, opened 2023-06-13
Build cpython for Mac OSX ARM64

nightscape:
I'm not sure if I'm overlooking something, but here's a shot of building cpython for Apple Silicon.
I tried a locally built JAR of this on my MacBook M1 by running the `Sample.java` file and it correctly printed the formatted date.
@saudet what am I missing? Is it really that easy? 😅

### DIFF
--- a/.github/workflows/cpython.yml
+++ b/.github/workflows/cpython.yml
@@ -41,6 +41,10 @@ jobs:
     runs-on: macos-11
     steps:
       - uses: bytedeco/javacpp-presets/.github/actions/deploy-macosx@actions
+  macosx-arm64:
+    runs-on: macos-11
+    steps:
+      - uses: bytedeco/javacpp-presets/.github/actions/deploy-macosx@actions
 #  windows-x86:
 #    runs-on: windows-2019
 #    steps:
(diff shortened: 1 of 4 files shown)


nightscape:
@saudet I think I missed changing the architecture to `arm64` while `configure`ing CPython.
This worked on my machine because that is a native ARM64 M1 chip, but in Github Actions we're cross-building.
Another question that comes to mind is how the cross-building of NumPy itself would work. It cannot use the published arm64 artifact of CPython because that wouldn't run on x86_64.
Did you already solve this in other places?

saudet:
I used crossenv for that before, but it doesn't work with CPython 3.11.

nightscape:
Looks like there are more puzzle pieces missing...
https://github.com/python/cpython/pull/23523
Good call in not merging to `master` 😉 
It all seemed too easy 😅

AutonomicPerfectionist:
@nightscape don't know if you are aware but it seems the recommended procedure for compiling Python on Mac OS is to compile it to a universal2 build, which is a fat build supporting both x86_64 and arm64 Macs. From googling around it *seems* you can still link to universal2 builds from JNI. Unfortunately I can't test any of this as I don't have a Mac myself, but figured you'd be interested in hearing about it. As for numpy, it *seems* using crossenv version 1.4 works for Python 11 now, it completes a build for Linux arm without any errors but I have no idea if it works with universal2 builds.

My issue with some info on Linux arm64 builds:

#1381



Issue with some info on compiling Python to universal2:

https://bugs.python.org/issue46749

PR #1374, opened 2023-06-13
Numpy arm64

nightscape:


saudet:
It's not building for the arm64 architecture. You'll need to fix that

nightscape:
@saudet afaiu, this PR depends on the [cpython-arm64 PR](https://github.com/bytedeco/javacpp-presets/pull/1373) getting merged and artifacts being published, right?
At least right now it seems to be failing because it can't resolve the cpython artifacts.

saudet:
That's the one that isn't getting built for arm64

nightscape:
Sorry, I think I'm not understanding something.
Have you had a look at https://github.com/bytedeco/javacpp-presets/pull/1373 ?
I introduced a cpython-arm64 build there and the corresponding Github action says it's building successfully.
Can we continue the discussion on #1373 ?

saudet:
I see, let's work on that in another branch. I've merged what you had in the cpython branch, and it got deployed as snapshots.

saudet:
Good news! We can use crossenv==1.4.0 to cross compile with CPython 3.11. I don't know if it works with Mac, but it works for Linux, see pull #1386, so please give it a try.

saudet:
I've upgraded the builds for NumPy that now use Meson instead of distutils. It should make it easier to cross compile, so if you'd like to give it another shot: https://numpy.org/doc/stable/user/building.html#cross-compilation

nightscape:
@saudet just to understand correctly (sorry if this is a stupid question):
You upgraded to a NumPy version that supports Meson, but (judging from my limited understanding of Meson that it would not use all these `..._PATH` variables) the corresponding `numpy/cppbuild.sh` still uses the old build system, right?
Would I adapt the `numpy/cppbuild.sh` to use Meson for all platforms as described [here](https://github.com/numpy/numpy/blob/221110fa3a39dffcf297bd8b1ea017d259e410b7/building_with_meson.md)?

nightscape:
I fear I'm in way over my head here...
I'm trying to wrap my head around how all the Python, C and C++ build tools interact with each other, but I'm a spoiled little JVM child and don't seem to be able to understand all that stuff with LIB_PATHs, compiler flags, cross-compilation, ...

saudet:
Meson probably doesn't need those paths, you can probably ignore them.

saudet:
Looks like GitHub Actions now has free M1 runners. Want to give this a try again with that? https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

saudet:
Duplicate of #1515

Issue #1381, opened 2023-07-07
CPython 3.11.3-1.5.9 No linux-armhf/linux-arm64 builds


AutonomicPerfectionist:
Hello, I've noticed that there are no builds in maven for the cpython preset for the linux-armhf, linux-arm64, and linux-ppc64le platforms while they exist for cpython-platform 3.10.8-1.5.8.

I've looked through the commit history, and I found those platforms were commented out in this commit: 57e3fb9

I'm just curious, was there a particular reason behind this change? I can't seem to find any PRs or discussions linked to the above commit, so I'm lost on how to rectify any build incompatibilities that may have warranted the change.

saudet:
See discussion in pull #1373 and #1374

AutonomicPerfectionist:
@saudet thanks for linking those issues. I've looked through them and everything else I could find related but as far as I can tell, the primary blockers are Mac specific. I did see your comment about crossenv not working in python 3.11 but according to [this](https://github.com/benfogle/crossenv/pull/104) issue, it should work now.

I pulled this repo and attempted cross compiling for arm64 Linux myself, it made it all the way through compiling Python before it blew up on the JNI compilation for some reason, an error about skipping the cross compiled .so because of incompatible architecture type. Probably just a misconfiguration on my end, I verified the cross compiled Python was a correctly formated .so for aarch64 but have not tested it on real/virtual hardware.

AutonomicPerfectionist:
Verification of successful cross-compilation:
```bash
$ file cpython/cppbuild/linux-arm64/bin/python3.11
cpython/cppbuild/linux-arm64/bin/python3.11: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, BuildID[sha1]=1e3b651d10f33551e1dd6a326cce7c74c06d4f70, for GNU/Linux 3.7.0, stripped

```

Please let me know if there's something amiss in the compiled binary, such as the interpreter location

AutonomicPerfectionist:
I did manage to compile a full `cpython-linux-arm64.jar`, for whatever reason the JavaCPP-compiler plugin was using my system `g++` even after I set all the compiler properties, so I just symlinked the cross compiler g++ there instead and it all worked just fine. I'll have someone else on our project test the jar on a raspberry pi to see if it works

saudet:
The problem isn't CPython, it's NumPy

[...]

PR #1386, opened 2023-07-10
Re-enable linux ARM builds for CPython and numpy

AutonomicPerfectionist:
This PR fixes an issue where numpy could not be cross-compiled to linux-armhf and linux-arm64. The issue was an incompatibility between `crossenv`, a cross-compiling virtual environment package, and Python 11. A fix was merged to crossenv in benfogle/crossenv#104 and released as crossenv version 1.4. This PR updates the crossenv version to 1.4, allowing cross-compilation again. Thus, it also re-enables builds for linux-armhf and linux-arm64 for both numpy and its dependency cpython.

Theoretically, this should also fix cross-compilation for ppc64le, but I have no way of testing that platform, so I have opted not to re-enable builds for it.

**Why is this a draft?**
On my local system, cpython required additional changes to the cppbuild.sh script to successfully cross-compile. I would like to verify that those changes are not just related to my local configuration before I push them.

Additionally, I spent a while researching Python cross-compilation, and I would like to compile what I've learned into a `CROSS-COMPILATION.md` information guide for anyone who stumbles into the same issues in the future.

Finally, I need to verify that cross-compiled builds don't have some fundamental problem with them even after compiling successfully. That will be verified on an armhf 32 bit raspberry pi by another member of my group.

**Future Maintenance**
I am willing to maintain the cpython linux-armhf and linux-arm64 builds as long as I can. If there are any future issues with them, please ping me in a new issue here on Github and I will respond as soon as possible. If the issue is serious enough or inhibits builds in other parts of the project, these arm builds may be disabled once more until I can fix them, if possible.

I *do not* intend to be the primary maintainer of numpy arm builds. Unfortunately, I have no use case for them and no way to test them myself. If they break and there is no one else to fix them, it is likely they will be disabled again.

**Additional Information**
See #1381 for more information.


saudet:
Thanks!! Do you need linux-armhf?? If not, please leave it disabled. If linux-ppc64le builds though, let's enable it. And let's also see if scipy works with that too?

AutonomicPerfectionist:
Yes, we do require armhf. There are a wide variety of single board computers still out there that only support armhf. For example, the raspberry pi 2, and any raspberry pi 3 or 4 still running the older 32 bit OS.

As for ppc64el and scipy, I can enable those and attempt to test cross compilation on my local machine, if I can get them working I'll push here

AutonomicPerfectionist:
I checked the failing workflows for cpython, that ipv6 error was one I had on my local system. I'll push the fix for it in a bit

saudet:
> Yes, we do require armhf. There are a wide variety of single board computers still out there that only support armhf. For example, the raspberry pi 2, and any raspberry pi 3 or 4 still running the older 32 bit OS.

Right, but like you say they are old boards. What's wrong with using CPython 3.10?

AutonomicPerfectionist:
> What's wrong with using python 3.10?

That's what we're doing right now. We depend on many other Javacpp bindings and require the latest versions where possible. On the Python side, we will be using bleeding edge and experimental packages. So being locked into a particular version of the cpython binding is concerning for us. If we are locked to 3.10 bindings and an incompatibility crops up, we would be forced to either be left behind or leave a significant portion of our users behind. While it might seem odd, we do have a very large number of older armhf boards in active use by our community, including the project leader, so abandoning them is just not an option.

If armhf is a deal breaker, I understand, but our project would have to consider other options.

AutonomicPerfectionist:
ipv6 fix should be pushed. There were two other issues I experienced in my local build, requiring the following additional configure options:

- ac_cv_file__dev_ptmx=yes
- ac_cv_file__dev_ptc=no

I have not looked too closely into what these do, so I'm hesitant to push them until I know for certain the workflow builds fail without them.

I'm checking scipy armhf now; it's getting late here, so I'm going to have to check arm64 scipy and ppc64le cpython/numpy/scipy tomorrow

saudet:
Well, you're the first user in years to mention the need for 32-bit builds, and since I'm not getting paid to do this, and no one is helping me, I have to cut somewhere, and that seems like a good place to start. You're going to need to get rid of your 32-bit boards eventually in a year or two anyway, so you know why not start the transition now before you're cornered by the market? If you have resources to spare though, that's a different question. Send them over here to help!

AutonomicPerfectionist:
It's not up to us on whether or not to get rid of older boards; our project is a hobbyist robotics framework designed to be easy-to-use and flexible. Our users include schools, clubs, and individual tinkerers that may not be able to find or afford newer boards. We're not afraid of breaking backwards compatibility so long as there is an easy migration path, but eliminating support for armhf has no easy migration path beyond forcing our users to junk their perfectly working hardware. We will eventually remove support when the community itself has decided it's time, as we have done for 32 bit x86, but that time is not now.

As I stated in the PR description, I am willing to maintain the armhf and arm64 cpython builds myself. If there is additional cost to supporting armhf beyond that please let me know and I'll see if it's something I can help with.

saudet:
The next release is probably going to be next year and enabling builds that I'm probably going to disable before that isn't something I'll be doing, sorry. You can easily publish builds from your own fork though. Maven Central Repository is a hassle, but with GitHub Packages, we can use a GitHub account and there doesn't seem to be any restrictions about the groupId we can use. Could you check that out?

https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-maven#publishing-packages-to-github-packages

AutonomicPerfectionist:
I don't think maintaining a separate fork is within our project scope, helping to maintain arm support in upstream was all the resource bandwidth we had available. We will stick with 3.10 for now and evaluate other options. I will remove armhf soon

saudet:
Well you asked me how you could help, and I think that making that "forking" process easy is one way you could help. I get a lot of requests for custom builds, like you want linux-armhf, but just for CPython and nothing else, someone else wants Alpine, but just for OpenCV and nothing else, headless builds for FFmpeg, but nothing else, etc, it never ends! But no one wants to pay. Making the process easy for users to create and publish their own builds seems like the way to go here. You don't want to help with this?

AutonomicPerfectionist:
I've disabled armhf and added the two fixes mentioned above. I attempted to get scipy to compile, but it has other issues. Spitting out errors that numpy isn't installed, it's a check done in their setup.py, the reason it errors out is because when cross-compiling, numpy is a different architecture than the host and it blows up trying to import it

AutonomicPerfectionist:
> Making the process easy for users to create and publish their own builds seems like the way to go here

Do you have something in mind for that? The only real way I could see that being accomplished is to have examples for the Github Packages thing you mentioned above and then require users to fork the repo, make their changes to configurations, and publish to Github Packages. Which is fine and all, but I would be fearful of those downstream forks getting angry when an upstream change eventually breaks their configuration.

agibsonccc:
I would be willing to publish bindings I use. At least cpython, cuda, openblas, maybe hdf5. I get requests once in a while for fixing upstream bugs but have the same problem Sam does.  @AutonomicPerfectionist  if your problem is only publishing we could probably do a different namespace with the changes and publish a bit more frequently while backporting to the presets for when sam publishes.

AutonomicPerfectionist:
Forgot to remove armhf from the redeploy jobs.

>  @AutonomicPerfectionist if your problem is only publishing we could probably do a different namespace with the changes and publish a bit more frequently while backporting to the presets for when sam publishes.

I might be interested, but I'm not sure I fully understand what you mean. Are you saying make a fork with the needed changes, publishing them on a faster cadence, and then try to upstream the changes when the next release is a bit closer?
If so, I could see potential issues with that if multiple forks attempt to upstream conflicting changes. If not, please feel free to correct me on my interpretation.

AutonomicPerfectionist:
It seems there was an untimely breakage in Github Actions arm64 package installation:

https://github.com/actions/runner-images/issues/7902

I saw the same issue in #1384

AutonomicPerfectionist:
> @AutonomicPerfectionist yeah I wouldn't really want to "fork" as much as "add the necessary changes and release more often" the goal would be to stay compatible but share the work with Sam to avoid burden on him and releasing when we want. I would prefer to properly push this to maven central. I can do that under my namespace.

I might be open to that, what do you think of it @saudet?

saudet:
So, I've deployed for now release builds for linux-armhf and linux-arm64 against the last release with your fixes:
https://repo1.maven.org/maven2/org/bytedeco/cpython/3.11.3-1.5.9/
I think that should satisfy your needs for now? Thanks for the fix, but next time please open a pull request *before* the release, not *after*. I received no feedback whatsoever about those missing builds from the snapshots before the release, so of course they will not magically appear after the release. The plan is to drop 32-bit builds from the next release, so let's see if you still need them next year.

That said, the next release is probably going to happen after CPython 3.12 gets released, and cross-compilation might break again at that point, so let's revisit all this once it gets release. For now, let's merge this PR in the cpython branch, alongside builds for macosx-arm64 from @nightscape, and enable linux-armhf if you wish. Then, if cross-compilation breaks with CPython 3.12 and there is no one to fix it at that time, just like there was no one to fix it when CPython 3.11 came out, I won't have to do anything but leave the branch to rot. Sounds good?

As for builds outside of the release cycle here, you can do whatever you want on your fork, and as long as it doesn't break any existing builds, we can merge fixes of your own here of course, even for builds that do not happen here. You can use Maven Central Repository, but like I said the easiest thing to do is probably go publish to your own personal repository from GitHub Packages since it works with your GitHub account and you won't have to fiddle with different package names.

AutonomicPerfectionist:
> So, I've deployed for now release builds for linux-armhf and linux-arm64 against the last release with your fixes:
https://repo1.maven.org/maven2/org/bytedeco/cpython/3.11.3-1.5.9/

Oh, thank you! Yes, that should satisfy our current needs.

saudet:
@AutonomicPerfectionist I've upgraded the presets to CPython 3.12. I will merge anything that you have that works for CPython 3.12, not CPython 3.11. However, if you do not provide an update to support ARM now, *before* the release, please do not expect them to be available *after* the release.

AutonomicPerfectionist:
@saudet thanks for the heads up! I've merged master into this branch but won't be able to test it until later. When is the expected release date?

saudet:
> @saudet thanks for the heads up! I've merged master into this branch but won't be able to test it until later. When is the expected release date?

In a couple of months probably

> Edit: Apologies, I did not see you had pushed updates yourself, let me know if you want me to revert that merge commit

I don't think I pushed anything on your branch, I had just changed the target branch, and I've just changed it back to master, but it does look broken now. If you want to continue on this pull request, you'll probably want to rebase onto master

AutonomicPerfectionist:
> I don't think I pushed anything on your branch, I had just changed the target branch, and I've just changed it back to master, but it does look broken now

I see what happened, Github put the commits you made to master above my comment since they're older so I misinterpreted that as being made to my branch before my comment. I did check out the workflows and the only failing ones so far appear to either be Mac-related or a failure to update Ubuntu packages. I'll check on it over the weekend to see if there's anything that needs to be changed in regards to that

saudet:
Please revert changes related to Mac, yes.

AutonomicPerfectionist:
I've rebased on master so we don't have to deal with the several merge commits I had done over the course of this PR. There is no longer anything Mac-related in here, but as a consequence, the commit IDs and timestamps have been reset. I'll be pushing the previous history of this PR to a separate branch of mine for archival reasons only

saudet:
Sounds good, but the build for NumPy fails with:
>      ../../meson.build:1:0: ERROR: Could not invoke sanity test executable: [Errno 8] Exec format error: '/home/runner/work/javacpp-presets/javacpp-presets/numpy/cppbuild/linux-arm64/numpy-1.26.1/.mesonpy-o0l2tw5i/build/meson-private/sanitycheckc.exe'.

AutonomicPerfectionist:
I've switched the numpy build to use `python -m build` instead of pip install, which lets us pass additional cross-compilation options to Meson. It certainly wasn't easy, but I got it to the point of attempting `cythonize`, which then fails because apparently it doesn't support Python 3.12 yet:

https://github.com/cython/cython/pull/5754

saudet:
From what I understand, it should work pretty much transparently with crossenv:
https://numpy.org/doc/stable/user/building.html#cross-compilation

AutonomicPerfectionist:
Unfortunately that was only the case before numpy switched their build system from the standard `setuptools` to `meson-python`, which doesn't seem to support cross compilation in the same way that setuptools did under crossenv. Looking into it further I found the following comment, which shows we can use the cross file with pip, which may help with the current issue and would let me revert the majority of the changes

https://github.com/mesonbuild/meson-python/issues/321#issuecomment-1444319881

AutonomicPerfectionist:
Diving ever deeper, it seems you can re-enable the older setuptools-based build system for now. I found this by looking at what Termux does:

https://github.com/termux/termux-packages/blob/9d6cd88d8385bc71c20493e85ba30f692ae203e5/packages/python-numpy/build.sh#L17-L22

Seems they left the old pyproject TOML in the repo and you can swap to it by renaming it. Testing this on my local machine resulted in a successful `pip install` in the crossenv, I'll revert my earlier changes and use this instead

AutonomicPerfectionist:
@saudet are there any outstanding issues with this PR? I believe it should be functional as is but I have not tested the resulting artifacts on arm hardware yet.

Also, out of curiosity, what's the position of this project on RISC-V? I'm considering getting some hardware for myself to help port some of our dependencies so we're ready when it starts being used en masse. Just wondering if additional support PRs would be welcome for a later release (probably won't have hardware in time to meet this next release)

saudet:
> @saudet are there any outstanding issues with this PR? I believe it should be functional as is but I have not tested the resulting artifacts on arm hardware yet.

distutils has been removed from CPython 3.12. It looks like the NumPy team has hacked it to make it work with CPython 3.12 and added to NumPy instead, but it doesn't work with dependencies like SciPy, so please make it work with distutils. This package is dead, I'm not going to deal with it.

> Also, out of curiosity, what's the position of this project on RISC-V? I'm considering getting some hardware for myself to help port some of our dependencies so we're ready when it starts being used en masse. Just wondering if additional support PRs would be welcome for a later release (probably won't have hardware in time to meet this next release)

If you can figure out how to make that cross compile too, sure, that sounds fine.

saudet:
The builds are still succeeding, so let's merge this for now for the release and I'll just disable it again when it breaks I guess. Merging this might give you more incentive to work on a long-term solution :)

saudet (approved):

saudet:
@AutonomicPerfectionist FYI, the build using distutils and setuptools is gone from NumPy 2.0.0

AutonomicPerfectionist:
@saudet thanks for letting me know! Sorry I didn't respond to your previous message, Github failed to notify me about that one... I'll work on fixing the builds as soon as possible, however I am in the middle of extensive and devastating flooding, so I'm unsure of when numpy can be fixed.

saudet:
Got it, take care!

saudet:
@AutonomicPerfectionist FYI, we should soon have ARM runners for GitHub, so we won't need to hack things for cross-compilation anymore: https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/

AutonomicPerfectionist:
That would be great! I looked into it a bit more and found that originally they were planned to be available for open source projects end of last year, but it was moved to Q1 '25 at some point. So hopefully they'll be available soon.

https://github.com/github/roadmap/issues/970

saudet:
Looks like they became available the next day: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

AutonomicPerfectionist:
Huh, what timing. I'll take a look and see about changing the build to use those runners

AutonomicPerfectionist:
Reverting the cross-compilation workaround and making a minor change to cpython (compiling openssl on arm failed because the architecture `arm64` was unknown, needed to replace it with aarch64) resulted in a successful build of both cpython and numpy on a native arm machine (raspberry pi 4). I'll push those changes to a new branch on my fork soon and make a PR, and I'll change the workflows to use the new runner as well

PR #1392, opened 2023-07-25
[Pytorch] Add an adapter for c10::string_view

HGuillemet:
See https://github.com/bytedeco/javacpp-presets/discussions/1390

I didn't use the owner field of the adapter. I don't think there is a case where we want to free something when the adapter is deleted. Is there ? string_view is supposed to be not-owning view of some string.


HGuillemet:
I mapped `string_view` with:
```Java
        new Info("c10::basic_string_view<char>", "c10::string_view").annotations("@StringView").valueTypes("String", "BytePointer))
```

While for `std::string` it's, by default:
```Java
        new Info("std::string").annotations("@StdString").valueTypes("BytePointer", "String").pointerTypes("BytePointer"))
```
The difference in order of valueTypes results in functions returning `string_view` return `String` in Java and functions returning `std::string` return `BytePointer` in Java. We'd better unify this.

Do you remember why it's this way for `std::string` and shall we do the same for `string_view` ?
Maybe this is to allow application of different encoding ?

saudet:
BytePointer is more general, yes.

sbrunk:
> @sbrunk, could you give that a try ?

Did a quick test with the original einsum example after building locally. Looking good now :)

```scala
println(torch.einsum("ii", torch.randn(Seq(4, 4)))) // -2.5606
```

saudet:
@sbrunk Please let us know if this sufficient for all your use cases! Thanks

sbrunk:
I have to admit I haven't done much testing with other functions using `c10::string_view` arguments beyond `einsum`.

This is partly because it's much easier to test with the prebuilt CI snapshots. For instance, right now I'm traveling without access to my Linux build machine and I can't get the preset to build locally on macOS.

So if you merge this now, I can't promise that I won't find any issue once I've improved test coverage. I don't want to cause you any additional overhead though, so if you'd rather like to wait until I've done more comprehensive testing, I'll do it as soon as I'm back/able to build locally again.

saudet:
Ok, in any case, the build fails on Windows, so that needs to be fixed before we can merge this:
```
2023-08-17T15:12:19.7007842Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project pytorch: Compilation failure: Compilation failure: 
2023-08-17T15:12:19.7008718Z [ERROR] /D:/a/javacpp-presets/javacpp-presets/pytorch/src/gen/java/org/bytedeco/pytorch/global/torch.java:[25536,119] cannot find symbol
2023-08-17T15:12:19.7009300Z [ERROR]   symbol:   class StringViewOptional
2023-08-17T15:12:19.7015054Z [ERROR]   location: class org.bytedeco.pytorch.global.torch
```

HGuillemet:
The checks should pass now. It passed on linux because the parser is run, and not on windows. Is that on purpose ?

saudet (approved):

PR #1412, opened 2023-09-11
Update cppbuild to ensure container has cmake and rapidjson dependencies

jbkyang-nvi:


saudet:
Since this modifies the system, please add that to the workflow instead

jbkyang-nvi:
@saudet This is only for when people want to use javacpp directly within the Tritonserver. Currently they need to do these steps before they do:
```
git clone https://github.com/bytedeco/javacpp-presets.git
cd javacpp-presets
mvn clean install --projects .,tritonserver
mvn clean install -f platform --projects ../tritonserver/platform -Djavacpp.platform.host
```
With the current changes the system installations are done with `cppbuild.sh`

saudet:
Then that's problem with the Docker image of Triton. That should probably be added to this file maybe?
https://github.com/triton-inference-server/server/blob/main/Dockerfile.sdk

jbkyang-nvi:
The developer_tools related bindings are not installed in the Triton container or the Triton SDK container by choice. I don't think it makes sense to install more dependencies in anticipation the the user will use CAPI bindings or Java CAPI bindings

saudet:
I suppose, but you could offer other container images that do have the required build dependencies. Anyway, most Java developers don't build from source, so this isn't something we need. Just offer prebuilt binaries and they will use those.

jbkyang-nvi:
Yep we provide the jars anyway and they can use the bytedeco pre-built binaries. Just wanted a complete set of build instructions so closing this PR and instead submitting #1413 for review 🙏

PR #1413, opened 2023-09-13
Update readme to include installation of additional dependenceis

jbkyang-nvi:
Update README as developer_tools CAPI bindings require more dependencies to install.


saudet:
These dependencies are NOT required for installation. Java developers using prebuilt binaries do not need those. They are only required for the build, but sure we can add information here that's fine.

jbkyang-nvi:
@saudet how would you rephrase this README? My build following the previous instruction fails.

saudet:
I think the changes to the README are fine, we don't need to rephrase anything. I was referring to your comment above only.

saudet (approved):

jbkyang-nvi:
Thanks!

Issue #1321, opened 2023-02-09
[pytorch] Can't get byte data from torch.ScalarType.Byte tensor


jxtps:
I'm using fairly large images and want to avoid the overhead of using floats. However, when I try to read byte data from a `torch.ScalarType.Byte` tensor I get a weird exception. 

Minimal code example to reproduce: 
```java
package misc;

import org.bytedeco.javacpp.BytePointer;
import org.bytedeco.javacpp.FloatPointer;
import org.bytedeco.pytorch.Tensor;
import org.bytedeco.pytorch.global.torch;

import java.nio.FloatBuffer;

public class PytorchUint8TestBench {
    public static void main(String[] args) {
        long[] shape = new long[]{1, 3, 20, 20};
        float[] floats = new float[3 * 20 * 20];
        FloatPointer p = new FloatPointer(FloatBuffer.wrap(floats));
        Tensor floatT = org.bytedeco.pytorch.global.torch.from_blob(p, shape);
        Tensor byteT = floatT.to(torch.ScalarType.Byte);
        System.out.println("byteT.dtype().isScalarType(torch.ScalarType.Byte): " + byteT.dtype().isScalarType(torch.ScalarType.Byte));
        BytePointer byteP = byteT.data_ptr_byte(); // <-- Exception here
        byte[] bytes = new byte[3 * 20 * 20];
        byteP.get(bytes);
    }
}
```
This produces:  
```
byteT.dtype().isScalarType(torch.ScalarType.Byte): true
Exception in thread "main" java.lang.RuntimeException: expected scalar type Char but found Byte
Exception raised from data_ptr at C:\build\aten\src\ATen\core\TensorMethods.cpp:18 (most recent call first):
00007FFAAF449CD200007FFAAF449C70 c10.dll!c10::Error::Error [<unknown file> @ <unknown line number>]
00007FFAAF44975E00007FFAAF449710 c10.dll!c10::detail::torchCheckFail [<unknown file> @ <unknown line number>]
00007FF8804BB95D00007FF8804BB8E0 torch_cpu.dll!at::TensorBase::data_ptr<signed char> [<unknown file> @ <unknown line number>]
00007FFA6E330B5700007FFA6E330AE0 jnitorch.dll!Java_org_bytedeco_pytorch_TensorBase_data_1ptr_1byte [<unknown file> @ <unknown line number>]
0000000003A217B0 <unknown symbol address> !<unknown symbol> [<unknown file> @ <unknown line number>]

[...]

PR #1426, opened 2023-10-16
[PyTorch] Update to 2.1

HGuillemet:
Included in this PR:
* Update to PyTorch 2.1
* Rationalization of function pointer classes. Group all of them in package `functions`
* Add `Tensor.item_bool` and `Tensor.item_byte`
* Add `Tensor.data_ptr_byte` and `Tensor.data_ptr_bool`
* Remove useless classes not part of the libtorch API
* Add `CUDACachingAllocator`
* Map most of missing data loader API, for `Example<Tensor, Tensor>` and `Example<Tensor,NoTarget>`
* Make all methods taking `ArrayRef` arguments of primitive types accept primitive Java array or variadic.
* Make `register_module` generic and its return type the class of the registered module.
* Other minor improvements


saudet:
@sbrunk Could you review this pull request?

HGuillemet:
I still plan to add another improvement: support for stateless datasets, dataloaders, etc... 
Hopefully coming later today or tomorrow. 
Please hold on before starting a review.

sbrunk:
> @saudet: shouldn't we embed libtorch binaries now that it links with CUDA 12 or do we continue building our version with limited compute capabilities ?

If the upstream libtorch binaries don't cause any issues, I'm in favor of switching. I think it has a number of advantages:
1. Less native building -> less strain on limited CI resources
2. It can reduce friction regarding PTX JIT compilation (s. https://github.com/bytedeco/javacpp-presets/pull/1360)
3. Somewhat relaxed driver version constraints due to (currently) targeting CUDA 12.1 instead of always the latest CUDA version.

I will do some tests like running the test suite of Storch using this branch and the upstream binaries to see if I run into any issues.

One thing that might be interesting to look into is MPS acceleration on MacOS, as soon most macs will run on ARM with GPU support.
I'm not sure if the upstream binaries have MPS acceleration enabled, I know the Python wheels do. I'm not even sure if they provide ARM builds at all for pure libtorch (and I don't have an M1/M2 machine to test). I think right now, we don't have support for ARM builds either right?

HGuillemet:
The libtorch binary for mac that can be downloaded from PyTorch main page is still x86_64 only.
I think we could build from source for Mac and use libtorch binaries for CUDA (and ROCm).

saudet:
There's no need to switch to anything, the presets already support LibTorch:
https://github.com/bytedeco/javacpp-presets/tree/master/pytorch#documentation

HGuillemet:
RIght, but what exactly does the user gain in using our binary compared to libtorch, now that libtorch links with CUDA 12 ? If nothing significant, it seems logical to embed libtorch which supports more hardware and for the reason listed by @sbrunk.

One problem I can think of is if the cuda presets is updated to a version and libtorch is not available with this cuda version, or the other way around.
Anything else ?

saudet:
Like I mentioned previously many times, LibTorch links statically with MKL, so it can't be used together with other libraries that use BLAS and LAPACK like GSL, OpenCV, NumPy, SciPy, Smile, etc

HGuillemet:
Ok. Maybe can we use libtorch_cuda only from provided binaries and compile libtorch_cpu ?

I'm done with what I planned to add to this PR. 
You can start reviewing. Comments or RFE welcomed.

HGuillemet:
The remaining error during build on Windows is:
```
D:\a\javacpp-presets\javacpp-presets\pytorch\target\native\org\bytedeco\pytorch\windows-x86_64\jnitorch.cpp(98904): error C2526: 'JavaCPP_org_bytedeco_pytorch_functions_GatheredContextSupplier_allocate_callback': C linkage function cannot return C++ class 'std::shared_ptr<c10::GatheredContext>'
```
This is related to issue bytedeco/javacpp#720.
@saudet Any idea if this can be fixed ? With some `@Convention` maybe ?

Else we can skip the single function needing this function pointer: `CUDAAllocator.recordHistory`.

saudet:
> The remaining error during build on Windows is:
> 
> ```
> D:\a\javacpp-presets\javacpp-presets\pytorch\target\native\org\bytedeco\pytorch\windows-x86_64\jnitorch.cpp(98904): error C2526: 'JavaCPP_org_bytedeco_pytorch_functions_GatheredContextSupplier_allocate_callback': C linkage function cannot return C++ class 'std::shared_ptr<c10::GatheredContext>'
> ```
> 
> This is related to issue [bytedeco/javacpp#720](https://github.com/bytedeco/javacpp/issues/720).
> @saudet Any idea if this can be fixed ? With some `@Convention` maybe ?

Sounds like a bug in MSVC 2019: https://stackoverflow.com/a/57429504/523744

> Else we can skip the single function needing this function pointer: `CUDAAllocator.recordHistory`.

If it doesn't look like an important function, we can do that, sure.

saudet:
> Ok. Maybe can we use libtorch_cuda only from provided binaries and compile libtorch_cpu ?

Where can we get libtorch_cuda for CUDA 12.3?

HGuillemet:
According to https://developer.nvidia.com/blog/cuda-toolkit-12-0-released-for-general-availability/ we could run with 12.3 a binary compiled for 12.1 ?

saudet:
CUDA doesn't maintain backward compatibility of the ABI, even for minor
versions.

HGuillemet:
Have you seen the "compatibility" paragraph on the link above ? It seems to be new since CUDA 11.

saudet:
Well, try it and tell me if it works! I don't think it does. NVIDIA says many things, but reality is often different.

sbrunk:
I'm trying to test this locally, but I'm running into issues.
```
[ERROR] Failed to execute JavaCPP Builder: Could not parse "c10/cuda/impl/cuda_cmake_macros.h": File does not exist
```
Fresh pytorch clone from the cppbuild. There's a `cuda_cmake_macros.h.in`. Any idea what I might be missing?

HGuillemet:
Have you run the cppbuid ?

sbrunk:
> Have you run the cppbuid ?

Yes sorry should have mentioned that. I'm running `mvn install --projects pytorch -Djavacpp.platform=linux-x86_64 -Djavacpp.platform.extension=-gpu` and the native build is working fine.

sbrunk:
I managed to build it locally now, and the Storch tests are passing (with minimal changes) with 2.1. :)

Testing with CUDA is causing issues though. My current assumption is that we might need to extend the CUDA arch list to support Ampere GPUs. See https://github.com/sbrunk/storch/issues/62

sbrunk:
I did some more testing. All tests were done on an Ampere GPU with compute capability 8.6 Ubuntu 22.04 with the latest Nvidia driver currently available in the official package sources which is 535.104.12 which means CUDA *driver* version 12.2

- [Snapshot](https://oss.sonatype.org/service/local/repositories/snapshots/content/org/bytedeco/pytorch/2.0.1-1.5.10-SNAPSHOT/pytorch-2.0.1-1.5.10-20231025.021128-64-linux-x86_64-gpu.jar) from current master (targeting PyTorch 2.0.1 and CUDA 12.3), with cuda-redist-12.3-8.9 or locally installed CUDA 12.3 -> PTX error
- Locally built libtorch from this branch. (targeting PyTorch 2.1.0 and CUDA 12.3) -> PTX error
  - Added compute capabilities 8.0 and 8.6 to `TORCH_CUDA_ARCH_LIST` -> PTX error
- Switched to CUDA 12.1 for the local build:
  - cuda-redist-12.3-8.9 -> works
  - Locally installed CUDA versions 12.1 and 12.3 -> works
  - Upstream libtorch (libtorch-cxx11-abi-shared-with-deps-2.1.0+cu121) when setting `pathsFirst` and lib path -> works

To make sure that the native libraries are chosen I always checked the symlinks in `$HOME/.javacpp/cache` and deleted the cache folder between runs.

So it looks like building against 12.1 makes it work again, and is compatible with 12.3 at runtime. I still need to verify this for 2.0.1

HGuillemet:
Thanks for reporting all these tests.

> 535.104.12 which means CUDA _driver_ version 12.2

That explains the whole thing.
So they release CUDA 12.3 but the 545 driver that is compatible is still in beta...

Hopefully upstream libtorch_cuda and libcusparse  will work with JavaCPP build of other libs and we could get rid of PTX nightmares.

saudet:
@HGuillemet Please update the CUDA directory paths to 12.3 in these files:
https://github.com/bytedeco/javacpp-presets/blob/master/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
https://github.com/bytedeco/javacpp-presets/blob/master/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java

saudet:
It looks like either the GitHub runners or NVCC from CUDA 12.3 are a bit faster. In any case, we have an extra build your that we could use for 8.0 arch.

sbrunk:
With **both** https://github.com/bytedeco/javacpp-presets/pull/1426/commits/41a97d1c77c841b0f59aaf0bf9fe595adcc1002d and https://github.com/bytedeco/javacpp-presets/pull/1426/commits/a3102e53589e37052dcf28663816ac643ecfe709 it seems to be working with my local build (built with locally installed CUDA 12.3 and linked against `cuda-redist` (also 12.3) at runtime). 🚀

Would be great if someone else could verify this.

saudet:
@sbrunk Apart from that, anything else that needs to be fixed?

HGuillemet:
Building passed !
What makes you confused in my suggestion to use `5.0;6.0;7.0;8.0;9.0` instead of  `5.0;6.0;7.0;8.0+PTX` ?  It will do the same, but happy users of 9.0 will benefit from their specific capabilities, will not have to compile PTX during minutes at first launch, will not have to ensure they have a big enough cache,.. Only difference I see is that the build won't work on next generation, but we will build several new versions of the presets before this generation arrives. Official binaries are compiled with `3.5;3.7;5.0;6.0;7.0;7.5;8.0;8.6;9.0`. 

> Would be great if someone else could verify this.

I'm running with either 12.2 and 12.3 without problem. But on Linux only. I cannot test on Windows.

saudet:
We can give it a try with `5.0;6.0;7.0;8.0;9.0` as well see how that goes

HGuillemet:
> We can give it a try with `5.0;6.0;7.0;8.0;9.0` as well see how that goes

Ok I'm pushing a commit.

sbrunk (approved):
Looks good to me. LGTM.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
```diff
@@ -1852,11 +1959,13 @@ We need either to put an annotation info on each member, or javaName("@NoOffset
...
+               .put(new Info("at::TensorBase::data_ptr<uint8_t>").javaNames("data_ptr_byte"))
                .put(new Info("at::TensorBase::data_ptr<int16_t>").javaNames("data_ptr_short"))
                .put(new Info("at::TensorBase::data_ptr<int>").javaNames("data_ptr_int"))
                .put(new Info("at::TensorBase::data_ptr<int64_t>").javaNames("data_ptr_long"))
                .put(new Info("at::TensorBase::data_ptr<float>").javaNames("data_ptr_float"))
                .put(new Info("at::TensorBase::data_ptr<double>").javaNames("data_ptr_double"))
+               .put(new Info("at::Tensor::item<bool>").javaNames("item_bool"))
                .put(new Info("at::Tensor::item<int8_t>").javaNames("item_char"))
                .put(new Info("at::Tensor::item<int16_t>").javaNames("item_short"))
                .put(new Info("at::Tensor::item<int>").javaNames("item_int"))
```
saudet, line 1971: Let's add data_ptr_bool() and item_byte() for consistency

HGuillemet: You're right `about data_ptr_bool`.
But `item<uint8_t>`, like said [in this thread](https://github.com/bytedeco/javacpp-presets/issues/1321#issuecomment-1768937456), means "get the unique element of this tensor and return it as a `uint8_t`".  Since we don't have `uint8_t` in Java, it doesn't make much sense. If we want an `item_byte` that returns a Java `int`, we would add a `javaNames("item_byte")` to `item<int>`, this would be weird.

saudet: Just add it, it doesn't matter if it's "weird", it does what it's supposed to do!

saudet: BTW,  I don't think item() will allow you to cast a uint8_t to int, so just return a byte that's fine.

HGuillemet: It does. `item<int>` is just a macro to `item().toInt()` where `item()` returns a `Scalar`.

saudet:
> 7,812 additions, 5,691 deletions not shown because the diff is too large. Please use a local Git client to view these changes. 

A lot of those "changes" is simply because you're changing around the parser order. If there are no good reasons to change the order, please revert the order to how it was before. It makes it hard to see the differences.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
saudet: Right, so for an uint_8 Scalar, toInt() fails, in my experience, so don't do that. Just return a byte, that's fine

HGuillemet:
`torch_include.h` is generated by a script from the output of `g++ -H`. This ensures we don't miss any new includes and that the includes are parsed in the order they are processed by g++.  If the order changed since 2.0.1 this is probably because of upstream changes (or maybe I fixed a bug in my script).

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
HGuillemet: Just tested, and it works. So we can do either, as you wish. In both case for me `item_byte` is meaningless and is just a placeholder to match the `data_ptr` methods.

saudet:
Where is that script?

HGuillemet:
Just added to the repo

saudet:
Ok, so can you please run that against master and make sure *nothing* changes? If something changes, please fix it so that nothing changes.

pytorch/include_list.pl
```diff
@@ -0,0 +1,65 @@
+#!/bin/perl
+
+# Must be run at from javacpp-presets/pytorch after cppbuild.sh has been run 
+# for linux-x86_64-gpu
```
saudet, line 4: Add it to the cppbuild.sh script

HGuillemet: This is not fully automatized, there are exceptions to comment out (not parseable headers...)  and it can only run on linux since we need g++.

saudet: If it's not fully automated, then it means you are manually creating the inlude files. So please fix the include files to preserve the existing order. Either that, or make it automated!

HGuillemet: Let's make it fully automated then. It will require less work when  a new pytoch version is released and we won't miss new API. 
I'll hard-code the list of exceptions related to non-parseable headers in the script, and move the lists to `src/gen/resources`

saudet: Look, you're doing all this very ad hocly, which is fine, but like I keep telling, this is equivalent to doing everything manually, so please revert any unnecessary changes! If something doesn't need to be changed, then don't change it. If doesn't matter if you're half-automated tool does half the work, because you're modifying the output of it, so modify it more so the order doesn't change. If you're able to come up with a fully automated method, that's different story. But you're not able to do it fully automated, so until you're able to do that, please revert all and every change that is not necessary. You're editing these files manually already, what's the problem?

saudet: BTW, why do you say this needs GCC? Is this version of Clang not working for some reason?
http://bytedeco.org/javacpp-presets/llvm/apidocs/org/bytedeco/llvm/program/clang.html

HGuillemet: > what's the problem?

The problem is that I just spent 1h30 swapping files into an unnatural order, possible introducing bugs, for no good reason. And that I'll have to redo the same thing each new release.
And no, I don't feel like looking to redo the same script using clang, while pytorch is built using g++ and the include list extraction works with g++. When I said I need g++, I meant that the script won't work on Windows.

pytorch/src/main/resources/org/bytedeco/pytorch/presets/torch_include.h
```diff
@@ -1398,18 +1410,15 @@
...
+#include "torch/csrc/jit/serialization/pickle.h"
 
-// Parsed and for inclusion in JNI
-// See also https://github.com/pytorch/pytorch/blob/main/docs/cpp/source/Doxyfile
-// for an approximation of what should be in API in addition to torch.h"
-// torch/csrc/jit/runtime/custom_operator.h:  Name conflict with torch::RegisterOperator + little chance to have any use
-#include "torch/torch.h"
-#include "ATen/native/TensorShape.h"
-#include "torch/csrc/jit/serialization/storage_context.h"
-#include "torch/csrc/jit/serialization/import.h"
```
saudet, line 1415: What's the reason we don't need those anymore?

HGuillemet: `torch.h` is the main root include that draws others. It only contains includes. We can add it for good measure, but this won't change anything.

It's unclear what exactly is part of the "public" C++ API besides `torch.h`. In 2.0.1 I mainly considered what is in the doxygen files that generates the online api doc. But then I realized that the normal, documented, way to access the TorchScript functionality is to include `script.h`. So in 2.1 what is parsed is what is drawn by `torch.h` + `script.h`. That's closer to how the lib is to be used.
`import.h` is drawn by `script.h` and is parsed. `TensorShape.h` and `storage_context.h` are not. I guess they are not meant to be accessed by users.

pytorch/README.md
```diff
@@ -93,9 +93,9 @@ public class SimpleMNIST {
     static class Net extends Module {
         Net() {
             // Construct and register two Linear submodules.
-            fc1 = register_module("fc1", new LinearImpl(784, 64));
-            fc2 = register_module("fc2", new LinearImpl(64, 32));
-            fc3 = register_module("fc3", new LinearImpl(32, 10));
+            register_module("fc1", fc1 = new LinearImpl(784, 64));
+            register_module("fc2", fc2 = new LinearImpl(64, 32));
+            register_module("fc3", fc3 = new LinearImpl(32, 10));
```
saudet, line 98: What's the reason why you removed those register_module() functions? If there is no particular reason why you removed them, please put them back.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
```diff
@@ -1824,6 +1930,7 @@ We need either to put an annotation info on each member, or javaName("@NoOffset
                .put(new Info("torch::jit::Node").pointerTypes("JitNode"))
                .put(new Info("torch::jit::Module").pointerTypes("JitModule"))
                .put(new Info("torch::jit::Object").pointerTypes("JitObject"))
+               .put(new Info("torch::jit::load").javaNames("jitLoad"))
                .put(new Info("torch::jit::String").pointerTypes("JitString"))
```
saudet, line 1934: Please remove the jit prefixes from all those names. Like I keep telling you, don't change anything that doesn't need to be changed.

pytorch/README.md
HGuillemet: `register_module(String, Module)` was broken in 2.0.0 and earlier since:
```
Module m = new LinearImpl(32,10);
...
register_module("fc", m);
```
caused a segmentation fault.
This is why you added all the `register_module` overrides in the first place.
This was fixed in 2.0.1 with the `upcast` feature. So the overrides and the corresponding template instantiations were not necessary any more.
All what they do is returning a more specific type, but this is not useful in the general case since with a module `m` from a custom java class `JavaModule` extending `Module`,`register_module("x", m)` returns a `Module` and not the `JavaModule`.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
HGuillemet: All those symbols from `torch::jit` module should really be in a separate Java package, since the related features are not needed by most users, but we cannot do this.
`JitModule` and `JitNode` are needed since we already other classes named `Module` and `Node` in the pytorch package.
`JitString` and `JitObject` are for convenience I guess. Anyway all those are not changed in 2.1.0, they were already mapped as such in your versions of the presets.

`jitLoad` is new and is a mistake. I though there were a conflicting `load` method but I cannot find it any more. I'll revert that.

pytorch/README.md
saudet: Right, but that doesn't justify breaking existing code. Put the old overrides back

HGuillemet: They were already removed in 2.0.1. And they alone increase the jni library size by about 4-5%. 
Can you confirm that's what you want before I push this commit ?

saudet: Sure, that's fine by me. We could also add them back as helper Java methods like the ones here:
https://github.com/bytedeco/javacpp-presets/tree/master/opencv/src/main/java/org/bytedeco/opencv/helper

HGuillemet: Good idea. This looks better because it:
* doesn't instantiate tones of templates.
* allows to return the exact same object rather than a new wrapper against a C++ pointer
* will work with Java modules

HGuillemet: Something like:
```Java
static  <M extends Module> M register_module(String name, M module) { 
  _register_module(name, module);
return module;
}
```

saudet: Oh and I think we can even use generics for that purpose

saudet (approved):

PR #1441, opened 2023-11-29
[DepthAI] Add support for Windows & Android

richard-xx:
@saudet RFR, thanks


saudet:
It would be nice to insert the new entries in the same order as with, for example, OpenCV, but it looks good otherwise. Thanks! :+1:

PR #1444, opened 2023-12-04
[PyTorch] Update to 2.1.2

HGuillemet:
Included in this PR:
* Add missing mapping for `std::array<CUDACachingAllocator::Stat>`
* Update to PyTorch 2.1.2


saudet:
Good to merge?

HGuillemet:
I think so, but I wonder:
Shouldn't we add a `preload` property in `torch_cuda` so that people running with the `cuda` presets and not a system-wide cuda installed have the libraries extracted, without explicitly loading the cuda classes ?
Something like:
```
  preload = {"cudart@.12", "cublasLt@.12", "cublas@.12", "cudnn@.8", "nvToolsExt@.1", "cusolver@.11", "cufft@.11", "curand@.10", "cusparse@.12"}
```

saudet:
That's already in presets/torch.java: https://github.com/bytedeco/javacpp-presets/blob/master/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java#L147

HGuillemet:
Ok. I forgot this.
So GtM.

saudet (approved):

sbrunk (approved):

PR #1446, opened 2023-12-11
Add presets for the `timeapi.h` of windows systems

devjeonghwan:


saudet:
It looks like you've mostly changed only the line endings. Please revert that

PR #1455, opened 2023-12-30
[Pytorch, OpenCV] Implicit ArrayRef, Updates for JavaCPP 1.5.10

HGuillemet:
### Pytorch ###
* Have methods taking an `ArrayRef` accept also a `Vector` (relying on C++ converting constructor of `ArrayRef`) when the corresponding vector template instance has been mapped in Java. This allows to write for instance:
  ```Java
  Tensor tensor2 = tensor1.index(new TensorIndexVector(new TensorIndex(x), new TensorIndex(new Slice())));
  ```
  instead of explicitly constructing the `TensorArrayRef` to pass to `index`. This is still much heavier that what we can write in C++:
  ```c++
  Tensor tensor2 = tensor1.index({x, Slice()});
  ```
  but it's better than with explicit use of `ArrayRef`.
* Adjust infos for templated constructors  (PR bytedeco/javacpp#732)
* Adjust infos for template specializations (PR bytedeco/javacpp#733)

### OpenCV ###
* Add `Info.upcast()` on `Algorithm` to prevent crashes when upcasting or downcasting
* Add template instantiations for `cv::read` and `cv::write`


saudet:
Is this pull request done? You're not going to add anything more here for other presets?

HGuillemet:
Done

saudet:
Please don't push 200 times a day, Google Drive really doesn't like it, and I don't know how to fix this

HGuillemet:
I haven't pushed anything for 3 days. Can I help somehow ?

saudet:
Start by not pushing more than once a day the next time. Why do you do that
anyway?

HGuillemet:
According to the log above, there was just 3 pushes the same day, 4 days ago. No particular reason. I just worked this whole day on the PR I guess.
Is the problem due to these 3 pushes a day or that the third one was a forced-push ?

saudet:
I asked you why do you had to push more than once a day, and you basically answered that you don't need to. So, don't push more than once a day! Ok? Whenever you push something, the builds for everything that you touched get triggered. Don't do that. Maybe I should just disable triggers for pull request if that feature exists, but jeez, just don't push more than once a day! Just don't do it

saudet (approved):

saudet:
@HGuillemet The SimpleMNIST sample code doesn't work anymore after merging this. Please fix it
```
java.lang.NullPointerException: Pointer address of argument 0 is NULL.
    at org.bytedeco.pytorch.MNISTBatchDataset.map (Native Method)
    at SimpleMNIST.main (SimpleMNIST.java:82)
    at org.codehaus.mojo.exec.ExecJavaMojo.lambda$execute$0 (ExecJavaMojo.java:283)
    at java.lang.Thread.run (Thread.java:840)
```

HGuillemet:
This is the issue about Stack mentioned in bytedeco/javacpp#733.
To sum it up:
Stack primary template is a declaration without body. 
Template specializations (that have bodies in this case), are now ignored.
So the Java classes mapping template instances are `@Opaque` with no constructor and `new ExampleStack()` used in SimpleMNIST returns an object with a null address.

I'll add an ugly `javaText` to give the text of the whole class and force the generation of a constructor. 
Unless you have a better idea ?

It could be nice to be able in Info to specify an arbitrary javaText to be appended to a class without having to give the text of the whole class or to hook to another existing, unrelated, declaration in the class.

Shall I push the commit to this PR even if it has already been merged ? or open another ?

saudet:
> I'll add an ugly `javaText` to give the text of the whole class and force the generation of a constructor. Unless you have a better idea ?

We can just replace the constructor, no?

> It could be nice to be able in Info to specify an arbitrary javaText to be appended to a class without having to give the text of the whole class or to hook to another existing, unrelated, declaration in the class.

Yeah, I thought I did that somewhere already, but I guess not. It's not obvious how it would work. We'd probably need to add new flags and what not, so unless we figure out a nice way to do it without that, let's not do that

> Shall I push the commit to this PR even if it has already been merged ? or open another ?

Another pull request please, but I'm pretty much done with the release I think, so if no one complained until now I guess no one needs it, so we can do that later I suppose, sounds good?

HGuillemet:
> > I'll add an ugly `javaText` to give the text of the whole class and force the generation of a constructor. Unless you have a better idea ?

Well, in fact I don't know how to replace the whole class text. If I add a javaText to the type, it's copied elsewhere when the type is used, not only in the class definition.

> 
> We can just replace the constructor, no?

The constructor is generated [here](https://github.com/bytedeco/javacpp/blob/e9b82ce41c5da5c5e453d9398d3cef1312f22ec4/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L3780C1-L3785C30).
How would you hook an info to target the constructor ?

> Another pull request please, but I'm pretty much done with the release I think, so if no one complained until now I guess no one needs it, so we can do that later I suppose, sounds good?

As you wish. No idea if ExampleStack and TensorExampleStack is of common use. I don't use the dataset api myself.
See also the JavaCPP issue 740 I posted some minutes ago and if you want to include something about it before the release.

saudet:
No, I'm not going to wait after you fixing things endlessly. You keep breaking things, so if someone complains, please fix it. Otherwise, it's not important I suppose.

PR #1462, opened 2024-01-19
Fix mapping of unique pointers

HGuillemet:
When a native function takes a `unique_ptr` to some class as an argument  AND if that class can be instantiated from Java, the Java constructors must by annotated with `@UniquePtr` and `@Name`. 

Affected presets: Pytorch and Tensorflow-lite.


saudet (approved):
Looks good, but is there anything else you'd like to fix before the release or is this it?

HGuillemet:
I'm done for the presets. I'm going to push another small fix for javacpp

saudet:
Any estimates on when that fix is going to land?

HGuillemet:
Regression tests on presets are currently running.
I'll check the result in ~ 8 hours.

Issue #1069, opened 2021-08-10
Apple Silicon Support


peterrojs:
Since there are going to be more and more Apple Silicon computers in the wild, it would make sense to make a JAR for macosx-aarch64

saudet:
Thanks to W. Bradford Paley and @oxisto, we now have builds for OpenBLAS, OpenCV, and LLVM!

Please give it a try with the snapshots: http://bytedeco.org/builds/

TimKieu:
For FFMPEG, I found this build for Mac M1: [ffmpeg Apple silicon](https://github.com/ssut/ffmpeg-on-apple-silicon).
Please take a look and integrate in the presets.

saudet:
In the case of FFmpeg, we also need to build the dependencies, so that means "porting" all of this:
https://github.com/bytedeco/javacpp-presets/blob/master/ffmpeg/cppbuild.sh#L1322-L1426
It's probably not a huge amount of work, but someone has to do it and contributions are welcome! 
Please open a pull request: https://github.com/bytedeco/javacpp-presets/pulls

saudet:
I've introduced `macosx-arm64` builds for FFmpeg in commit https://github.com/bytedeco/javacpp-presets/commit/e38c9d192505b76227acb8a645a405e986bfc696.
Please give it a try with the snapshots: http://bytedeco.org/builds/

Faz95210:
Hi,

I'm currently experiencing more or less the same issue as the author of the ticket https://github.com/bytedeco/javacv/issues/1696 . 
I'm compiling a project in Kotlin through gradle while using your libs
```
    javacpp_version = '1.5.7-SNAPSHOT'
    ffmpeg_version = "4.4.1-$javacpp_version"

    runtimeOnly "org.bytedeco:ffmpeg:$ffmpeg_version:macosx-x86_64"
    runtimeOnly "org.bytedeco:javacpp:$javacpp_version:macosx-arm64"
```
I also tried with other mixes of version (4.4-1.5.6/ 4.3.1-1.5.4)

The project used to run perfectly on intel Macs with version 4.3.1-1.5.4
But now, on a Mac M1 whatever the versions used I always get the same error
```
├─ org.bytedeco.ffmpeg.global.avutil.<clinit>(avutil.java:14)
├─ org.bytedeco.javacpp.Loader.load(Loader.java:1123)
├─ org.bytedeco.javacpp.Loader.load(Loader.java:1300)
[...]

PR #1463, opened 2024-01-22
Add PyTorch macosx-arm64 cross-compiled build

### LINKED ISSUE #1461: PyTorch macosx-arm64 cross-build
TL;DR
Can we use the x86_64 Python interpreter for cross-compilation of PyTorch to macosx-arm64 or will it cause issues? If we can, what is a good way to configure it especially for CI builds? If not, what might be other options?

## Background

We only have x86_64 machines in CI, so we currently have to cross-compile macosx-arm64 artifacts. With https://github.com/bytedeco/javacpp-presets/issues/1459 out of the way, I got cross-compilation PyTorch working locally for #1069. I can compile arm64 artifacts on x86_64 that are working fine on an arm64 machine. 

What I'm not sure about though is the Python build dependency. One thing it's used for is to call the build:

https://github.com/bytedeco/javacpp-presets/blob/ee1401da0b612ad515f06fa6846937ebb87f2a18/pytorch/cppbuild.sh#L185

Which means for cross-compilation we need an interpreter of the **host** arch (x86_64 here). It is also used for other things though, i.e. and for numpy stuff and in `CFLAGS` here:

https://github.com/bytedeco/javacpp-presets/blob/ee1401da0b612ad515f06fa6846937ebb87f2a18/pytorch/cppbuild.sh#L105-L105

The PyTorch build depends on cpython of its **target** arch though. Locally via 

https://github.com/bytedeco/javacpp-presets/blob/ee1401da0b612ad515f06fa6846937ebb87f2a18/pytorch/cppbuild.sh#L59-L59

For a maven build it seems `BUILD_PATH` is set and therefore `CPYTHON_PATH` is overwritten to use cpython cached from maven:

https://github.com/bytedeco/javacpp-presets/blob/ee1401da0b612ad515f06fa6846937ebb87f2a18/pytorch/cppbuild.sh#L63-L68

I'm not sure how `CPYTHON_PATH` is set exactly (maven plugin?) but I think it is coming in through the  transitive  `maven-compiler-plugin` dependency on cpython via `numpy-platform` in `pom.xml`:

https://github.com/bytedeco/javacpp-presets/blob/ee1401da0b612ad515f06fa6846937ebb87f2a18/pytorch/pom.xml#L52-L56

To get it working locally, I did a little hack: Build a native x86_64 cpython but put it into the `cppbuild/macosx-arm64` directory so `CPYTHON_PATH` is set to it and we can actually run that Python interpreter for kicking off the PyTorch cross-compilation.

So the open question is: Can we use the x86_64 Python interpreter for cross-compilation of PyTorch or will it cause issues? If we can, what is a good way to configure it especially for CI builds? If not, what might be other options?

sbrunk:
Fixes #1461 

Add PyTorch cross-compilation `macosx-x86_64` -> `macosx-arm64`. The idea is to use the host Python interpreter to run `setup.py build` See #1461 for more details about this. Cross-compilation settings are derived from the [PyTorch builder](https://github.com/pytorch/builder/blob/74b04f302afede5c25275d8026f34a06330cc515/wheel/build_wheel.sh#L201-L206).

After some experimentation, I figured out that we can use a Maven profile to to use the host Python interpreter for cross-compilation in CI without affecting the existing native builds.

Since we're building just the C++ part of PyTorch (LibTorch), I think it should be OK to add the host Python interpreter like this. We might need to adapt a little bit once we have CPython and NumPy for arm, if it requires running the host Python but also linking against target libpython but I'm not sure (see #1461).

Perhaps a somewhat cleaner approach could be trying to keep `$CPYTHON_PATH` set to the `$PLATFORM` cpython and get the host interpreter out another way in the build script but that would need some more experimentations.

@saudet what do you think?

BTW I also tried cross-compiling cpython as well before realizing that we need the host interpreter for running `setup.py` but couldn't get it working as it's not really supported currently. See [Support cross compilation on macOS #90905](https://github.com/python/cpython/issues/90905). Perhaps GitHub will provide free native macosx-arm64 runners at some point which should make this easier.

#1069

### DIFF
--- a/pytorch/cppbuild.sh
+++ b/pytorch/cppbuild.sh
@@ -56,11 +56,8 @@ git submodule foreach --recursive 'git reset --hard'
 # https://github.com/pytorch/pytorch/pull/66219
 #patch -Np1 < ../../../pytorch.patch
 
+CPYTHON_HOST_PATH="$INSTALL_PATH/../../../cpython/cppbuild/$PLATFORM/host/"
 CPYTHON_PATH="$INSTALL_PATH/../../../cpython/cppbuild/$PLATFORM/"
-# local cross-compilation requires x86_64 cpython
-if [ "$ARCH" = "x86_64" ] && [ "$PLATFORM" = "macosx-arm64" ]; then
-  CPYTHON_PATH="$INSTALL_PATH/../../../cpython/cppbuild/macosx-x86_64/"
-fi
 OPENBLAS_PATH="$INSTALL_PATH/../../../openblas/cppbuild/$PLATFORM/"
 NUMPY_PATH="$INSTALL_PATH/../../../numpy/cppbuild/$PLATFORM/"
 
@@ -69,7 +66,12 @@ if [[ -n "${BUILD_PATH:-}" ]]; then
     IFS="$BUILD_PATH_SEPARATOR"
     for P in $BUILD_PATH; do
         if [[ $(find "$P" -name Python.h) ]]; then
-            CPYTHON_PATH="$P"
+            if [[ "$(basename $P)" == "$PLATFORM_HOST" ]]; then
+                CPYTHON_HOST_PATH="$P"
+            fi
+            if [[ "$(basename $P)" == "$PLATFORM" ]]; then
+                CPYTHON_PATH="$P"
+            fi
         elif [[ -f "$P/include/openblas_config.h" ]]; then
             OPENBLAS_PATH="$P"
         elif [[ -f "$P/python/numpy/core/include/numpy/numpyconfig.h" ]]; then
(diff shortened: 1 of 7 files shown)


saudet:
I've already been doing something like that for NumPy and SciPy, so could you bring in the host version of CPython the same way as those?
https://github.com/bytedeco/javacpp-presets/blob/master/numpy/cppbuild.sh
https://github.com/bytedeco/javacpp-presets/blob/master/numpy/pom.xml
https://github.com/bytedeco/javacpp-presets/blob/master/scipy/cppbuild.sh
https://github.com/bytedeco/javacpp-presets/blob/master/scipy/pom.xml

pytorch/cppbuild.sh
```diff
@@ -75,36 +81,46 @@ if [[ -n "${BUILD_PATH:-}" ]]; then
...
+    export PYTHONPATH="$PYTHON_INSTALL_PATH:$NUMPY_PATH/python/"
+    mkdir -p "$PYTHON_INSTALL_PATH"
+
+    export CFLAGS="-I$CPYTHON_PATH/include/ -I$PYTHON_LIB_PATH/include/python/ -L$CPYTHON_PATH/lib/ -L$CPYTHON_PATH/libs/"
+    $PYTHON_BIN_PATH -m pip install --target=$PYTHON_LIB_PATH $TOOLS
+else # cross-compile
+    export PYTHON_BIN_PATH="$CPYTHON_HOST_PATH/bin/python3.12"
+    chmod +x $PYTHON_BIN_PATH
+    $PYTHON_BIN_PATH -m pip install --target="$CPYTHON_HOST_PATH/lib/python3.12/" $TOOLS
+fi
```
saudet, line 123: Please remove the `if [[ $PLATFORM == $PLATFORM_HOST ]]; then` and do something like this instead:
```python
if ! $PYTHON_BIN_PATH -m pip install --target=$PYTHON_LIB_PATH $TOOLS; then
    chmod +x "$CPYTHON_HOST_PATH/bin/python3.12"
    export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$CPYTHON_HOST_PATH/lib/:$CPYTHON_HOST_PATH"
    chmod +x $CPYTHON_HOST_PATH/lib/python3.12/bin/*
    export PATH="$CPYTHON_HOST_PATH/lib/python3.12/bin/:$PATH"
    export PYTHON_BIN_PATH="python"
fi
```

pytorch/pom.xml
```diff
@@ -135,4 +137,45 @@
...
+                <groupId>org.bytedeco</groupId>
+                <artifactId>cpython</artifactId>
+                <version>3.12.1-${project.parent.version}</version>
+                <classifier>${os.name}-${os.arch}</classifier>
+              </dependency>
+              <!-- we need to exclude target cpython until we have a macosx-arm64 build -->
+              <dependency>
+                <groupId>org.bytedeco</groupId>
+                <artifactId>numpy-platform</artifactId>
+                <version>1.26.3-${project.parent.version}</version>
```
saudet, line 166: I suppose we need to do something similar in platform/pom.xml:
```
2024-01-24T00:31:14.1394220Z [ERROR] Failed to execute goal on project pytorch-platform: Could not resolve dependencies for project org.bytedeco:pytorch-platform:jar:2.1.2-1.5.10-SNAPSHOT: Could not find artifact org.bytedeco:cpython:jar:macosx-arm64:3.12.1-1.5.10-SNAPSHOT in sonatype-nexus-snapshots (https://oss.sonatype.org/content/repositories/snapshots) -> [Help 1]
2024-01-24T00:31:14.1403910Z org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project pytorch-platform: Could not resolve dependencies for project org.bytedeco:pytorch-platform:jar:2.1.2-1.5.10-SNAPSHOT: Could not find artifact org.bytedeco:cpython:jar:macosx-arm64:3.12.1-1.5.10-SNAPSHOT in sonatype-nexus-snapshots (https://oss.sonatype.org/content/repositories/snapshots)
```

pytorch/cppbuild.sh
sbrunk:
```shell
if ! $PYTHON_BIN_PATH -m pip install --target=$PYTHON_LIB_PATH $TOOLS; then
```
fails (stopping the script) in the cross-compilation setting since `PYTHON_LIB_PATH` is not set at all in that case, that's why I put the whole block in a conditional.

saudet: Right, so instead of all that, let's simply put `CPYTHON_PATH=CPYTHON_HOST_PATH` before 
`if [[ -f "$CPYTHON_PATH/include/python3.12/Python.h" ]]; then`

pytorch/pom.xml
saudet: Actually, I don't think we need to include those optional dependencies at all...

sbrunk:
Ah that's much simpler. Thanks @saudet. I just ran cross-compilation locally from your latest commit [bda4475](https://github.com/bytedeco/javacpp-presets/pull/1463/commits/bda4475e646619a6071ad6f6f37e5ea8b3335b39) and then some smoke tests on an arm64 machine sucessfully. 🎉

saudet (approved):

PR #1466, opened 2024-02-02
[PyTorch] Update to 2.2.1 and minor changes

### LINKED ISSUE #1468: Problems deploying pytorch 2.1.2-1.5.10
I'm running into some problems deploying the (as of this writing latest) version of https://github.com/bytedeco/javacpp-presets/tree/master/pytorch `2.1.2-1.5.10`. 

If I try to use libtorch (https://download.pytorch.org/libtorch/cu121/libtorch-shared-with-deps-2.1.2%2Bcu121.zip), then I get: 
```
java: symbol lookup error: /mnt/lib/cache/org.bytedeco.pytorch-2.1.2-1.5.10-linux-x86_64.jar/org/bytedeco/pytorch/linux-x86_64/libjnitorch.so: undefined symbol: _ZNK3c106Device3strB5cxx11Ev
```
Not sure where `_ZNK3c106Device3strB5cxx11Ev` would be defined? Hmm... maybe I should try the cxx11 ABI version? (i.e. https://download.pytorch.org/libtorch/cu121/libtorch-cxx11-abi-shared-with-deps-2.1.2%2Bcu121.zip)

If I switch to using
```
    "org.bytedeco" % "pytorch-platform" % "2.1.2-1.5.10", // https://mvnrepository.com/artifact/org.bytedeco/pytorch-platform
    "org.bytedeco" % "pytorch-platform-gpu" % "2.1.2-1.5.10", // https://mvnrepository.com/artifact/org.bytedeco/pytorch-platform-gpu
    "org.bytedeco" % "cuda-platform-redist" % "12.3-8.9-1.5.10", // https://mvnrepository.com/artifact/org.bytedeco/cuda-platform-redist
    "org.bytedeco" % "mkl-platform-redist" % "2024.0-1.5.10", // https://mvnrepository.com/artifact/org.bytedeco/mkl
```
then I get: 
```
java.lang.RuntimeException: nvrtc: error: failed to open libnvrtc-builtins.so.12.3.
  Make sure that libnvrtc-builtins.so.12.3 is installed correctly.
nvrtc compilation failed:
...
```
when I try to actually load a pytorch model. 

Looking in `/mnt/lib/cache/org.bytedeco.cuda-12.3-8.9-1.5.10-linux-x86_64-redist.jar/org/bytedeco/cuda/linux-x86_64$` I can see: 
```
drwxr-xr-x 2 ubuntu ubuntu      4096 Feb 13 19:37 ./
drwxr-xr-x 3 ubuntu ubuntu      4096 Feb 13 19:37 ../
lrwxrwxrwx 1 ubuntu ubuntu        15 Feb 13 19:37 libcublas.so -> libcublas.so.12
-rw-r--r-- 1 ubuntu ubuntu 106679344 Jan 26 13:29 libcublas.so.12
lrwxrwxrwx 1 ubuntu ubuntu        17 Feb 13 19:37 libcublasLt.so -> libcublasLt.so.12
-rw-r--r-- 1 ubuntu ubuntu 518358624 Jan 26 13:29 libcublasLt.so.12
lrwxrwxrwx 1 ubuntu ubuntu        15 Feb 13 19:37 libcudart.so -> libcudart.so.12
-rw-r--r-- 1 ubuntu ubuntu    703808 Jan 26 13:29 libcudart.so.12
lrwxrwxrwx 1 ubuntu ubuntu        13 Feb 13 19:37 libcudnn.so -> libcudnn.so.8
-rw-r--r-- 1 ubuntu ubuntu    142008 Jan 26 13:29 libcudnn.so.8
lrwxrwxrwx 1 ubuntu ubuntu        23 Feb 13 19:37 libcudnn_adv_infer.so -> libcudnn_adv_infer.so.8
-rw-r--r-- 1 ubuntu ubuntu 125098064 Jan 26 13:29 libcudnn_adv_infer.so.8
lrwxrwxrwx 1 ubuntu ubuntu        23 Feb 13 19:37 libcudnn_adv_train.so -> libcudnn_adv_train.so.8
-rw-r--r-- 1 ubuntu ubuntu 116106128 Jan 26 13:29 libcudnn_adv_train.so.8
lrwxrwxrwx 1 ubuntu ubuntu        23 Feb 13 19:37 libcudnn_cnn_infer.so -> libcudnn_cnn_infer.so.8
-rw-r--r-- 1 ubuntu ubuntu 570673312 Jan 26 13:29 libcudnn_cnn_infer.so.8
lrwxrwxrwx 1 ubuntu ubuntu        23 Feb 13 19:37 libcudnn_cnn_train.so -> libcudnn_cnn_train.so.8
-rw-r--r-- 1 ubuntu ubuntu 125431984 Jan 26 13:29 libcudnn_cnn_train.so.8
lrwxrwxrwx 1 ubuntu ubuntu        23 Feb 13 19:37 libcudnn_ops_infer.so -> libcudnn_ops_infer.so.8
-rw-r--r-- 1 ubuntu ubuntu  90829248 Jan 26 13:29 libcudnn_ops_infer.so.8
lrwxrwxrwx 1 ubuntu ubuntu        23 Feb 13 19:37 libcudnn_ops_train.so -> libcudnn_ops_train.so.8
-rw-r--r-- 1 ubuntu ubuntu  70926584 Jan 26 13:29 libcudnn_ops_train.so.8
lrwxrwxrwx 1 ubuntu ubuntu        14 Feb 13 19:37 libcufft.so -> libcufft.so.11
-rw-r--r-- 1 ubuntu ubuntu 177827520 Jan 26 13:30 libcufft.so.11
lrwxrwxrwx 1 ubuntu ubuntu       122 Feb 13 19:37 libcupti.so.12 -> /mnt/lib/cache/org.bytedeco.pytorch-2.1.2-1.5.10-linux-x86_64-gpu.jar/org/bytedeco/pytorch/linux-x86_64-gpu/libcupti.so.12
lrwxrwxrwx 1 ubuntu ubuntu        15 Feb 13 19:37 libcurand.so -> libcurand.so.10
-rw-r--r-- 1 ubuntu ubuntu  96259504 Jan 26 13:30 libcurand.so.10
lrwxrwxrwx 1 ubuntu ubuntu        17 Feb 13 19:37 libcusolver.so -> libcusolver.so.11
-rw-r--r-- 1 ubuntu ubuntu 115640600 Jan 26 13:30 libcusolver.so.11
lrwxrwxrwx 1 ubuntu ubuntu        17 Feb 13 19:37 libcusparse.so -> libcusparse.so.12
-rw-r--r-- 1 ubuntu ubuntu 267184960 Jan 26 13:30 libcusparse.so.12
lrwxrwxrwx 1 ubuntu ubuntu        39 Feb 13 19:37 libgcc_s.so.1 -> /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
lrwxrwxrwx 1 ubuntu ubuntu       119 Feb 13 19:37 libgfortran.so.5 -> /mnt/lib/cache/org.bytedeco.openblas-0.3.26-1.5.10-linux-x86_64.jar/org/bytedeco/openblas/linux-x86_64/libgfortran.so.5
lrwxrwxrwx 1 ubuntu ubuntu        38 Feb 13 19:37 libgomp.so.1 -> /usr/lib/x86_64-linux-gnu/libgomp.so.1
lrwxrwxrwx 1 ubuntu ubuntu       119 Feb 13 19:37 libiomp5.so -> /mnt/lib/cache/org.bytedeco.pytorch-2.1.2-1.5.10-linux-x86_64-gpu.jar/org/bytedeco/pytorch/linux-x86_64-gpu/libiomp5.so
lrwxrwxrwx 1 ubuntu ubuntu        12 Feb 13 19:37 libnccl.so -> libnccl.so.2
-rw-r--r-- 1 ubuntu ubuntu 209818616 Jan 26 13:30 libnccl.so.2
lrwxrwxrwx 1 ubuntu ubuntu        18 Feb 13 19:37 libnvJitLink.so -> libnvJitLink.so.12
-rw-r--r-- 1 ubuntu ubuntu  52190720 Jan 26 13:30 libnvJitLink.so.12
lrwxrwxrwx 1 ubuntu ubuntu        18 Feb 13 19:37 libnvToolsExt.so -> libnvToolsExt.so.1
-rw-r--r-- 1 ubuntu ubuntu     40136 Jan 26 13:39 libnvToolsExt.so.1
lrwxrwxrwx 1 ubuntu ubuntu        14 Feb 13 19:37 libnvrtc.so -> libnvrtc.so.12
-rw-r--r-- 1 ubuntu ubuntu  60792048 Jan 26 13:39 libnvrtc.so.12
lrwxrwxrwx 1 ubuntu ubuntu       119 Feb 13 19:37 libopenblas.so.0 -> /mnt/lib/cache/org.bytedeco.openblas-0.3.26-1.5.10-linux-x86_64.jar/org/bytedeco/openblas/linux-x86_64/libopenblas.so.0
lrwxrwxrwx 1 ubuntu ubuntu        42 Feb 13 19:37 libquadmath.so.0 -> /usr/lib/x86_64-linux-gnu/libquadmath.so.0
lrwxrwxrwx 1 ubuntu ubuntu       119 Feb 13 19:37 libtbb.so.2 -> /mnt/lib/cache/org.bytedeco.pytorch-2.1.2-1.5.10-linux-x86_64-gpu.jar/org/bytedeco/pytorch/linux-x86_64-gpu/libtbb.so.2
```
i.e. `libnvrtc.so` is there, but no `libnvrtc-builtins.so.12.3`. 

???

HGuillemet:
### Work in Progress ###

Included in this PR:
* Update to PyTorch 2.2.1
* Restore `ExampleStack` and `TensorExampleStack` constructors
* Generate more overloads for methods taking an array ref
* Add parsing of `CUDAFunctions.h` (wrappers around commonly used CUDA API functions)
* ~~Add `AOTInductor` (new way to run models exported from Python)~~
* Virtualize `FunctionPreHook` and `FunctionPostHook` to enable setting hooks on autograd graph
* Passing a vector of tensors, and some other classes,  by value does not remove the data anymore
* Add `Module.asXXX` to test if a `Module` is of a specific subclass (and do the cast).
* Add new AMD Float8 types `Float8_e5m2fnuz` and `Float8_e4m3fnuz` with unsigned zero
* Add preload of `nvrtc-builtins` (fixes #1468)


sbrunk:
@HGuillemet thanks for the update and especially adding the [AOTInductor](https://pytorch.org/docs/main/torch.compiler_aot_inductor.html) mapping. I think that's an interesting new variant to be able to use the new optimizations at least for inference. Training still needs to happen in Python in that case but we can export it to a C++ lib and then use that from Java with the bindings.

Have you been able to try it already?I'll give it a try myself,  just curious if you got something working already.

HGuillemet:
No I didn't, yet.
I'm training from Java, so there is little chance I use it in the near future. I must rely on Torchscript when I need to import python models, and not a lot of people bother to writer torchscript compatible models :(

saudet:
There's probably some template somewhere that requires them. You'll probably get the same error on Linux and Mac if you try to link with `-Wl,--no-undefined`, so try to fix the errors you get with that, and it should fix those errors on Windows too.

HGuillemet:
Thanks for the suggestion. Adding the linker option raised an error about cudnn not linked to jnitorch_cuda. Let's see but I doubt it's related to the error on windows.

HGuillemet:
It seems it has been spotted and fixed upstream in https://github.com/pytorch/pytorch/commit/79ba39710e89bf9f2f12b199aae87afc402a3176 after 2.2.0 release.
I guess we'd better postpone the inclusion of the AOTInductor feature to next release.

HGuillemet:
To enable the setting of hooks in autograd graphs, I need to virtualize `FunctionPreHook` and `FunctiontHook`, which have a virtual method taking a vector of tensors and returning a vector of tensors. Compilation passes only if I remove the `valueTypes` in this info:
```Java
new Info("std::vector<torch::Tensor>", "std::vector<at::Tensor>", "std::vector<torch::autograd::Variable>", "torch::autograd::variable_list")
  .valueTypes("@Cast({\"\", \"std::vector<torch::Tensor>\"}) @StdMove TensorVector")
  .pointerTypes("TensorVector").define())
```
I wonder why this `valueTypes` is here.
It will save a copy when we pass a vector of tensors to a native functions but OTOH it will destroy the vector, while the user could need it after the function call.
If I understand well, if a native function takes a rvalue ref (`&&`) , parser will generate `@ByRef(true)` which is enough to avoid copies.
@saudet could you share your infinite knowledge about this point ?
Could it break something if I remove the `valueTypes` ? First attempts seem to show it does not.

There are some other types with this kind of `@Cast @StdMove` value types (`DataPtr`, `Storage`, `TensorMaybeOwned`, `TensorBaseMaybeOwned`, `TensorName`, `EdgeVector`)

saudet:
If you're not getting any compile errors, then I guess PyTorch's API was improved so that we don't need them anymore, yes

HGuillemet:
@sbrunk could you run your tests on this PR ? 
Anything you'd like to be added ?

sbrunk:
> @sbrunk could you run your tests on this PR ? Anything you'd like to be added ?

Tests are looking good!

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
```diff
@@ -75,7 +77,7 @@ public void map(InfoMap infoMap) {
 
             .put(new Info().enumerate().friendly())
             .put(new Info().javaText("import org.bytedeco.pytorch.*;"))
-            .put(new Info().javaText("import org.bytedeco.pytorch.functions.*;"))
+            .put(new Info().javaText("import org.bytedeco.pytorch.cuda.functions.*;"))
```
saudet, line 80: Why do we need a new package for this?

HGuillemet: If I remember well, all classes in package `functions` will be compiled during a build without cuda. 
And I added a function that needs a class only available during cuda builds: `TraceEntry`.
So builts without cuda would fail if this function is not in a cuda specific package.

HGuillemet: Last time I tried to understand how the builder determines the list of packages/directories to compile, I failed.
But the thing is that `cuda.functions` is included in a cuda built and is not in a non-cuda built.

saudet: Where would that be happening?! I don't get how this is supposed to work like this, but not if they are all inside `functions` instead.

saudet: Oh wait a minute, that's the target package for torch_cuda ... I'm not sure I understand the reason why you want to make things complicated like this, but ok that's why it's working like this

saudet:
If you're not planning on making more changes for now, we can merge this?

HGuillemet:
Ok for me.

pytorch/include_list.pl
```diff
@@ -62,4 +64,6 @@ sub go {
 // c10/cuda/CUDAGuard.h
 EOF
 
+# Doesn't compile on Windows. Waiting for 2.2.1.
+#go('ATen/cudnn/Descriptors.h', 'ATen/cudnn/Types.h', 'c10/cuda/CUDAGuard.h', '-I/opt/cuda/targets/x86_64-linux/include/', 'torch/csrc/inductor/aoti_model_container_runner_cuda.h');
 go('ATen/cudnn/Descriptors.h', 'ATen/cudnn/Types.h', 'c10/cuda/CUDAGuard.h', '-I/opt/cuda/targets/x86_64-linux/include/');
```
saudet, line 69: Can we add those back now?

pytorch/src/gen/java/org/bytedeco/pytorch/AdaptiveAvgPool1dImpl.java
```diff
@@ -44,9 +44,6 @@ private native void allocate(
           @Const @ByRef AdaptiveAvgPool1dOptions options_);
     /** Pointer cast constructor. Invokes {@link Pointer#Pointer(Pointer)}. */
     public AdaptiveAvgPool1dImpl(Pointer p) { super(p); }
-    /** Downcast constructor. */
-    public AdaptiveAvgPool1dImpl(Module pointer) { super((Pointer)null); allocate(pointer); }
-    @Namespace private native @SharedPtr @Name("SHARED_PTR_NAMESPACE::dynamic_pointer_cast<torch::nn::AdaptiveAvgPool1dImpl, torch::nn::Module>") void allocate(@SharedPtr Module pointer);
```
saudet, line 49: Why have **all** the "Downcast constructor" been removed??

pytorch/include_list.pl
HGuillemet: No, I checked and the fix mentioned above has not been merged for 2.2.1. 
I will probably be for 2.3.

pytorch/src/gen/java/org/bytedeco/pytorch/AdaptiveAvgPool1dImpl.java
HGuillemet: In 2.2, they removed the virtual inheritance between `Module` and its subclasses. So I removed the `upcast` info.

saudet (approved):

saudet:
> 2024-03-04T11:44:04.1783504Z Caused by: java.io.IOException: No space left on device

Could you try to fix this? We probably just need to uninstall a couple of large unnecessary packages...

HGuillemet:
I had already added a bunch of `rm` in deploy-ubuntu once, on downloaded archives after there installation, and you reverted that.
I can try to add them again, that seems the easiest and fastest way to make room.
In a new PR ?

saudet:
Really? Could you point me to that revert and I'll try it on the actions branch here

HGuillemet:
I'm seeing it was on deploy-centos, in fact:
https://github.com/bytedeco/javacpp-presets/pull/1360/commits/3e3fe5c096b76b0e8cd9a116d1986f5858308391

I'm reviewing deploy-ubuntu and adding similar cleanup. Shall I push the commit here or on a new PR ?

saudet:
Ah, that won't be enough. We'll probably need to remove a lot more stuff. You can try it here, but we won't know if it works until actual deploy, so I don't know. Let's check how many more GB we can with `df -h` I guess for now is good indication

HGuillemet:
I pushed https://github.com/bytedeco/javacpp-presets/commit/aaa37a17afba0d06fc4803a48d5caa6d69eb2224 on my branch but it doesn't update this PR now that it's merged.

Anyway, this commit will indeed only save ~ 700Mb for pytorch build, due to mkl archive removal.

If it's not enough, what about a maven clean phase on main artifact after its deploy phase and before the deploy phase of the platform/ext artifact ? This should get rid of cppbuild directory (about 7 or 8G)

saudet:
That could work, yes

PR #1490, opened 2024-04-08
[Modesecurity] Upgrade modesecurity to be able to use it at Debian12 that does not contain libpcre++-dev lib anymore

yaandy:


modsecurity/cppbuild.sh
```diff
@@ -23,15 +23,16 @@ fi
 
 case $PLATFORM in
     linux-x86_64)
+        yum install -y pcre2-devel
```
andriyyarish, line 26: To be removed - in forked repo changes to .github/actions/deploy-centos/action.yml does not take any affect somewhy and original code was executed in the job

saudet:
Right, please update the Ubuntu one

andriyyarish:
image
image
image

As I understand it runs on Centos, but looks like the action definition is not from a forked branch with changes.
I even try to add it to ubuntu 
image

but looks like it is not used for modsecurity module

saudet:
> As I understand it runs on Centos, but looks like the action definition is not from a forked branch with changes. I even try to add it to ubuntu image
> 
> but looks like it is not used for modsecurity module.

Right, so I'm asking you to ignore the centos action and update the workflow to use the ubuntu one instead

> @saudet Could you please suggest how to get a snapshot artifacts ? can't understand where they should be posted after the build in a branch

If you have an account for Sonatype OSSRH, you can set those secrets in your own fork:
https://github.com/bytedeco/javacpp-presets/blob/master/.github/workflows/modsecurity.yml#L15-L18

PR #1497, opened 2024-04-24
Add vulkan patch for segmentation fault error

grill2010:
Some users of my application reported a crash when using Vulkan in combination with an Nvidia GPU. Upon further investigation it seems that most likely only Nvidia GPUs from the 30xx series are affected (but not 100% sure). It seems to happen in other programs as well e.g. Chiaki and luckily @nowrep there found the issue/ workaround

https://github.com/streetpea/chiaki4deck/issues/239#issuecomment-1931408400

This was also reported to ffmpeg now

https://trac.ffmpeg.org/ticket/10847

I applied the fix in a custom build of javacpp-presets and created a version of my application with this patch, the users who encountered this crash confirmed that it is no longer happening and the Vulkan decoder can be used successfully.


grill2010:
The patch seems to work for others as well

https://github.com/mpv-player/mpv/issues/13375#issuecomment-2074852316

saudet:
Sure, but please rename it to ffmpeg-vulkan.patch

grill2010:
Sure, done :)

saudet (approved):

PR #1498, opened 2024-05-03
Update to Pytorch 2.3.0

HGuillemet:
Included in this PR:
* Update to PyTorch 2.3.0
* Add AOTInductor (new way to run models exported from Python)


HGuillemet:
The C++ API can change depending on the platform.
Currently blocking is `Half` constructor and cast operator that uses `float16_t` or ARM64 and `float` on x86_64 (which doesn't have float16_t).
I'm considering adding some explicit JNI bridge that will call either the float or float16 version depending on the platform, and skip the platform-dependent C++ API.
@saudet, any better idea ?

saudet:
Just disable float16_t support, it's not portable.

HGuillemet:
If Parser is told to parse only the `#if !defined(__arch64__)` I get a linking error on Mac Apple Silicon because there is no C++ constructor taking a float 32 or a casting operator returning a float 32 (see check error log).

HGuillemet:
I can try to patch libtorch source to disable the `#if defined(__arch64__)` branch and force the creation of float 32 variants on Mac, hoping there is no other arm64-specific  parts of he code in libtorch relying on the `float16` variant. But it seems more hazardous than the custom JNI trick. What do you think ?

saudet:
> If Parser is told to parse only the `#if !defined(__arch64__)` branch I get a linking error on Mac Apple Silicon because there is no C++ constructor taking a float 32 or a casting operator returning a float 32 (see check error log).

I assume we can convert from/to float and float16_t just like we can between float and double, so it should work if we add casts.

pytorch/src/main/resources/org/bytedeco/pytorch/include/platform_unification.h
```diff
@@ -0,0 +1,21 @@
+/* Add the necessary C++ calls to present a unified C++ API whatever the platform */
+
+namespace javacpp {
+
+   inline c10::Half *allocate_Half(jfloat value) {
+#if defined(__aarch64__) && !defined(C10_MOBILE) && !defined(__CUDACC__)
+       return new c10::Half((float16_t) value);
```
saudet, line 7: Please don`t create a header file just for this. Just use `@Cast()` from JavaCPP

HGuillemet: For the constructor, we need to cast the float coming from Java to a float16_t on ARM64 only. 
So something like this could do the trick, as a replacement for the header file:
```Java
  private native void allocate(@Cast("/**/\n#if defined(__aarch64__) && !defined(C10_MOBILE) && !defined(__CUDACC__)\nfloat16_t\n#else\nfloat\n#endif\n/**/") float value);
```
Or do you have something else in mind ?

saudet: Yeah, just a normal `@Cast("float16_t") float`. What kind of error do you get with that??

HGuillemet:
```
error: 'float16_t' was not declared in this scope
```
on non-arm64 platforms. Why would you expect anything else ?

saudet: Of course, on other platforms, don't use the cast

saudet:
Increasing swap space doesn't work?

pytorch/src/main/resources/org/bytedeco/pytorch/include/platform_unification.h
HGuillemet: You mean generating different Java code for each platform ? A code that will end up in the shared Java jar ?
I guess that it's not a big deal as long the change only concerns JavaCPP annotations and not the Java API itself,
and that all github workers run the parser instead of using what's in `gen` directory of the repo.
But is it really better that the `#ifdef` solution ?

HGuillemet:
I'll try that.
Do I change `deploy-ubuntu/action.yml` or do I add a swap file in pytorch cppbuild ?

HGuillemet:
I added a check for pytorch in `action.yml`, but if that works, the best would be to set a dedicated environment variable in workflow files, like `SWAP_SPACE` and use it in `action.yml`

pytorch/src/main/resources/org/bytedeco/pytorch/include/platform_unification.h
saudet: Yes, something like that. I think it's better because we can write it in Java.

HGuillemet: I agree than the `ifdef` in the info is quite ugly. That's why I thought about a specific header file to bridge between a unique Java API and changes in C++ API.
In such header, we can use the exact same `#ifdef` clauses that the one used in pytorch source and it's quite readable.
There are other things we could add there. For instance I skipped `getCurrentCUDABlasLtHandle` for all platforms because Pytorch does not include the function when compiling with VS for some reason. Probably nobody will ever try to call this from Java, but if it's needed, we could add a wrapper with proper `#ifdef` and either return null or throws an exception in the case we compiled with VS. We cannot do this in Java because we cannot declare a native function only for some platforms and keep the same java bytecode.

saudet:
Like you said, after build, it doesn't matter if the annotation is there or
not, so don't worry about it

saudet:
To use the actions from your fork, you'll need to change the URL in the workflow

HGuillemet:
A swap space of 2Gb works, 4 and 0 do not.

saudet:
Why not 4GB? What happens?

HGuillemet:
I'm not sure because the log file of the 4G attempt seems truncated. But there remains only 10G of disk space after installations and creation of swap file, and before build. I won't be surprised if we run out of disk space.
We really are at the limits of what the worker can offer...

saudet:
So, it probably a better idea to figure out how to free some space?

HGuillemet:
Better than what ? 
Build passes for now, so we are ok. 
But in prevision of next Pytorch releases that will need probably more disk space and more memory to compile, finding spare disk space would be useful.
Also upgrading to cuda 12.4 maybe could help.

saudet:
2GB might not be enough for PyTorch 2.4.0, so before merging this let's wait and see I guess

saudet:
This looks useful: https://github.com/marketplace/actions/free-disk-space-ubuntu

saudet:
We can probably just erase a couple of those without the tool though:

> ```
> => Android library: Saved 14GiB
> => .NET runtime: Saved 2.7GiB
> => Haskell runtime: Saved 0B
> => Large misc. packages: Saved 5.3GiB
> => Tool cache: Saved 5.9GiB
> => Swap storage: Saved 4.0GiB
> 
> Total: Saved 31GiB
> ```

Does that mean swap is enabled by default now??

pytorch/src/main/resources/org/bytedeco/pytorch/include/platform_unification.h
saudet: APIs that differ like that between platform that isn't portable in the first place probably isn't part of the actual intended external API, so let's just skip those.

HGuillemet:
> This looks useful: https://github.com/marketplace/actions/free-disk-space-ubuntu

Indeed. We can use this freeing action.

But why waiting for 2.4.0 before merging ?

> Does that mean swap is enabled by default now??

I changed `deploy-ubuntu` so that it reads a `SWAP_SIZE` env variable and set this variable to 4 in the mkl workflow (`deploy-ubuntu` added 4G of swap for mkl before my change), and 2 for pytorch.

saudet:
Let's not make an option, let's just set the swap to a value that works for everything like 4GB, which means using an action here is annoying, so let's just ditch android and dotnet instead like https://github.com/actions/runner-images/issues/2606#issuecomment-772683150

HGuillemet:
Ok for the ditching, but keeping option to use some disk space for extra swap, depending on the workflow needs, seems interesting to me.

saudet:
Why do you want to make an option? It's not going to used by anything

HGuillemet:
Maybe some other builds would fail if we remove 4G of disk for the swap. Like pytorch currently does.
Why do you want to remove the option ? There are already other options like `CI_DEPLOY_NEED_*` why not `CI_DEPLOY_NEED_SWAP` ?

saudet:
Those options are there because the some builds fail when they are true, but some others fail when they are false. That doesn't happen for something like swap space

HGuillemet:
Is dotnet used by any build ?
Is /usr/local/android used by android builds ?

HGuillemet:
11Gb saved from ditching android and dotnet

.github/actions/deploy-ubuntu/action.yml
```diff
@@ -272,7 +270,13 @@ runs:
...
           find -name *.deb | while read f; do ar -xv $f; $SUDO tar --totals -xvf data.tar.xz -C /; done;
         fi
-
+        
+        # Add some swap. Needed for pytorch and mkl.
+        $SUDO fallocate -l 4GB /swapfile
+        $SUDO chmod 600 /swapfile
+        $SUDO mkswap /swapfile
+        $SUDO swapon /swapfile
+
```
saudet, line 279: Let's move that to just below the "Free disk space" section since it's related

saudet (approved):

PR #1499, opened 2024-05-06
Enable Va-API hw acceleration in ffmpeg for h264 and HEVC

grill2010:
+) Enables Va-API  support on Linux x86 and Linux x64
+) Add ffmpeg patches for invalid reference frames in Va-API hevc/ h264 to avoid stream glitches (https://github.com/streetpea/chiaki4deck/issues/108). (patches are used in Chiaki)


grill2010:
Was in contact with one of the Chiaki devs, the patches are not needed as this is just because of a custom logic within Chiaki itself. Will do some further testing the next couple of days and will update my pull request.

saudet:
You're going to need to use temporarily the actions from your fork, but why not simply do apt-get install libva-dev?

grill2010:
Because I have another branch where I back ported some stuff and this didn't work as was downloading an older version of libva-dev and the ffmpeg build failed. Building it itself might not be needed when we upgrade to a newer ubuntu build image but not sure if that will break other things. I will continue testing in a temp branch.

PR #1510, opened 2024-06-07
[PyTorch] Update to 2.4.0, add distributed

### LINKED ISSUE #1503: [PyTorch 2.2.2-1.5.11-SNAPSHOT] Training produces poor MNIST model on Windows
On macOS and Linux, PyTorch 2.2.2 training with MNIST produces good models that has training accuracy > 90%. However, the same code reports very low accuracy (about 11%) on Windows. The same code works fine with 2.2.1 on Windows. You can run your [sample code](https://github.com/bytedeco/javacpp-presets/tree/master/pytorch) to reproduce the issue. The below code is to calculate the accuracy:

```
        net.eval();
        int correct = 0;
        int n = 0;
        for (ExampleIterator it = dataLoader.begin(); !it.equals(dataLoader.end()); it = it.increment()) {
            Example batch = it.access();
            var output = net.forward(batch.data());
            var prediction = output.argmax(new LongOptional(1), false);
            correct += prediction.eq(batch.target()).sum().item_int();
            n += batch.target().size(0);
            // Explicitly free native memory
            batch.close();
        }
        System.out.format("Training accuracy = %.2f%%\n", 100.0 * correct / n);
```

### LINKED ISSUE #1504: [PyTorch] Training is very slow on Linux.
Training 10 epochs of MNIST (the sample code from your project README) on takes > 500 seconds on Linux (24 cores, ubuntu 22.04). It takes only about 50 seconds on an old mac (4 cores).  Both use CPU (no GPU or MPS).

HGuillemet:
Included in this PR:

* Update to PyTorch 2.4.0
* Add distributed framework (backend Gloo)
* Add a minimal binding for std::chrono, needed by distributed (TBC)
* Add an adapter for `intrusive_ptr`, enabling transparent usage like for `shared_ptr`
* Add an adapter for `weak_ptr`, enabling transparent usage (could be moved to JavaCPP ?)
* Add an optional Maven dependency to cuda. This allows to use types defined in the cuda presets. Users calling a method of PyTroch presets using one of these types should then have the cuda presets. Which also means the cuda library of the presets will be used, and not the system one (unless the user sets `org.bytedeco.javacpp.pathsFirst` and their system cuda has same version than the cuda presets).
* Merge `functions` packages to main package
* Fix #1503
* Fix #1504
* Fix support for OpenMP on macos

Remains to be done:
* Determine if other distributed backends are needed (MPI or UCC.  NCCL is not supported on Windows yet).


pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
```diff
@@ -311,22 +320,26 @@ public void map(InfoMap infoMap) {
             .put(new Info().javaText("import org.bytedeco.pytorch.Allocator;"))
             .put(new Info().javaText("import org.bytedeco.pytorch.Function;"))
             .put(new Info().javaText("import org.bytedeco.pytorch.functions.*;"))
+            .put(new Info().javaText("import org.bytedeco.pytorch.chrono.*;"))
             .put(new Info().javaText("import org.bytedeco.pytorch.Module;"))
             .put(new Info().javaText("import org.bytedeco.javacpp.annotation.Cast;"))
+            .put(new Info().javaText("import org.bytedeco.pytorch.presets.torch.IntrusivePtr;"))
```
saudet, line 326: Please put those in the helper package and make the presets depend on that

HGuillemet: Done

saudet: Actually, we only need packages when there is a name conflict. Why do you want to put things in different packages??

HGuillemet: Which package are you talking about ? 
New packages in this PR are:
*  gloo: this is a "third-party" independent piece of software bundled into pytorch. It wont't be used by 99% of users. It contains simple class names like "Device", "Buffer"...  that would conflict with Pytorch names unless renamed.
* chrono : these are manual classes for binding some std::chrono classes. Again this is separate software from Pytorch. It won't be used by most people either. It could eventually be moved to another presets if completed enough.

Main package is already quite crowded, and I'm sure there are some names shadowed by others. If redesigning from scratch I would  suggest to move some of torch namespaces that are well separated and targeted for some specific uses to separate Java packages, like `torch::jit`.

HGuillemet: ( I was tempted to use a separate package for `distributed` (`c10d` namespace) but I knew you won't like that :). Also it would have really needed 2 packages: `distributed` and `cuda.distributed`. So I prefered to put them in main packages.)

saudet: Like all of them, if you want to make packages, at least make them related to the C++ API and not just come up with random names like you're doing. Like where does "cuda" and "functions" come from? What do they correspond to exactly with the C++ API?

HGuillemet: `cuda` contains the classes than must only by used if you have the `-gpu` classifier. I think it's a good thing to have them in a separate package. It contains classes from the `c10::cuda` namespace and some dependency classes.
`function` contains Java mapping for the `std::function` instances used by the presets. They don't map to Pytorch C++ names.
Where is randomness here ?
But if you want me to merge the function package into the main one, I won't mind.

saudet: The convention that I've been trying to keep for that is to match the library names. We already have "torch" and "torch_cuda" as libraries, so we should be using that. However, I understand that you wanted to not break backward compatibility so you came up with torch.cuda, so that's find I guess because torch::cuda exists, but there's no torch::functions or torch::gloo. Like I keep telling you, if you want to create some other API, go do it somewhere else. The idea here is to have the closest possible mapping to the C++ API. Please do that

HGuillemet: About chrono: what about creating an independent presets for that ?
I'm seeing that depthai and tesseract use some chrono classes and the presets map them to `Pointer`.

About gloo: if I move them to main packages, I'll have to prefix all names by `Gloo`. Is that really better ? I remind you that this is not pytorch code. I can first only keep the classes that are strictly necessary when used from PyTorch.

About function: ok, I'll remove these packages. This could break some code of users of last versions, they will have to remove their import.

saudet: The point is about making this consistent with the C++ API, so like we could put std:: function in "function", std::chrono in "chrono", torch::jit in "jit", whatever::gloo in "gloo", that's fine. Would you be ok following this pattern?

HGuillemet: The namespace structure of PyTorch is quite messy. We have:
```
at
at::cuda
at::functorch
at::impl
at::indexing
at::internal
at::namedinference
at::native
c10
c10::cuda
c10d
c10::detail
c10::detail::infer_schema
c10d::intra_node_comm
c10::impl
c10::ivalue
c10::util
c10::utils
caffe2
caffe2::serialize
torch
torch::autograd
torch::data
torch::data::datasets
torch::data::example
torch::data::samplers
torch::dynamo::autograd
torch::enumtype
torch::inductor
torch::jit
torch::jit::detail
torch::jit::interpreter
torch::nn
torch::nn::detail
torch::nn::functional
torch::nn::utils::rnn
torch::optim
torch::profiler::impl
torch::serialize
torch::utils
```
Some of them could be merged since PyTorch defined namespace aliases if I remember well, but it doesn't seem feasible without support in JavaCPP to automatically map namespaces to Java packages.

What we can do today is to target a Java package when we parse a set of header files. So, rather than namespaces, Java packages must match either independent/thirdparty API used by the library (like gloo or std:chrono), or subsets of the library API that is only enabled when including some specific headers (like cuda, maybe jit, maybe distributed/c10d).

saudet: Yes, that's why I keep saying let's not do packages. You're the one that wants to do them, so please find a way to make them consistent with the C++ API!!

HGuillemet: chrono and gloo are clearly independent piece of soft, with their own includes. So do you agree to keep packages for those ?
I can merge the "manual" function classe in main package, and I can also try to merge the cuda package. 
Would that be ok ?

saudet: > chrono and gloo are clearly independent piece of soft, with their own includes. So do you agree to keep packages for those ?

Right, so use the names of those libraries, not the namespaces. That works fine for that purpose

> I can merge the packages for "manual" std::function instances in main package, and I can also try to merge the cuda package.

We already have separate torch and torch_cuda libraries, so that's the name we should give those packages, but like you said to keep backward compatibility, you left torch in org.bytedeco.pytorch and put torch_cuda in org.bytedeco.pytorch.cuda, so that's alright I guess. As for the std::function instances, yeah, I don't see why you want them in a separate package

HGuillemet: Ok, so moving `functins` only.
chrono is part of standard C++, and the name of the library is chrono. gloo lbrary name is gloo.
libtorch libraries are really torch_cpu and torch_cuda. torch is an empty library linking to both (or only one if pytorch was not compiled with cuda).

saudet: > Ok, so moving `functins` only.

Ok, sounds consistent

> chrono is part of standard C++, and the name of the library is chrono. gloo lbrary name is gloo.
libtorch libraries are really torch_cpu and torch_cuda. torch is an empty library linking to both (or only one if pytorch was not compiled with cuda).

There is no "libchrono" though, it's libc++, libstdc++, etc...

saudet: Could you try to move your chrono package here?
https://github.com/bytedeco/javacpp/tree/master/src/main/java/org/bytedeco/javacpp
If that works, it's probably a good place for that

HGuillemet: It needs to be completed to be anything useful, but I think it's a good thing to mutualize it somewhere.
Why not directly in JavaCPP. But since it'll concern few people and will grow if completed enough, isn't it better to put it in a separate presets ?

saudet: The JavaCPP presets is meant to do C++ itself. Trying to chunk down of C++ is not something I want to do, no. I keep telling to come up with some sort of consistent way to using packages and you just keep trying to make it more complicated with no rationale. Try to start with the rationale

saudet: So, let's start with how big you anticipate those "C++ presets" to become at some point in the future. Is it close to 1 MB, 10 MB, 100 MB, or 1 GB?

HGuillemet:
@halfeng @sbrunk 
This PR can be tested using artifact `fr.apteryx:pytorch-platform-gpu:2.3.1-1.5.11-SNAPSHOT` or `fr.apteryx:pytorch-platform:2.3.1-1.5.11-SNAPSHOT` from maven repo `https://s01.oss.sonatype.org/content/repositories/snapshots`.

saudet:
@HGuillemet Your account doesn't already have permission to publish for org.bytedeco??

HGuillemet:
I don't have the secrets. Anyway PR builds don't publish.

saudet:
You  should be able to publish to org.bytedeco using your own account. I'm pretty sure I gave you access a long time ago.

saudet:
Or maybe not, what the name of your OSSRH account?

HGuillemet:
`HGuillemet`. But don't bother.

saudet:
@HGuillemet Let's just merge this and fix anything broken with the upgrade to PyTorch 2.4?

HGuillemet:
I'm about to add a commit for 2.4.0 upgrade

saudet:
Ok, we can do that too

HGuillemet:
Before merging, you must decide what to do with bytedeco/javacpp#766

HGuillemet:
Also to decide: whether adding the cuda dependency is a good thing or not.

saudet:
An optional dependency on the presets for CUDA that only torch_cuda needs, sure, that's fine.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
```diff
@@ -72,12 +68,16 @@ public void map(InfoMap infoMap) {
...
-            .put(new Info().javaText("import org.bytedeco.pytorch.cuda.functions.*;"))
-            .put(new Info().javaText("import org.bytedeco.pytorch.Error;"))
+            .put(new Info().javaText("import org.bytedeco.pytorch.helper.*;"))
+            .put(new Info().javaText("import org.bytedeco.cuda.cudart.*;"))
+            .put(new Info().javaText("import org.bytedeco.cuda.cusparse.*;"))
+            .put(new Info().javaText("import org.bytedeco.cuda.cublas.*;"))
+            .put(new Info().javaText("import org.bytedeco.cuda.cusolver.*;"))
+            .put(new Info().javaText("import org.bytedeco.cuda.cudnn.*;"))
+            // .put(new Info().javaText("import org.bytedeco.cuda.nccl.*;")) // Not on Windows
+            .put(new Info().javaText("import org.bytedeco.pytorch.chrono.*;"))
```
saudet, line 80: Instead of doing that, please add cudnn.class to the `@Platform(inherit=...` list.

HGuillemet:
> An optional dependency on the presets for CUDA that only torch_cuda needs, sure, that's fine.

Consider my comment on top post: the dependency will be mandatory when using CUDA. Presently, most users use their system cuda I guess, they will now have to use cuda presets.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
HGuillemet: Done for cuda, but what `helper.class` and `chrono.class` are you talking about ?

saudet: Right, JavaCPP assumes the helper classes to be in the same package as the generated classes, so let's remove that package too?

As for chrono, see https://github.com/bytedeco/javacpp/pull/766#issuecomment-2264430580

HGuillemet:
@saudet please restart the jobs that failed

.github/actions/deploy-macosx/action.yml
```diff
@@ -26,8 +26,7 @@ runs:
 
         brew uninstall --force --ignore-dependencies gcc gcc@7 gcc@8 gcc@9 gcc@10 gcc@11 gcc@12 gcc@13 gcc@14 little-cms2 maven openblas r
         brew install boost ccache swig autoconf-archive automake cmake libomp libtool libusb ant nasm xz pkg-config sdl2 gpg1 bison flex perl ragel binutils gradle gmp isl libmpc mpfr geoip pcre ssdeep yajl
-        brew link --force libomp
-
```
saudet, line 30: Why do you need to remove this line?

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
```diff
@@ -72,12 +75,8 @@ public void map(InfoMap infoMap) {
...
         infoMap
-            .put(new Info("basic/containers").cppTypes("c10::optional"))
-
             .put(new Info().enumerate().friendly())
             .put(new Info().javaText("import org.bytedeco.pytorch.*;"))
-            .put(new Info().javaText("import org.bytedeco.pytorch.cuda.functions.*;"))
-            .put(new Info().javaText("import org.bytedeco.pytorch.Error;"))
             .put(new Info().javaText("import org.bytedeco.pytorch.global.torch.DeviceType;"))
             .put(new Info().javaText("import org.bytedeco.pytorch.global.torch.ScalarType;"))
             .put(new Info().javaText("import org.bytedeco.pytorch.global.torch.MemoryFormat;"))
```
saudet, line 82: Why do we need those imports here?

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
```diff
@@ -21,37 +21,40 @@
...
-                "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.3/lib/x64/",
-                "/usr/local/cuda-12.3/lib64/",
-                "/usr/local/cuda/lib64/",
-                "/usr/lib64/"
-            }
+            exclude = {
+                "<cublas.h>", // pytorch includes cublas_v2, which is not compatible with cublas included from inherited cudnn presets
+                "<driver_functions.h>" // causes #warning
+            },
+            library = "jnitorch"
```
saudet, line 57: Please don't do that, leave it as jnitorch_cuda

pytorch/src/main/java/org/bytedeco/pytorch/presets/gloo.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.bytedeco.pytorch.presets.torch.template;
+
+/**
+ * @author Hervé Guillemet
+ */
+@Properties(
+    inherit =  { torch.class, chrono.class },
+    value = {
+        @Platform(
+            library = "jnitorch"
```
saudet, line 40: Please don't do that, leave it as jnigloo

pytorch/cppbuild.sh
```diff
@@ -181,22 +203,37 @@ TORCH_API std::ostream& operator<<(std::ostream& stream, const nn::Module& modul
...
 ln -sf pytorch/torch/bin ../bin
 
-# fix library with correct rpath on Mac
 case $PLATFORM in
     macosx-*)
-        cp /usr/local/lib/libomp.dylib ../lib/libiomp5.dylib
-        chmod +w ../lib/libiomp5.dylib
-        install_name_tool -id @rpath/libiomp5.dylib ../lib/libiomp5.dylib
-        install_name_tool -change @rpath/libomp.dylib @rpath/libiomp5.dylib ../lib/libtorch_cpu.dylib
+        cp $(brew ls libomp|grep libomp.dylib) ../lib
```
saudet, line 232: Please revert that. This is necessary to make it work with MKL

.github/actions/deploy-macosx/action.yml
HGuillemet: This line is harmful when compiling with gcc, because the linker will pick libomp (that is llvm implementation of openmp) instead of libgomp. In the case of Pytorch and other presets using llvm, it's ok but not necessary because cmake, at least in its recent versions, can find libomp in the keg path.
If you want to keep it, it'd be better in a  `else` block of a an `if [[ -n ${CI_DEPLOY_NEED_GCC:-} ]];` block

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
HGuillemet: We don't indeed.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
HGuillemet: If I remember well, I had to do this because some callback `std::function` were used from headers parsed both in torch and torch_cuda, which results in Generator producing same glue code in torch JNI and torch_cuda JNI, and in a linking error.
But I just tried again without `library =` and it worked, so I guess I finally didn't map the offending function.

pytorch/cppbuild.sh
HGuillemet: There is no MKL anymore on MacOSX.

.github/actions/deploy-macosx/action.yml
saudet: Yes, let's do that

pytorch/cppbuild.sh
saudet: Well, there's still an old version I suppose. Anyway, any harm in leaving this as is? Actually OpenMP is PITA, and I like to disable it everywhere I can. Does PyTorch now have it's own threading library that we can use instead of OpenMP?

HGuillemet: > OpenMP is PITA

Right, I spent a lot of time to make OpenMP work in all three platforms. But I think it's ok now. What is important is to prevent different OpenMP libraries to be linked to same executable, because this can silently cause incorrect computations, like what happened in #1503. MKL is supposed to detect this and throws an exception when loaded. I guess this hack making LLVM lib disguise as Intel lib is meant to prevent MKL to throw this exception. But then we cannot know which lib is actually used and I doubt MKL will work perfectly with llvm lib.

MKL adds more PITA, and I don't think we can ship a pytorch presets that can work with or without MKL.
In official binary, they chose to force MKL(link statically) on platforms where it's available, using intel omp shipped with MKL. Either we do something similar or we don't use MKL at all.

>  Does PyTorch now have it's own threading library that we can use instead of OpenMP?
No, it can use TBB instead of OpenMP, but I think TBB is more or less abandoned.
There is a native mechanism for inter-op threading in ATen  but it doesn't replace OpenMP for other threading tasks.

saudet: So, anyway this hack does work fine with MKL, so please revert this

HGuillemet: Why no doing something similar for other platforms then ?

saudet: GCC and VS don't use the same library so they don't conflict, or didn't used to anyway. That's the point here, please don't try fix what isn't broken. These lines aren't broken. Don't fix them

HGuillemet: On Linux, a user linking with MKL could set `MKL_THREADING_LAYER=GNU`. 
On Windows, I don't think MKL can use `vcomp` which is the OpenMP we currently linked with. So I don't know what happens if MKL is loaded. It will probably not use any threading layer at all.

I restored the hack. It was broken because the `install_name_tool -change` command supposed the installed path to `libomp` is `@rpath/libomp.dylib`, which is the case only if we run in the github runner where we changed the rpath. I fixed that. Also it didn't work on recent versions of MacOS because signature was incorrect after `install_name_tool`. I fixed that too.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
```diff
@@ -72,15 +94,8 @@ public void map(InfoMap infoMap) {
         torch.sharedMap(infoMap);
 
         infoMap
-            .put(new Info("basic/containers").cppTypes("c10::optional"))
-
             .put(new Info().enumerate().friendly())
             .put(new Info().javaText("import org.bytedeco.pytorch.*;"))
```
saudet, line 98: What about this one? Why do we need it?

HGuillemet: We don't anymore. I'll remove it.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch_cuda.java
```diff
@@ -72,15 +94,8 @@ public void map(InfoMap infoMap) {
...
-            .put(new Info().javaText("import org.bytedeco.pytorch.cuda.functions.*;"))
-            .put(new Info().javaText("import org.bytedeco.pytorch.Error;"))
-            .put(new Info().javaText("import org.bytedeco.pytorch.global.torch.DeviceType;"))
-            .put(new Info().javaText("import org.bytedeco.pytorch.global.torch.ScalarType;"))
-            .put(new Info().javaText("import org.bytedeco.pytorch.global.torch.MemoryFormat;"))
             .put(new Info().javaText("import org.bytedeco.pytorch.Allocator;"))
```
saudet, line 99: And this one too? Or is that required to override a conflict somewhere...

HGuillemet: Yes, this one is useful to disambiguate between `org.bytedeco.pytorch.Allocator` and `org.bytedeco.javacpp.annotation.Allocator`.

saudet:
Please upgrade the builds to CUDA 12.6, see commit https://github.com/bytedeco/javacpp-presets/commit/6853dc41a91cb46bc6c19484cae8f2a333cd8d7e

.github/workflows/pytorch.yml
```diff
@@ -33,16 +33,13 @@ jobs:
...
+    runs-on: macos-14
     steps:
-      - uses: bytedeco/javacpp-presets/.github/actions/deploy-macosx@actions
+      - uses: HGuillemet/javacpp-presets/.github/actions/deploy-macosx@hg_pytorch
   macosx-x86_64:
-    runs-on: macos-12
-#    strategy:
-#      matrix:
-#        ext: ["", -gpu]
+    runs-on: macos-13
```
saudet, line 40: Let's leave that to 12

HGuillemet: I had errors compiling Pytorch 2.4.0 with Apple Clang from macos-12 runner.
What is the problem with using the macos-13 runner ?

saudet: The problem is that it doesn't work either, right? I don't think the issue is with the version of macOS

HGuillemet: With macos-13, there was a little issue because, apparently, new versions of MacOS 13 provide Python 12 with symlink in /usr/local. And  Gradle from brew has a dependency towards Python 12, and triggers the creation of symlinks, clashing with the one already existing. It works by installing gradle with `--ignore-dependencies`. macos-12 runner also provides python3.12, so I think the workflows using macos-12 and needing Python or Gradle will still work but this will require verification. Or we can use macos-13 runner everywhere.

Pytorch compilation error on macos-12 is completly different. Something related to c++17 (`std::unordered_map`) support of the included Clang. See https://github.com/bytedeco/javacpp-presets/actions/runs/10197815123/job/28211441407.

I won't mind dropping macosx-x86_64 platform for Pytorch if this cannot be solved easily. I doubt many people will try to run Pytorch for Java on an Intel Mac nowadays.

saudet: It just sounds like `#include <unordered_map>`  and `#include <vector>` are missing from Handlers.cpp, let's just do that instead?

HGuillemet:
Pytorch uses cupti for profiling, but it's not provided by the cuda presets. Could we add it ?

saudet:
Sure, feel free to give it a try

HGuillemet:
Ready for me. We can look at MPI or UCC distributed backend later if needed.
Artifacts for this PR are available on usual snapshot repository  as `org.bytedeco:pytorch-platform-gpu:2.4.0-1.5.11-SNAPSHOT` or `org.bytedeco:pytorch-platform:2.4.0-1.5.11-SNAPSHOT`.

pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java
```diff
@@ -73,30 +79,51 @@
...
             },
             exclude = {"openblas_config.h", "cblas.h", "lapacke_config.h", "lapacke_mangling.h", "lapack.h", "lapacke.h", "lapacke_utils.h"},
-            link = {"c10", "torch_cpu", "torch"},
-            preload = {"gomp@.1", "iomp5", "omp", "tbb@.2", "asmjit", "fbgemm"}
+            link = { "c10", "torch", "torch_cpu" }
         ),
         @Platform(
             value = {"linux", "macosx", "windows"},
-            link = { "c10", "c10_cuda", "torch_cpu", "torch_cuda", "torch" },
-            preload = {"gomp@.1", "iomp5", "omp", "tbb@.2", "asmjit", "fbgemm", "cupti@.12"},
```
saudet, line 85: Please put back the preloads here and above

HGuillemet: preloads were moved to platform-specific sections, because the lib to preloads differ for each platform, in particular the openmp library. It's cleaner than trying to load everything for each platform. Is there a problem with that ? 
Do you spot a library that should be preloaded and is not anymore ?
gomp is not in the linux jar. 
iomp5 is for macos only
omp is not used by any platform anymore
tbb is not used
asmjit and fbgemm exist for windows only
cupti is now handled by inherited cuda preset

saudet: Add the missing one like gomp at least yes

HGuillemet: gomp added, but I don't think it has any effect since there is no jar containing libgomp.

saudet: If it doesn't get bundled, that's a bug

HGuillemet: It does, sorry. It didn't with my local build for some reason but it's bundled by the github runner.

saudet (approved):

PR #1511, opened 2024-06-25
[CPython] Add platform macosx-arm64

HGuillemet:
Add support of apple silicon mac to cpython presets.

There are constants in global/python.java that map platform-specific configurations like `HAVE_*`. They'd better be skipped.


saudet:
Please do the same with numpy and scipy

.github/actions/deploy-macosx/action.yml
```diff
@@ -108,6 +108,11 @@ runs:
         export DYLD_LIBRARY_PATH="$LIBRARY_PATH"
         export DYLD_FALLBACK_LIBRARY_PATH="$LIBRARY_PATH"
 
+        # Change default JVM on macos-14 runner
+        if [[ -n ${JAVA_HOME_17_arm64:-} ]]; then
+          export JAVA_HOME=$JAVA_HOME_17_arm64
+        fi
+
```
saudet, line 115: Java 8 or 11 isn't available??

HGuillemet:
> Please do the same with numpy and scipy

I'm waiting for the merge of this one since numpy depends on cpython

.github/actions/deploy-macosx/action.yml
HGuillemet: > Java 8 or 11 isn't available??

In the github runner, 11 is, 8 is not.
We can install 8 from brew.
But I don't see the interest.

saudet: I don't think Java 8 is available for arm64. Meh, Java 17 is fine I guess.

saudet (approved):

saudet:
This breaks the build for at least OpenCV:
https://github.com/bytedeco/javacpp-presets/actions/runs/9632851195/job/26858888577
Could you fix that before trying to get SciPy working?

HGuillemet:
> This breaks the build for at least OpenCV: https://github.com/bytedeco/javacpp-presets/actions/runs/9632851195/job/26858888577 Could you fix that before trying to get SciPy working?

The macosx-arm64 builds should now use the macos-14 runner, which is an arm64 machine.

HGuillemet:
Do you want a separate PR for this change in opencv workflow ?

saudet:
Yes, please!

PR #1515, opened 2024-06-28
[Numpy] Add platform macosx-arm64

HGuillemet:


.github/workflows/numpy.yml
```diff
@@ -48,10 +51,12 @@ jobs:
 #      - uses: bytedeco/javacpp-presets/.github/actions/deploy-windows@actions
   windows-x86_64:
     runs-on: windows-2019
+    env:
+      CI_DEPLOY_NEED_GCC: 1
```
saudet, line 55: I don't think it needs GCC on Windows either

HGuillemet: Ok trying...

saudet (approved):

PR #1516, opened 2024-06-30
[Scipy] Add platform macosx-arm64

HGuillemet:


.github/actions/deploy-macosx/action.yml
```diff
@@ -34,25 +34,27 @@ runs:
...
+            sudo install_name_tool -add_rpath /usr/local/lib/gcc/10/ -add_rpath @loader_path/. -id @rpath/libgcc_s.1.dylib /usr/local/lib/gcc/10/libgcc_s.1.dylib
+            sudo install_name_tool -change $(otool -L /usr/local/lib/gcc/10/libgfortran.5.dylib | grep /usr/local/Cellar/gcc@10/ | cut -f1 -d ' ') @rpath/libquadmath.0.dylib /usr/local/lib/gcc/10/libgfortran.5.dylib
+            sudo install_name_tool -change /usr/local/lib/gcc/10/libgcc_s.1.dylib @rpath/libgcc_s.1.dylib /usr/local/lib/gcc/10/libgomp.1.dylib
+            sudo install_name_tool -change /usr/local/lib/gcc/10/libgcc_s.1.dylib @rpath/libgcc_s.1.dylib /usr/local/lib/gcc/10/libstdc++.6.dylib
+            sudo install_name_tool -change /usr/local/lib/gcc/10/libgcc_s.1.dylib @rpath/libgcc_s.1.dylib /usr/local/lib/gcc/10/libgfortran.5.dylib
+            sudo install_name_tool -change /usr/local/lib/gcc/10/libgcc_s.1.dylib @rpath/libgcc_s.1.dylib /usr/local/lib/gcc/10/libquadmath.0.dylib
+            otool -L /usr/local/lib/gcc/10/*.dylib
+          fi
```
saudet, line 57: Please update all that for GCC 13

HGuillemet: Done. The "transitive" rpath is already ok for gcc@13, so commands like `install_name_tool -change /usr/local/lib/gcc/10/libgcc_s.1.dylib @rpath/libgcc_s.1.dylib` are not necessary. 
It would be easier if we used gcc@13 (or 14) on macosx-x86_64 too. Is it still necessary to stick with gcc@10 ?

saudet: > Done. The "transitive" rpath is already ok for gcc@13, so commands like install_name_tool -change /usr/local/lib/gcc/10/libgcc_s.1.dylib @rpath/libgcc_s.1.dylib are not necessary.

I'm pretty sure that's still going to be necessary, let's check the output of `otool -L`

> It would be easier if we used gcc@13 (or 14) on macosx-x86_64 too. Is it still necessary to stick with gcc@10 ?

Sure, would be a good time to upgrade to gcc@13 and fix all broken builds before the next release

HGuillemet: Do you prefer gcc@13, or the last default homebrew gcc (currently 14) ?

HGuillemet: > I'm pretty sure that's still going to be necessary, let's check the output of `otool -L`

Looks ok to me

saudet: Let's not do GCC 14, no

saudet (approved):

HGuillemet:
The `otool` command should be removed, or commented out, or moved in a `if DEPLOY_NEED_GCC` block.

PR #1522, opened 2024-07-31
Add `ClangMemoryMgmtExample` demonstrating how to avoid memory leaks

0x6675636b796f75676974687562:
The sample demonstrates how to avoid memory leaks when traversing the AST and tokenizing ranges.

For more context, see:

 - bytedeco/javacpp#767

### DIFF
--- a/llvm/samples/clang/ClangMemoryMgmtExample.java
+++ b/llvm/samples/clang/ClangMemoryMgmtExample.java
@@ -1,0 +1,229 @@
+import org.bytedeco.javacpp.BytePointer;
+import org.bytedeco.javacpp.Pointer;
+import org.bytedeco.javacpp.PointerPointer;
+import org.bytedeco.javacpp.PointerScope;
+import org.bytedeco.llvm.clang.CXClientData;
+import org.bytedeco.llvm.clang.CXCursor;
+import org.bytedeco.llvm.clang.CXCursorVisitor;
+import org.bytedeco.llvm.clang.CXIndex;
+import org.bytedeco.llvm.clang.CXSourceRange;
+import org.bytedeco.llvm.clang.CXString;
+import org.bytedeco.llvm.clang.CXToken;
+import org.bytedeco.llvm.clang.CXTranslationUnit;
+import org.bytedeco.llvm.clang.CXUnsavedFile;
+import org.bytedeco.llvm.global.clang;
+
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import java.util.stream.IntStream;
+
+import static java.lang.System.out;
+import static java.util.Arrays.asList;
+import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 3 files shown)


adjtime (approved):

0x6675636b796f75676974687562:
Samuel @saudet, the CI pipeline has succeeded, so please review.

llvm/samples/clang/ClangMemoryMgmtExample.java
```diff
@@ -0,0 +1,229 @@
...
+		Literal,
+
+		Comment,
+		;
+
+		private static TokenKind valueOf(final int ordinal) {
+			return values()[ordinal];
+		}
+	}
+}
```
saudet, line 229: Please convert tabs to spaces

llvm/samples/clang/pom.xml
```diff
@@ -16,6 +16,19 @@
...
+                <includes>
+                    <include>**/*.cc</include>
+                </includes>
+                <excludes>
+                    <exclude>**/*.java</exclude>
+                    <exclude>**/pom.xml</exclude>
+                    <exclude>target/**/*</exclude>
+                </excludes>
+            </resource>
+        </resources>
```
saudet, line 32: Why do we need this?

llvm/samples/clang/ClangMemoryMgmtExample.java
0x6675636b796f75676974687562: Done with https://github.com/bytedeco/javacpp-presets/pull/1522/commits/d903d6ba115f0ecd201f9da9e9024d5bfbc8da54.

I just followed the code style of the existing samples (which use tabs).

llvm/samples/clang/pom.xml
0x6675636b796f75676974687562:
```xml
        <resources>
            <resource>
                <directory>${project.basedir}</directory>
            </resource>
        </resources>
```

allows to load _Maven_ resources from the very same directory where Java sources reside (note that this project doesn't follow the standard project layout of `src/main/{java,resources}`). In other words, sample1.cc can now be opened not only as an external file, but also as a classpath resource (via `Class.getResource()`/`Class.getResourceAsStream()`).

The rest just adds inclusion/exclusion filters, so that Java sources and compiled classes are not treated as resources:

```xml
                <includes>
                    <include>**/*.cc</include>
                </includes>
                <excludes>
                    <exclude>**/*.java</exclude>
                    <exclude>**/pom.xml</exclude>
                    <exclude>target/**/*</exclude>
                </excludes>
```

Now, we can [load `sample1.cc` as a _classpath resource_](https://github.com/bytedeco/javacpp-presets/pull/1522/files#diff-f55f5480bb711e90a230d9805d56cad95cab99215c438b08c75722169439c098R78):

```java
        final URL codeUrl = requireNonNull(ClangMemoryMgmtExample.class.getResource("/sample1.cc"));
```

llvm/samples/clang/ClangMemoryMgmtExample.java
saudet: They do? Haven't notice...

saudet (approved):

PR #1524, opened 2024-08-03
Spinnaker: update to 4.0.0.116

jpsacha:
Note: 4.0.0.116 dropped support for windows-x86


saudet:
Looks like the new module name for the CLI of gdown is gdown, so please try to replace gdown.cli with gdown

jpsacha:
Something is wrong here. The workflow executed is not what is in the PR. It is still running Spinnaker 3 setup:
https://github.com/bytedeco/javacpp-presets/actions/runs/10225285866/job/28294117320?pr=1524#step:2:275

saudet:
Right, we need to change temporarily the actions in the workflow when we do that

jpsacha:
Looks that is working now

saudet (approved):

PR #1526, opened 2024-08-08
bump ModSecurity to 3.0.12

paul-hammant:
Newbie attempt at a dep upgrade  ModSecurity from 3.0.8 to 3.0.12


saudet:
Thanks! We'll also need to update the version in the pom.xml files to 1.5.11-SNAPSHOT and the GitHub Actions workflow to deploy-ubuntu, like other presets.

paul-hammant:
Those two done, and the start of a line in CHANGELOG

saudet:
Instead of trying to change everything at the same time, I'd recommend doing one small change at a time. As for debugging builds, we can do that with https://github.com/mxschmitt/action-tmate

paul-hammant:
I think it's time to admit I've made a mess of this, and should start over in the way you say

paul-hammant:
To be attempted again .. another day

PR #1531, opened 2024-08-19
[CUDA] Add cupti

HGuillemet:


HGuillemet:
I haven't added the new gen/ classes because I currently only have cuda 12.5 installed locally.

cuda/src/main/java/org/bytedeco/cuda/presets/cupti.java
```diff
@@ -0,0 +1,54 @@
...
+import org.bytedeco.javacpp.annotation.NoException;
+import org.bytedeco.javacpp.annotation.Platform;
+import org.bytedeco.javacpp.annotation.Properties;
+import org.bytedeco.javacpp.tools.Info;
+import org.bytedeco.javacpp.tools.InfoMap;
+import org.bytedeco.javacpp.tools.InfoMapper;
+
+/**
+ *
+ * @author Samuel Audet
```
saudet, line 34: Let's put your name here and above as well

cuda/src/main/java/org/bytedeco/cuda/presets/cupti.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+@Properties(inherit = {cudart.class}, value = {
+    @Platform(include = {"cupti_result.h", "cupti_version.h", "cupti_activity.h", "cupti_callbacks.h", "cupti_events.h", "cupti_metrics.h", "cupti_driver_cbid.h", "cupti_runtime_cbid.h", "cupti_nvtx_cbid.h"},
+              link = "cupti@.12"),
+    @Platform(value = "windows-x86_64", includepath = "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.6/extras/CUPTI/include/", linkpath = "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.6/extras/CUPTI/lib64/"),
```
saudet, line 39: There's a missing preload with the DLL name here

HGuillemet: Why is that needed when the library is already in link ? The dll will already be extracted by the loader into the cache because of the link. What do I miss ?

saudet: The name of the dll is different from the lib

cuda/src/main/java/org/bytedeco/cuda/presets/cupti.java
```diff
@@ -38,7 +38,8 @@
               link = "cupti@.12"),
     @Platform(value = "windows-x86_64", includepath = "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.6/extras/CUPTI/include/", linkpath = "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.6/extras/CUPTI/lib64/"),
     @Platform(value = {"linux-x86_64", "linux-arm64", "linux-ppc64le"}, includepath = {"/usr/local/cuda-12.6/extras/CUPTI/include/", "/usr/local/cuda/extras/CUPTI/include/"}, linkpath = {"/usr/local/cuda-12.6/extras/CUPTI/lib64/", "/usr/local/cuda/extras/CUPTI/lib64/"}),
-    @Platform(value = "macosx-x86_64", includepath = "/Developer/NVIDIA/CUDA-12.6/extras/CUPTI/include/", linkpath = "/Developer/NVIDIA/CUDA-12.6/extras/CUPTI/lib64/"),
+    @Platform(value = "macosx-x86_64", includepath = "/Developer/NVIDIA/CUDA-12.6/extras/CUPTI/include/", linkpath = "/Developer/NVIDIA/CUDA-12.6/extras/CUPTI/lib64/",
+              preload = "cupti64_2024.3.0", preloadpath = "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.6/lib64/"),
```
saudet, line 42: You'll need to add that above to the lines for Windows

HGuillemet: Oops. Fixed.

PR #1571, opened 2025-01-22
Enable non-standard OpenBLAS extensions

blueberry:
This contribution enables the operations that are present in OpenBLAS, but might not be directly supported by Accelerate and/or MKL. The previous behavior is preserved, and the new openblas_full class added, which extends the openblas class and enables the "extra" operations.


openblas/src/main/java/org/bytedeco/openblas/presets/openblas_full.java
```diff
@@ -0,0 +1,71 @@
...
+import org.bytedeco.javacpp.tools.InfoMap;
+
+/**
+ *
+ * @author Samuel Audet
+ * @author Dragan Djuric
+ */
+@Properties(inherit = openblas_nolapack.class, global = "org.bytedeco.openblas.global.openblas_full", value = {
+    @Platform(
+        include = {"openblas_config.h", "cblas.h", "lapacke_config.h", "lapacke_mangling.h", "lapack.h", "lapacke.h", "lapacke_utils.h"})})
```
saudet, line 42: Try to inherit from openblas.class instead so that we get the init() method and the include list too

openblas/src/main/java/org/bytedeco/openblas/presets/openblas_full.java
```diff
@@ -0,0 +1,71 @@
...
+            
+            // deprecated
+            "cgegs",   "cggsvd",  "ctzrqf",  "dgeqpf",  "dlatzm",  "sgelsx",  "slahrd",  "zgegv",   "zggsvp",
+            "cgegv",   "cggsvp",  "dgegs",   "dggsvd",  "dtzrqf",  "sgeqpf",  "slatzm",  "zgelsx",  "zlahrd",
+            "cgelsx",  "clahrd",  "dgegv",   "dggsvp",  "sgegs",   "sggsvd",  "stzrqf",  "zgeqpf",  "zlatzm",
+            "cgeqpf",  "clatzm",  "dgelsx",  "dlahrd",  "sgegv",   "sggsvp",  "zgegs",   "zggsvd",  "ztzrqf"};
+          
+        for (String f : functions) {
+            infoMap.put(new Info(f, "LAPACK_" + f, "LAPACK_" + f + "_base", "LAPACKE_" + f, "LAPACKE_" + f + "_work").skip());
+        }
```
saudet, line 68: Even if they are deprecated, we can still leave them around for simplicity?

In any case, let's take out the common code here into a new method.

openblas/src/main/java/org/bytedeco/openblas/presets/openblas_full.java
blueberry: Will do. This was ommision. Just to check: It is enough to only replace the reference at this exact place?

saudet: Should be, yes

openblas/src/main/java/org/bytedeco/openblas/presets/openblas_full.java
blueberry: done.

openblas/src/main/java/org/bytedeco/openblas/presets/openblas_full.java
blueberry: Thanks. Fixed.

openblas/src/main/java/org/bytedeco/openblas/presets/openblas_full.java
```diff
@@ -0,0 +1,48 @@
...
+import org.bytedeco.javacpp.tools.Info;
+import org.bytedeco.javacpp.tools.InfoMap;
+
+/**
+ *
+ * @author Samuel Audet
+ * @author Dragan Djuric
+ */
+@Properties(inherit = openblas.class, global = "org.bytedeco.openblas.global.openblas_full", value = {
+    @Platform(include = {"openblas_config.h", "cblas.h", "lapacke_config.h", "lapacke_mangling.h", "lapack.h", "lapacke.h", "lapacke_utils.h"})})
```
saudet, line 41: If the include list is the same as openblas.class, we probably don't need to add it here. Can you try to remove this line?

blueberry: Yes, of course. Sorry for these sloppy edits, I'm doing this through GitHub editor :) Update's comming in a minute.

blueberry: Fixed.

blueberry: Sorry, I misread that you mean about java imports. I'll be back in a minute :)

blueberry: Changed.

blueberry: Yes, yes, I know... Sorry. I hope the next edit will be ok :)

blueberry: Maybe the previous actions should be cancelled before they finish...

blueberry: Now the build works, but looks suspiciously quick in comparison to the last time it worked. As I don't know where to look for the resulting artifacts, I'll have to wait for your judgment.

saudet: Right, it doesn't work, because what inherit does is closer to a dependency declaration, so the build doesn't reparse the same files. Meh, the list is not long, so let's just leave it like you had it. Sorry

blueberry: I've re-included it. Fingers crossed :)

saudet (approved):

blueberry:
Is this ready to go now?

blueberry:
It seems that the refactoring to functionsToSkip() did not work as intended.
(copypasta from the issue discussion)
Unfortunately, it seems that the final versions with functionsToSkip() did not work as intended. I've included the last SNAPSHOT in a project; the openblas_full class has been generated, but it does not contain any of the "extra" methods that we supposed to "unskip".

Looking at the code, Java-wise, I can't see why openblas_full is not getting an empty string array to skip, but JavaCPP-wise, there might be a gotcha.

Is this worth pursuing, or should we revert to the first solution that repeats some code? (but who knows, maybe the same gotcha is relevant there?)

BTW, the openblas_full class repeats the methods that are already in the openblas class. I'm not sure is that the JavaCPP generator is supposed to do that when properly configured (judging by the openblas_nolapack and openblas, it is not...).

blueberry:
So, if we just removed the inherit value in the annotation, the result will be that openblas_full reuses the Java code, while being configured from scratch, independently from openblas_nolapack and openblas? It won't mess up some fragile mechanism that's supposed to weed out repetitions, take care of overrides and such?

saudet:
It's not as simple as that I'm sure, but someone had to try it. Do whatever it takes to make it work like you want it, that's fine!

blueberry:
Sure. Being a total noob, I would just like some pointers what would likely work, and what wouldn't, to avoid poking in the dark :)

saudet:
Actually, there's already a pattern like that here where a static mapCommon() method does the trick:
https://github.com/bytedeco/javacpp-presets/tree/master/cminpack/src/gen/java/org/bytedeco/cminpack/global

PR #1572, opened 2025-01-24
Enable non-standard OpenBLAS extensions v2

blueberry:
The second try of the #1571 
The macos build passed, so I'm opening the PR so you can see whether it now works with the changes you suggested (mapCommon).

I've tried to build everything on my machine, but the openblas build fails (independently of this change) due to some LLVM error that is probably related to some mismatch in I-don-t-know-which combination of native compilers, so I couldn't reach the Java gen stage and see how the final opencl_java class looks like. That's why I'm opening this PR. (btw I've tried the docker way too, as per JavaCPP guide, but then some internal docker stuff didn't work. I couldn't solve it right now).


openblas/src/main/java/org/bytedeco/openblas/presets/openblas_nolapack.java
```diff
@@ -136,7 +139,8 @@ public class openblas_nolapack implements LoadEnabled, InfoMapper {
     public String[] functionsToSkip() {
```
saudet, line 139: Please revert these changes that we don't need anymore.

blueberry: I left functionsToSikip() intentionally, as an option for future configuration. But, of course, if you think it's better to remove it, I'll do it.

blueberry: reverted.

saudet:
I think I got it working. Could you check that the functions you want are all there?

blueberry:
As soon as it appears in the snapshot repo (or is there a way to check right now?)

saudet:
https://github.com/bytedeco/javacpp-presets/pull/1572/files

blueberry:
I checked for a few of them and they are present. I'll test how they work when they appear in the snapshot repo, but I don't see why they wouldn't!. 

Thanks!

saudet:
Nope, it doesn't look like it's going to work that easily...

blueberry:
Why? (the failing build is due to a download issue, if you are referring o that)

blueberry:
I see what you mean now. 
However, it seems that only a small-ish subset of functions is not declared and is causing trouble.

As a first, step, maybe it would be good if we just skipped these functions.

As a second step, it seems that the cause is some misconfiguration between what should be prefixed with LAPACK_ vs LAPACKE_

Fox example, LAPACK_sgerfsx can be fount in Opelblas source, just as LAPACKE_sgerfsx, in connection with the header lapacke_utils.h, which is not included in JavaCPP OpenBLAS build. Maybe including this would be enough?

blueberry:
My latest version passes the build phase, I just don't know where to look for the generated Java class to check whether everything is really there.

saudet:
I've updated the files, but you could set CI_DEPLOY_USERNAME and  CI_DEPLOY_PASSWORD to get the files deployed to your Sonatype OSSRH account

blueberry:
Nope, the functions are not there. Maybe we should just duplicate the code, instead of inheriting, or we should have a plain base class with common methods, and then inherit independent classes openblas_nolapack and openblas_full from it?...
I don't have a clue what's going on in the internals to be able to fix this in an elegant way.

saudet:
You can do whatever you want, but your hack is hard to maintain, and you are not there to maintain it, I will simply disable it when it breaks on the next release.

blueberry:
I will maintain it, and I will try to improve it when I learn more JavaCPP (which I plan to do in a few months). But it even won't be a hack, it will just be a bit repetitive.

blueberry:
Of course, if you can offer some hints about what could be a cause of these methods not being in the generated code, that would help me a lot.

saudet:
If it's repetitive code, then just use methods and variables, etc. I will not merge anything that we can obviously simplify in the first place

saudet:
Start here: https://github.com/bytedeco/javacpp/wiki/Mapping-Recipes#skipping-lines-from-header-files

blueberry:
@saudet After a few weeks of being away from computer I returned to this, fixed my local system to be able to build this locally, and, after a bit of experimentation, I think I've finally come up with a straightforward solution that properly generates the missing functions.
I hope this will be good enough for merging!

openblas/pom.xml
```diff
@@ -31,11 +31,14 @@
...
       <plugin>
-        <artifactId>maven-compiler-plugin</artifactId>
+          <artifactId>maven-compiler-plugin</artifactId>
       </plugin>
       <plugin>
         <groupId>org.bytedeco</groupId>
         <artifactId>javacpp</artifactId>
+        <configuration>
+            <includePath>/usr/include/openblas/</includePath>
+        </configuration>
```
saudet, line 41: Please revert those changes

openblas/src/main/java/org/bytedeco/openblas/presets/openblas.java
```diff
@@ -60,4 +60,5 @@ public class openblas extends openblas_nolapack {
             }
         }
     }
+
```
saudet, line 63: This too

openblas/src/main/java/org/bytedeco/openblas/presets/openblas_nolapack.java
```diff
@@ -136,6 +136,22 @@ public class openblas_nolapack implements LoadEnabled, InfoMapper {
...
+            "cgesvdq", "dgesvdq", "sgesvdq", "zgesvdq", "clangb", "dlangb", "slangb", "zlangb",
+            "ctrsyl3", "dtrsyl3", "strsyl3", "ztrsyl3",
+            "sgedmd", "dgedmd", "cgedmd", "zgedmd", "sgedmdq", "dgedmdq", "cgedmdq", "zgedmdq",
+            // deprecated
+            "cggsvd", "dggsvd", "sggsvd", "zggsvd", "zggsvp", "cggsvp", "dggsvp", "sggsvp",
+            // extended
+            "cgbrfsx", "cporfsx", "dgerfsx", "sgbrfsx", "ssyrfsx", "zherfsx", "cgerfsx", "csyrfsx", "dporfsx", "sgerfsx", "zgbrfsx", "zporfsx",
+            "cherfsx", "dgbrfsx", "dsyrfsx", "sporfsx", "zgerfsx", "zsyrfsx", "cgbsvxx", "cposvxx", "dgesvxx", "sgbsvxx", "ssysvxx", "zhesvxx",
+            "cgesvxx", "csysvxx", "dposvxx", "sgesvxx", "zgbsvxx", "zposvxx", "chesvxx", "dgbsvxx", "dsysvxx", "sposvxx", "zgesvxx", "zsysvxx"};
+
```
saudet, line 150: Most of those functions are probably available

blueberry: They are available only with the LAPACKE_ prefix, so I enabled these versions.

openblas/src/main/java/org/bytedeco/openblas/presets/openblas.java
blueberry: done.

openblas/pom.xml
blueberry: Reverted.

openblas/src/main/java/org/bytedeco/openblas/presets/openblas_nolapack.java
blueberry: These functions fail on android. Should we exclude them completely (as it was) or there is a way to make this part of the generator conditional on the operating system?
I'm fine with either, but you'll have to decide what is best aligned with javacpp-presets goals.

saudet: It's not like they are not failing on Linux or Mac either, these functions do not exist at all. Linux and Mac simply allow to link against undefined functions, but Android and Windows do not, that's all. There is no reason to leave them there though, so please remove them.

blueberry: I've done this, but the error log identified a handful of other troublemakers from the "extended" sections, so I skipped them too. Fingers crossed...

blueberry:
Now the logical part of the code seems to be working everywhere, but the build fails on Windows for the misterious reason, after 6 hours. Here's the relevant part at the end of the log. I don't have a windows machine locally where I can experiment with this. Moreover, I didn't managed to build openblas itself on my linux machine. I managed to use the openblas installed by Arch linux. Any ideas what might be the stopper?

===============
2025-02-19T04:03:58.4572042Z [INFO] --- javacpp:1.5.12-SNAPSHOT:build (javacpp-cppbuild-clean) @ openblas ---
2025-02-19T04:03:58.4592803Z [INFO] Detected platform "windows-x86_64"
2025-02-19T04:03:58.4606695Z [INFO] Building platform "windows-x86_64"
2025-02-19T04:03:58.4613408Z [INFO] bash cppbuild.sh clean openblas 
2025-02-19T04:03:58.6310128Z Detected platform "windows-x86_64"
2025-02-19T04:03:58.6328578Z Building platform "windows-x86_64"
2025-02-19T04:03:58.6418238Z Cleaning "openblas"
2025-02-19T04:03:58.7095979Z In file included from generic/ztrmm_uncopy_4.c:40:
2025-02-19T04:03:58.7102908Z ../common.h:423:10: fatal error: common_x86_64.h: No such file or directory
2025-02-19T04:03:58.7396978Z compilation terminated.
2025-02-19T04:03:58.7437143Z make[1]: *** [Makefile.L3:1826: ztrmm_ounncopy_SANDYBRIDGE.obj] Error 1
2025-02-19T04:03:58.7438934Z make[1]: *** Waiting for unfinished jobs....
2025-02-19T04:03:59.5684628Z 2025-02-19 04:03:59
2025-02-19T04:03:59.5689911Z Full thread dump OpenJDK 64-Bit Server VM (25.442-b06 mixed mode):
====================

blueberry:
@saudet Maybe you'd prefer to disable automatic build runs on this PR until I figure it out.

saudet:
I'm not sure what you're trying to do, bu the build on Windows is failing because of these 2 functions, not cblas or lapack:
```
2025-02-19T19:21:22.2210191Z jniopenblas_full.obj : error LNK2001: unresolved external symbol openblas_set_num_threads_local
2025-02-19T19:21:22.2376894Z jniopenblas_full.obj : error LNK2001: unresolved external symbol openblas_set_threads_callback_function
2025-02-19T19:21:22.2377579Z jniopenblas_full.dll : fatal error LNK1120: 2 unresolved externals
2025-02-19T19:21:22.2746229Z [INFO] --------------------------------------------------------------------
```

blueberry:
> I'm not sure what you're trying to do, bu the build on Windows is failing because of these 2 functions, not cblas or lapack:
> 
> ```
> 2025-02-19T19:21:22.2210191Z jniopenblas_full.obj : error LNK2001: unresolved external symbol openblas_set_num_threads_local
> 2025-02-19T19:21:22.2376894Z jniopenblas_full.obj : error LNK2001: unresolved external symbol openblas_set_threads_callback_function
> 2025-02-19T19:21:22.2377579Z jniopenblas_full.dll : fatal error LNK1120: 2 unresolved externals
> 2025-02-19T19:21:22.2746229Z [INFO] --------------------------------------------------------------------
> ```

Thank for that pointer! I was not able to find the exact function that was the culprit, so I was trying to find that out by guesses. I'll remove these two and then I hope it will work!

blueberry:
@saudet Good news! The last commit builds on all platforms!

I can see that a few builds (windows is not among them) are still waiting in the queue due to previous actions still running (if you can, maybe you'd want to cancel them anyway), but on my repository, all build actions completed successfully.

Thanks for helping with this!

openblas/pom.xml
```diff
@@ -31,7 +31,7 @@
         <artifactId>maven-resources-plugin</artifactId>
       </plugin>
       <plugin>
-        <artifactId>maven-compiler-plugin</artifactId>
+          <artifactId>maven-compiler-plugin</artifactId>
```
saudet, line 34: Please revert

openblas/src/main/java/org/bytedeco/openblas/presets/openblas_nolapack.java
```diff
@@ -136,43 +137,79 @@ public class openblas_nolapack implements LoadEnabled, InfoMapper {
...
-            "cgeqpf",  "clatzm",  "dgelsx",  "dlahrd",  "sgegv",   "sggsvp",  "zgegs",   "zggsvd",  "ztzrqf",
-            // extended
-            "cblas_sbstobf16", "cblas_sbdtobf16", "cblas_sbf16tos", "cblas_dbf16tod", "cblas_sbdot", "cblas_sbgemv", "cblas_sbgemm", "cblas_sbgemm_batch",
-            "cgbrfsx", "cporfsx", "dgerfsx", "sgbrfsx", "ssyrfsx", "zherfsx", "cgerfsx", "csyrfsx", "dporfsx", "sgerfsx", "zgbrfsx", "zporfsx",
-            "cherfsx", "dgbrfsx", "dsyrfsx", "sporfsx", "zgerfsx", "zsyrfsx", "cgbsvxx", "cposvxx", "dgesvxx", "sgbsvxx", "ssysvxx", "zhesvxx",
-            "cgesvxx", "csysvxx", "dposvxx", "sgesvxx", "zgbsvxx", "zposvxx", "chesvxx", "dgbsvxx", "dsysvxx", "sposvxx", "zgesvxx", "zsysvxx"};
+            "cgegs", "cggsvd", "ctzrqf", "dgeqpf", "dlatzm", "sgelsx", "slahrd", "zgegv", "zggsvp",
+            "cgegv", "cggsvp", "dgegs", "dggsvd", "dtzrqf", "sgeqpf", "slatzm", "zgelsx", "zlahrd",
+            "cgelsx", "clahrd", "dgegv", "dggsvp", "sgegs", "sggsvd", "stzrqf", "zgeqpf", "zlatzm",
+            "cgeqpf", "clatzm", "dgelsx", "dlahrd", "sgegv", "sggsvp", "zgegs", "zggsvd", "ztzrqf"
```
saudet, line 198: Please revert all unnecessary formatting changes. It makes it hard to compare what actually changed

openblas/pom.xml
blueberry: reverted.

openblas/src/main/java/org/bytedeco/openblas/presets/openblas_nolapack.java
blueberry: I've rewritten it now to match the original one as much as possible.

saudet:
Looks alright, but could you double check that all the functions you're interested in are there?

blueberry:
Thanks! I've checked several and they are in the generated java file for openblas_full.

PR #1599, opened 2025-03-11
Add pytorch-vision

ivolucas:
Module for adding the ability to run on pytorch nn models that use operation implemented on torch vision.

### DIFF
--- a/pytorch-vision/LICENSE
+++ b/pytorch-vision/LICENSE
@@ -1,0 +1,73 @@
+From PyTorch:
+
+Copyright (c) 2016-     Facebook, Inc            (Adam Paszke)
+Copyright (c) 2014-     Facebook, Inc            (Soumith Chintala)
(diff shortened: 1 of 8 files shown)


saudet:
This looks like something that we could add to the presets for PyTorch?

ivolucas:
You are suggesting to be included in the normal build of pytorch presets?

Yes I can do that.

saudet:
Yes, that's easier to maintain, but it looks like this only works on Linux? There is no support for Mac or Windows? Let me think about what we should do with this...

ivolucas:
I didn't test on other OS. Because of that, I didn't want to add code that
I wasn't sure it worked.
Hopping on a windows and/or Mac developer to add it later on :)

A quarta, 12/03/2025, 00:26, Samuel Audet ***@***.***>
escreveu:

saudet:
Right, we can add that as a Linux-only module to the presets for PyTorch for now, that's fine I think

Issue #1604, opened 2025-03-16
OpenCV JPEG SIMD off


fischermario:
Using OpenCV's `imdecode` is rather slow. I found out that SIMD is intentionally turned off by setting

`-DENABLE_LIBJPEG_TURBO_SIMD=OFF`

in

[cppbuild.sh](https://github.com/bytedeco/javacpp-presets/blob/master/opencv/cppbuild.sh#L129)

which leads to the following output of `Core.getBuildInformation()`:

```
  Media I/O: 
  ...
    JPEG:                        build-libjpeg-turbo (ver 3.0.3-70)
      SIMD Support Request:      NO
```

Btw. it is available for PNG.

Is there a specific reason for turning SIMD off?

saudet:
Yes, it was buggy, but maybe it's not anymore. Please open a pull request

fischermario:
Please see #1605 and give it a try.

Sometimes the option `-DWITH_SIMD=ON` may be necessary too.

Thank you!

saudet:
Fixed in pull #1651. Enjoy!

PR #1605, opened 2025-03-16
OpenCV: SIMD on for libjpeg-turbo

fischermario:
As requested in #1604


saudet:
Please fix the build

saudet:
Using native linux-arm64 runners instead of cross compiling might help: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

saudet:
Duplicate of #1651

PR #1609, opened 2025-03-25
Use ARM runners for CPython and Numpy

AutonomicPerfectionist:
This PR enables ARM builds for CPython and Numpy by using the Github Actions ARM runners. It also makes minor changes to the `cppbuild.sh` scripts for both to fix compilation issues on ARM. For numpy, this entailed removing the workaround implemented in #1386.

### Note
I made the minimal changes necessary to fix ARM compilation. This means that the cpython build still essentially does cross-compilation to target ARM even when running on ARM hosts. It may be desirable to add a check for whether the target and host platforms match to improve build times.

The cpp build was verified to succeed on a Raspberry Pi 4, I have not yet verified full end-to-end success via maven due to a missing dependency on the pi.

### DIFF
--- a/.github/actions/deploy-ubuntu/action.yml
+++ b/.github/actions/deploy-ubuntu/action.yml
@@ -78,7 +78,9 @@ runs:
         source /etc/os-release
         export CODENAME=$UBUNTU_CODENAME
 
-        if [[ ! "$ARCH" == "amd64" ]]; then
+        # arm64 runners don't work with apt-spy2,
+        # the selected mirrors don't have arm64 binaries
+        if [[ ! "$ARCH" == "amd64" && ! "$ARCH" == "arm64" ]]; then
           # https://github.com/actions/runner-images/issues/675
           $SUDO gem install apt-spy2
           $SUDO apt-spy2 check
--- a/.github/actions/deploy-ubuntu/action.yml
+++ b/.github/actions/deploy-ubuntu/action.yml
@@ -20,12 +20,14 @@ runs:
         $SUDO rm -rf /usr/local/lib/android
         $SUDO rm -rf /usr/share/dotnet
 
-        # Allocate a swapfile on Linux as it's not enabled by default. Needed for pytorch and mkl.
-        $SUDO fallocate -l 4GB /swapfile
-        $SUDO chmod 600 /swapfile
-        $SUDO mkswap /swapfile
-        $SUDO swapon /swapfile || true
-
+        # fallocate fails on arm64 runners
+        if [[ ! "$CI_DEPLOY_PLATFORM" == "linux-arm64" ]]; then
+          # Allocate a swapfile on Linux as it's not enabled by default. Needed for pytorch and mkl.
+          $SUDO fallocate -l 4GB /swapfile
+          $SUDO chmod 600 /swapfile
+          $SUDO mkswap /swapfile
+          $SUDO swapon /swapfile || true
+        fi
         mkdir -p .ccache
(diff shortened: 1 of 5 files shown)


saudet:
Sounds good, but why are the CI actions not running? 🤔

saudet:
Ok, the formatting of the workflows was incorrect. Now it still fails, so anyway please fix it!

AutonomicPerfectionist:
The numpy workflow seems to have failed because the ubuntu image has the wrong sources.list for apt, ARM64 packages are hosted at `http://ports.ubuntu.com/ubuntu-ports`, not `http://archive.ubuntu.com/ubuntu`. The cpython workflow is just showing me an fallocate error, which shouldn't have anything to do with us and is probably a transient error.

AutonomicPerfectionist:
It seems the github action tries to workaround an Actions issue by selecting a mirror, but the mirror selected is the `archive.ubuntu.com` one and not the `ports.ubuntu.com`:

```
Mirror: http://archive.ubuntu.com/ubuntu/ - UP
The closest mirror is: http://archive.ubuntu.com/ubuntu/
Updating /etc/apt/sources.list
Updated '/etc/apt/sources.list' with http://archive.ubuntu.com/ubuntu/
```

AutonomicPerfectionist:
Appears to be a problem with `apt-spy2`, the lines in the action that appear to select a mirror use it:
https://github.com/bytedeco/javacpp-presets/blob/754110de6ee2ae3f9226de7c6a5c0595c32cbfb1/.github/actions/deploy-ubuntu/action.yml#L84-L85

saudet:
Ok, so could you try to fix it?

saudet:
If I remember correctly, the checks for "amd64" is to detect cross compilation, so we should be able to skip all that when running actually running on ARM...

AutonomicPerfectionist:
I've attempted a fix by simply disabling mirror selection for arm64 runners

AutonomicPerfectionist:
Hmm, both runs failed again, with an error on fallocate... I do see that the fallocate command is being used to create a 4GB swap file, but I can't imagine the arm64 runners would be so resource constrained that they can't allocate an additional 4GB on the filesystem...

https://github.com/bytedeco/javacpp-presets/blob/754110de6ee2ae3f9226de7c6a5c0595c32cbfb1/.github/actions/deploy-ubuntu/action.yml#L24

I've pushed a new commit that just disables the fallocate call, but it didn't seem to change anything. I'll revert that commit and continue looking for the cause of the error

saudet:
To test the actions, you'll need to change temporarily the workflows to point to your branch

.github/actions/deploy-ubuntu/action.yml
```diff
@@ -78,7 +80,9 @@ runs:
         source /etc/os-release
         export CODENAME=$UBUNTU_CODENAME
 
-        if [[ ! "$ARCH" == "amd64" ]]; then
+        # arm64 runners don't work with apt-spy2,
+        # the selected mirrors don't have arm64 binaries
+        if [[ ! "$ARCH" == "amd64" && ! "$ARCH" == "arm64" ]]; then
```
saudet, line 85: We need this when cross compiling though. Please disable it only when we're running on ARM

AutonomicPerfectionist: Will do

AutonomicPerfectionist:
> To test the actions, you'll need to change temporarily the workflows to point to your branch

Yeah I figured that out, been playing whackamole with some more issues on a different branch of mine. Slowly getting there

saudet:
Duplicate of #1652

PR #1637, opened 2025-06-16
MKL 2025.0 build for JavaCPP 1.5.12

blueberry:
I've just updated the MKL build for 2025.0 (the same as it was in JavaCPP 1.5.11) to fit into the rest of javacpp-presets which are nearing the 1.5.12 release. 

I did a local build on my Arch Linux machine, and it went without any problems. (The Github Actions failed quickly due to, it was reported, a "outdated Ubuntu version"). So I hope it's a matter of updating the setup to a more recent stack (I don't know where to change this).

Windows build in GitHub Actions succeeded without a hitch.

I hope this could be included in JavaCPP 1.5.12, since MKL is not just a very featureful BLAS/LAPACK implementation, but also comes with DSP, sparse matrices, statistics, and load of other operations useful in number crunching, which are missing from the alternatives such as OpenBLAS.

### DIFF
--- a/mkl/platform/pom.xml
+++ b/mkl/platform/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.bytedeco</groupId>
     <artifactId>javacpp-presets</artifactId>
-    <version>1.5.11</version>
+    <version>1.5.12-SNAPSHOT</version>
     <relativePath>../../</relativePath>
   </parent>
 
--- a/mkl/platform/redist/pom.xml
+++ b/mkl/platform/redist/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.bytedeco</groupId>
     <artifactId>javacpp-presets</artifactId>
-    <version>1.5.11</version>
+    <version>1.5.12-SNAPSHOT</version>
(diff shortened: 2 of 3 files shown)


saudet:
There is no point in just changing the version of the presets. If you want to update the version of MKL though, please do it.

blueberry:
I'll see whether I can update it, but I think even the version of the presets have some value, since mixing javacpp versions gives a warning (that could be safely ignored for now, but confuses people nevertheless).

blueberry:
I see now that this was indeed built with a slightly newer version of MKL 2025.0.1, which is current on Arch Linux. When the Arch Linux repositories upgrade to 2025.1, I'll build it with then current JavaCPP, and create a PR for, I suppose, future 1.5.13-SNAPSHOT.

After that, I hope oneMath will soon become a viable alternative (it seems to be Intel's rebranding of MKL, and a wrapper that supports many backends, similarly to how oneDNN works)...

saudet:
They've made a second release of MKL this year, so it doesn't look totally dead just yet.
Anyway, I've released 2025.2-1.5.12. Enjoy!

blueberry:
Thanks!

Issue #1650, opened 2025-07-21
arm64 build of opencv-gpu


barrypitman:
I'm interested in building an arm64 build of the opencv cuda (GpuMat) support. I see that currently there is only support for linux-x86_64 and windows-x86_64. I would use this on a Jetson. 

My [(naive first attempt)](https://github.com/barrypitman/javacpp-presets/commit/f92b97dbb800a73c90a34fa815372f1fed7b5483) failed and I see this in the [logs:](https://github.com/barrypitman/javacpp-presets/actions/runs/16397984775/job/46333225391)

```
/usr/local/cuda/bin/nvcc: 1: ELF�0B@��O@8: not found
/usr/local/cuda/bin/nvcc: 2: Syntax error: word unexpected (expecting ")")
```
I suppose that its mixing up x86 with arm64 binaries. 

Do you have any suggestions or pointers for this, maybe this type of cross-compilation with CUDA has been tackled elsewhere in the project?

Thanks,

saudet:
It seems to work fine when building on an ARM machine, but I'm not aware of anyone attempting to cross compile:
https://github.com/search?q=org%3Abytedeco+linux-arm64+opencv+cuda&type=issues

saudet:
BTW, now that GitHub Actions has free ARM runners, it should be easy to add a linux-arm64-gpu build to the presets.

PR #1651, opened 2025-07-22
Build opencv-gpu for arm64

barrypitman:
relates to #1650 . I've made some changes to the deploy-ubuntu action, so that it can run on an arm64 runner. But when the opencv cppbuild.sh script runs, it runs into an error relating to the python build target:

```
CMake Error at modules/python/common.cmake:191 (install):
  install TARGETS given no LIBRARY DESTINATION for module target
  "opencv_python3".
Call Stack (most recent call first):
  modules/python/python3/CMakeLists.txt:11 (include)
```
I think that this is in related to cpython. In my local testing, I had to build cpython module first, but I'm not sure how to solve this in github actions.


barrypitman:
My build logs are here - https://github.com/barrypitman/javacpp-presets/actions/runs/16437328793/job/46449841489 . Both the -gpu and normal opencv builds are failing at the same place

saudet:
Right, CPython, ideally we'd have to finish pull #1609 first, but you can start by modifying the build script to use the version of CPython installed on the system.

barrypitman:
I ran into some issues with updating the script to use the system-installed python (related to numpy), so i've submitted a separate pull request  #1652 to run cpython and numpy workflows on arm runners.

barrypitman:
I've updated this PR to enable the opencv-gpu build on linux-arm64. It also enables "NEON" optimisations

saudet (approved):

PR #1652, opened 2025-07-22
build cpython and numpy on arm github actions runners

barrypitman:
The numpy build is failing, but i think its just because the cpython linux-arm64 isn't published yet:

Could not find artifact org.bytedeco:cpython:jar:linux-arm64:3.13.5-1.5.13-SNAPSHOT in central-portal-snapshots


.github/actions/deploy-ubuntu/action.yml
```diff
@@ -20,12 +20,13 @@ runs:
         $SUDO rm -rf /usr/local/lib/android
         $SUDO rm -rf /usr/share/dotnet
 
-        # Allocate a swapfile on Linux as it's not enabled by default. Needed for pytorch and mkl.
-        $SUDO fallocate -l 4GB /swapfile
-        $SUDO chmod 600 /swapfile
-        $SUDO mkswap /swapfile
-        if ! $SUDO swapon /swapfile; then
-          $SUDO rm /swapfile
+        # fallocate fails on arm64 runners
```
saudet, line 23: What kind of error are you getting?

.github/actions/deploy-ubuntu/action.yml
```diff
@@ -80,8 +81,9 @@ runs:
         $SUDO apt-get -y install gnupg
         source /etc/os-release
         export CODENAME=$UBUNTU_CODENAME
-
-        if [[ ! "$ARCH" == "amd64" ]]; then
+        # arm64 runners don't work with apt-spy2,
+        # the selected mirrors don't have arm64 binaries
+        if [[ ! "$ARCH" == "amd64" && ! "$ARCH" == "arm64" ]]; then
```
saudet, line 86: This needs to be enabled for cross-compilation though, so disable it only when it's on an ARM runner

.github/actions/deploy-ubuntu/action.yml
```diff
@@ -95,28 +97,34 @@ runs:
...
         fi
-
-        if [[ "$ARCH" == "i386" ]]; then
+        
+        HOST_ARCH=$(dpkg --print-architecture)   # amd64 or arm64
+        if [[ "$ARCH" != "$HOST_ARCH" ]]; then
+          # we are cross-compiling
           $SUDO dpkg --add-architecture $ARCH
-          TOOLCHAIN="gcc-$PREFIX g++-$PREFIX gfortran-$PREFIX"
-        elif [[ ! "$ARCH" == "amd64" ]]; then
```
saudet, line 102: That's not correct, try to replace `"$ARCH" == "amd64"` with `"$ARCH" != "$HOST_ARCH"` instead

.github/actions/deploy-ubuntu/action.yml
barrypitman: I took some of the code from the previous PR. If we try to run that on the arm runner, we get `fallocate: fallocate failed: Text file busy`. That might indicate that /swapfile is already being used. Maybe we just need to change the path?

.github/actions/deploy-ubuntu/action.yml
barrypitman: done, replaced the guard with `if [[ "$ARCH" != "arm64" ]]; then`

.github/actions/deploy-ubuntu/action.yml
barrypitman: the highlighted code is already inside a `"$ARCH" != "$HOST_ARCH"` block. Here is the full thing:

```
HOST_ARCH=$(dpkg --print-architecture)   # amd64 or arm64
if [[ "$ARCH" != "$HOST_ARCH" ]]; then
  # we are cross-compiling
  $SUDO dpkg --add-architecture $ARCH
  if [[ "$ARCH" == "i386" ]]; then
    TOOLCHAIN="gcc-$PREFIX g++-$PREFIX gfortran-$PREFIX"
  elif [[ "$ARCH" == "amd64" ]]; then
    $SUDO sed -i 's/deb http/deb [arch=amd64] http/g' /etc/apt/sources.list
    $SUDO echo deb [arch=$ARCH] http://ports.ubuntu.com/ubuntu-ports $CODENAME main restricted universe multiverse | $SUDO tee -a /etc/apt/sources.list
    $SUDO echo deb [arch=$ARCH] http://ports.ubuntu.com/ubuntu-ports $CODENAME-updates main restricted universe multiverse | $SUDO tee -a /etc/apt/sources.list
    $SUDO echo deb [arch=$ARCH] http://ports.ubuntu.com/ubuntu-ports $CODENAME-backports main restricted universe multiverse | $SUDO tee -a /etc/apt/sources.list
    $SUDO echo deb [arch=$ARCH] http://ports.ubuntu.com/ubuntu-ports $CODENAME-security main restricted universe multiverse | $SUDO tee -a /etc/apt/sources.list
    TOOLCHAIN="gcc-$PREFIX g++-$PREFIX gfortran-$PREFIX linux-libc-dev-$ARCH-cross binutils-multiarch"
  fi 
fi
        ``

File: .github/actions/deploy-ubuntu/action.yml (hunk 1 above)
Comment (saudet) [reply .github/actions/deploy-ubuntu/action.yml]: Maybe we just need to call swapoff before?

Comment (barrypitman) [reply .github/actions/deploy-ubuntu/action.yml]: OK, I've added that, and its working on the arm runner now

**Comment by saudet** [comment]
I've uploaded snapshots for CPython, but NumPy's build still fails

File: .github/actions/deploy-ubuntu/action.yml [hunk 4]
```diff
@@ -95,28 +103,33 @@ runs:
...
+          fi 
         fi
         # $SUDO echo deb [arch=amd64] http://ppa.launchpad.net/openjdk-r/ppa/ubuntu $CODENAME main | $SUDO tee -a /etc/apt/sources.list
         # $SUDO echo deb [arch=amd64] http://ppa.launchpad.net/deadsnakes/ppa/ubuntu $CODENAME main | $SUDO tee -a /etc/apt/sources.list
         # $SUDO apt-key adv --keyserver keyserver.ubuntu.com --recv-keys EB9B1D8886F44E2A
         # $SUDO apt-key adv --keyserver keyserver.ubuntu.com --recv-keys BA6932366A755776
         $SUDO apt-get update
-        $SUDO apt-get -y install gcc-multilib g++-multilib gfortran-multilib python3 python3-minimal rpm libasound2-dev:$ARCH freeglut3-dev:$ARCH libfontconfig-dev:$ARCH libgtk2.0-dev:$ARCH libusb-dev:$ARCH libusb-1.0-0-dev:$ARCH libffi-dev:$ARCH libbz2-dev:$ARCH zlib1g-dev:$ARCH libxcb1-dev:$ARCH libpulse-dev:$ARCH libva-dev:$ARCH libdrm-dev:$ARCH
+        if [[ "$HOST_ARCH" == "amd64" ]]; then
+          $SUDO apt-get -y install gcc-multilib g++-multilib gfortran-multilib
```
saudet, line 127: Let's move those to TOOLCHAIN with some default that works for both arm64 and amd64

saudet:
We'll also need to update the parent pom.xml and platform/pom.xml files

barrypitman:
In terms of updating the parent pom.xml, I'm not sure what to change. Under the javacpp-platform-linux-arm64 profile, cpython and numpy are already listed. And the platform/pom.xml, also not sure what's required there. This pull request deals only with enabling cpython and numpy, and using arm runners. Previously those builds weren't enabled for linux-arm64.

numpy/cppbuild.sh
```diff
@@ -85,6 +85,11 @@ fi
 export PYTHONPATH="$PYTHON_INSTALL_PATH"
 mkdir -p "$PYTHON_INSTALL_PATH"
 
+# Install pip for the cross-compiled interpreter
+if ! "$PYTHON_BIN_PATH" -m pip --version >/dev/null 2>&1; then
+    "$PYTHON_BIN_PATH" -m ensurepip --upgrade
+fi
+
```
saudet, line 92: The missing pip in the CPython package is probably caused  by getting it built in cross-compilation mode. Try to update the linux-arm64 as per the old linux-armhf patterns in cppbuild.sh to build in native mode

barrypitman: thanks, I'll test that out locally first

.github/actions/deploy-ubuntu/action.yml
barrypitman: I've cleaned this up a bit by removing the multilib packages and including default TOOLCHAIN packages, trying to keep the overall logic intact. It relies on the runner image already having the gcc, g++ and gfortran packages installed by default, i.e. the TOOLCHAIN is only for cross-compiling

barrypitman:
OK, numpy and cpython builds are both passing on the arm runners. I tried to retain the support for cross-compiling those from x86_64, so the cppbuild.sh scripts have branches like this:
 
```
linux-arm64)
  HOST_ARCH="$(uname -m)"
  if [[ $HOST_ARCH == "aarch64" ]]; then
    echo "Detected arm arch so not cross compiling";
    ...
  else
   echo "Detected non arm arch so cross compiling";
   ...
  fi
  ;;
```

numpy/cppbuild.sh
saudet: Can we remove this now?

barrypitman: It seems like it is still required. When I remove it, we get the original error - 
`/home/runner/.javacpp/cache/cpython-3.13.5-1.5.13-SNAPSHOT-linux-arm64.jar/org/bytedeco/cpython/linux-arm64/bin/python3.13: No module named pip`
I assume its now picking up the natively compiled cpython snapshot that ran [6 hours ago](https://github.com/bytedeco/javacpp-presets/actions/runs/16479122470/job/46603356679#step:2:9278) -

saudet:
These are the platform/pom.xml files that need updating:
https://github.com/bytedeco/javacpp-presets/blob/master/cpython/platform/pom.xml
https://github.com/bytedeco/javacpp-presets/blob/master/numpy/platform/pom.xml

barrypitman:
Ah, right, thanks. I've enabled the linux-arm64 dependencies in those poms.

saudet:
There's a couple of lines to update in those files

barrypitman:
sorry, missed those. Updated now

saudet (approved):

Issue #813, opened 2019-11-18
Dynamic binary analysis with Java


eix128:
Hi , i have seen a very good framework named
Panda Qemu
With this framework we can make binary analysis.
Catching system calls.
But all framework plugins can only be written by C++ and phyton
There is no java support.
It would be good to add to javacpp for reverse engineering all the things with Java
https://github.com/panda-re/panda
https://adalogics.com/blog/Building-a-custom-malware-sandbox-with-PANDA-Part-1
With this framework on 
https://vimeo.com/113442048

PR #1653, opened 2025-07-23
Add Windows ARM64 support for windows deploy action

devjeonghwan:
Related to JavaCPP's [PR #813](https://github.com/bytedeco/javacpp/pull/813)

### DIFF
--- a/.github/actions/deploy-windows/action.yml
+++ b/.github/actions/deploy-windows/action.yml
@@ -8,13 +8,12 @@ runs:
         GITHUB_EVENT_HEAD_COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
       run: |
         cd /d %USERPROFILE%
-        
+
         echo Installing MSYS2
-        
         if "%CI_DEPLOY_PLATFORM%"=="windows-arm64" (
           curl -Lo msys2.exe https://github.com/msys2/msys2-installer/releases/download/2025-06-22/msys2-base-x86_64-20250622.sfx.exe
           msys2.exe -oC:\ -y
-        
+
           C:\msys64\usr\bin\bash -lc "pacman -S --needed --noconfirm pkg-config"
           C:\msys64\usr\bin\bash -lc "pacman -S --needed --noconfirm base-devel git tar unzip p7zip zip autoconf autoconf-archive automake libtool make patch gnupg"
           C:\msys64\usr\bin\bash -lc "pacman -S --needed --noconfirm mingw-w64-clang-aarch64-nasm mingw-w64-clang-aarch64-clang mingw-w64-clang-aarch64-libtool mingw-w64-clang-aarch64-gcc-compat mingw-w64-clang-aarch64-libwinpthread mingw-w64-clang-aarch64-SDL2 mingw-w64-clang-aarch64-ragel mingw-w64-clang-aarch64-vulkan-headers mingw-w64-clang-aarch64-vulkan-loader"
@@ -25,11 +24,11 @@ runs:
           C:\msys64\usr\bin\bash -lc "pacman -S --needed --noconfirm mingw-w64-x86_64-nasm mingw-w64-x86_64-toolchain mingw-w64-x86_64-libtool mingw-w64-x86_64-gcc mingw-w64-i686-gcc mingw-w64-x86_64-gcc-fortran mingw-w64-i686-gcc-fortran mingw-w64-x86_64-libwinpthread-git mingw-w64-i686-libwinpthread-git mingw-w64-x86_64-SDL2 mingw-w64-i686-SDL2 mingw-w64-x86_64-ragel mingw-w64-x86_64-vulkan-headers mingw-w64-i686-vulkan-headers mingw-w64-x86_64-vulkan-loader mingw-w64-i686-vulkan-loader"
           set "PATH=C:\hostedtoolcache\windows\Python\3.10.11\x64;C:\msys64\usr\bin;%PATH%"
         )
-        
+
         C:\msys64\usr\bin\bash -lc "pacman -Q"
-        
+
         if "%CI_DEPLOY_PLATFORM%"=="windows-arm64" (
-          echo Installing Windows SDK 10
+          echo Installing Windows SDK 10.0.26100.4654
           curl -Lo winsdksetup.exe https://go.microsoft.com/fwlink/p/?LinkId=2327008
           start /wait winsdksetup.exe /features OptionId.DesktopCPParm64 OptionId.NetFxSoftwareDevelopmentKit /quiet
         ) else (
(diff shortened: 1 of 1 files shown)


.github/actions/deploy-windows/action.yml
```diff
@@ -53,47 +71,51 @@ runs:
...
+            curl -LO http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-winpthreads-git-11.0.0.r750.g05598db99-1-any.pkg.tar.zst
+            curl -LO http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-winstorecompat-git-11.0.0.r750.g05598db99-1-any.pkg.tar.zst
+            curl -LO http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-vulkan-headers-1.3.280.0-1-any.pkg.tar.zst
+            curl -LO http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-vulkan-loader-1.3.280.0-1-any.pkg.tar.zst
+            curl -LO http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-SDL2-2.30.12-1-any.pkg.tar.zst
+            curl -LO http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-python-3.11.9-1-any.pkg.tar.zst
+            bash -c "pacman -U --noconfirm *.pkg.tar.zst"
+          ) else (
+            bash -c "pacman -Rc --noconfirm mingw-w64-i686-headers-git mingw-w64-x86_64-headers-git mingw-w64-i686-zlib mingw-w64-x86_64-zlib"
+          )
```
saudet, line 118: I don't think we need to modify these lines since those are only related to GCC which isn't available for ARM64 if I read this correctly?

.github/actions/deploy-windows/action.yml
```diff
@@ -10,24 +10,42 @@ runs:
...
-        rm "C:/Strawberry/c/lib/libz.a" "C:/Strawberry/c/lib/libzlib.a" "C:/Strawberry/c/lib/libzdll.a" "C:/Strawberry/c/bin/cmake.exe"
+        if not "%CI_DEPLOY_PLATFORM%"=="windows-arm64" (
+          echo Removing broken stuff from WSL, MSYS2, etc
+          rm "C:/msys64/usr/bin/curl.exe" "C:/msys64/mingw32/bin/curl.exe" "C:/msys64/mingw64/bin/curl.exe"
+          rm "C:/WINDOWS/system32/bash.EXE" "C:/msys64/usr/bin/link.exe" "C:/msys64/usr/bin/timeout.exe" "C:/msys64/usr/bin/python.exe" "C:/msys64/usr/bin/python3.exe"
+          rm "C:/ProgramData/chocolatey/bin/gfortran.exe" "C:/msys64/mingw32/bin/gfortran.exe" "C:/msys64/mingw32/bin/python.exe" "C:/msys64/mingw32/bin/python3.exe"
+          rm "C:/Strawberry/c/bin/gfortran.exe" "C:/msys64/mingw64/bin/gfortran.exe" "C:/msys64/mingw64/bin/python.exe" "C:/msys64/mingw64/bin/python3.exe"
+          rm "C:/msys64/mingw32/bin/clang-cl.exe" "C:/msys64/mingw64/bin/clang-cl.exe" "C:/msys64/mingw32/bin/cmake.exe" "C:/msys64/mingw64/bin/cmake.exe"
+          rm "C:/Strawberry/c/lib/libz.a" "C:/Strawberry/c/lib/libzlib.a" "C:/Strawberry/c/lib/libzdll.a" "C:/Strawberry/c/bin/cmake.exe"
+        )
```
saudet, line 48: Since those lines can safely fail, we don't need to modify them, right?

devjeonghwan: Some files are in a Windows on ARM + MSYS2 environment. I'm not sure, but I think it's okay to revert the changes.

saudet: Right, we want to remove those files anyway that's OK

.github/actions/deploy-windows/action.yml
devjeonghwan: Install the 'mingw-w64-clang-aarch64-gcc-compat' package for GCC support. Yes, that's right. We can change it to a smaller range of if blocks.

.github/actions/deploy-windows/action.yml
```diff
@@ -10,16 +10,32 @@ runs:
...
+          set "PATH=C:\msys64\usr\bin;%PATH%"
+        ) else (
+          C:\msys64\usr\bin\bash -lc "pacman -S --needed --noconfirm pkg-config"
+          C:\msys64\usr\bin\bash -lc "pacman -S --needed --noconfirm base-devel git tar unzip p7zip zip autoconf autoconf-archive automake libtool make patch gnupg"
+          C:\msys64\usr\bin\bash -lc "pacman -S --needed --noconfirm mingw-w64-x86_64-nasm mingw-w64-x86_64-toolchain mingw-w64-x86_64-libtool mingw-w64-x86_64-gcc mingw-w64-i686-gcc mingw-w64-x86_64-gcc-fortran mingw-w64-i686-gcc-fortran mingw-w64-x86_64-libwinpthread-git mingw-w64-i686-libwinpthread-git mingw-w64-x86_64-SDL2 mingw-w64-i686-SDL2 mingw-w64-x86_64-ragel mingw-w64-x86_64-vulkan-headers mingw-w64-i686-vulkan-headers mingw-w64-x86_64-vulkan-loader mingw-w64-i686-vulkan-loader"
+          set "PATH=C:\hostedtoolcache\windows\Python\3.10.11\x64;C:\msys64\usr\bin;%PATH%"
```
saudet, line 25: I take it Python isn't available for windows-arm64 yet?

saudet: No, it's supposed to be available:
https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md
Please add it to the PATH here

devjeonghwan: Yes I know, but Python already seems to be added in PATH. Do I still have to manually set PATH?

devjeonghwan:
I added JDK and Python installation scripts for Windows ARM64, and added environment variable settings.

.github/actions/deploy-windows/action.yml
```diff
@@ -8,18 +8,50 @@ runs:
...
+          set "JAVA_HOME=C:\Program Files\Microsoft\jdk-11.0.28.6-hotspot"
+          set "PATH=C:\Program Files\Microsoft\jdk-11.0.28.6-hotspot\bin;%PATH%"
+        )
+
+        if "%CI_DEPLOY_PLATFORM%"=="windows-arm64" (
+          echo Installing Python 3.11.4 for Windows on ARM
+          curl -Lo python-3.11.4-arm64.exe https://www.python.org/ftp/python/3.11.4/python-3.11.4-arm64.exe
+          start /wait python-3.11.4-arm64.exe /quiet InstallAllUsers=1 PrependPath=1 Include_test=0
+          set "PATH=C:\Program Files\Python311-arm64;C:\Program Files\Python311-arm64\Scripts;%PATH%"
+        )
```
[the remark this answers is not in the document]
devjeonghwan: No, the image does come with JDK and Python installed. However, they are x86 builds, not ARM64 builds. Therefore, they run under emulation, causing platform detection APIs to return incorrect values. For example, `System.getProperty("os.arch")`, `System.getProperty("os.name")` and `platform.machine()`, `platform.architecture()`, `os.uname()` will return emulated x86 architecture information instead of the actual host architecture.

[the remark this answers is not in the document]
devjeonghwan: Do I need to merge Windows SDK installation blocks as well?

PR #1657, opened 2025-07-25
Target LLVM 20.1.7

koutheir:
Target LLVM 20.1.7. The [previous commit](https://github.com/bytedeco/javacpp-presets/commit/4bfcd53496e3d8674bdf90959c13ecd6d9f89bed), that was supposed to do this, missed the changes I'm proposing here.


koutheir:
For reference, this fixes a bug where the LLVM shared libraries stored inside `llvm-linux-x86_64.jar` for [the latest release `1.5.12`](https://github.com/bytedeco/javacpp-presets/releases/tag/1.5.12) do not include the LLVM version number after `.so`, which causes `UnsatisfiedLinkError`s because the libraries are linked against the shared library names that **include** the version number.

For instance, the dynamic loader trying to load `libjniLLVM.so` would be looking for `libRemarks.so.20.1` in the same directory, and would complain due to its absence (only `libRemarks.so` is stored under `$HOME/.javacpp`).

saudet:
Oops, I guess I forgot to update that, yes. Let's update the version to 20.1.8 while we're at it though. Thanks!

koutheir:
> Let's update the version to 20.1.8 while we're at it though.

I don't know how to do that.

saudet (approved):

saudet:
We just need to replace "20.1.7" with "20.1.8" in these files, but we can do that another time:
```
javacpp-presets$ grep -r '20\.1\.7' 
platform/pom.xml:      <version>20.1.7-${project.version}</version>
llvm/README.md: * LLVM 20.1.7  http://llvm.org/
llvm/README.md:            <version>20.1.7-1.5.12</version>
llvm/pom.xml:  <version>20.1.7-${project.parent.version}</version>
llvm/platform/pom.xml:  <version>20.1.7-${project.parent.version}</version>
llvm/samples/polly/pom.xml:            <version>20.1.7-1.5.13-SNAPSHOT</version>
llvm/samples/clang/pom.xml:            <version>20.1.7-1.5.13-SNAPSHOT</version>
llvm/samples/llvm/pom.xml:            <version>20.1.7-1.5.13-SNAPSHOT</version>
```

PR #1656, opened 2025-07-25
Add windows-arm64 builds to presets for LLVM

devjeonghwan:
## Result of `llvm/samples/llvm/EmitBitcode.java`
```
OS: Windows 11
Arch: aarch64
Version: 10.0

The result of add(42, 30) imported from bitcode and executed with LLVM interpreter is: 72

[Incubating] Problems report is available at: file:///C:/Users/USER_NAME/Desktop/llvm_test/build/reports/problems/problems-report.html

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.13/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD SUCCESSFUL in 3s
```

PR #1658, opened 2025-07-26
Add windows-arm64 builds to presets for systems

devjeonghwan:
This PR should be merged after #1656 is merged.

We can update the Windows SDK version for `windows-x86_64` to match `windows-arm64` (10.0.26100.4654), and it has been tested on x86 machines. However, I decided to keep the current SDK version for now.

Note that the generated `systems-1.5.13.jar` differs when built for `windows-x86_64` versus `windows-arm64` due to architecture-specific macro definitions during JavaCPP header parsing. This results in different APIs in `global/windows.class` and other classes across architectures.

### DIFF
--- a/.github/workflows/systems.yml
+++ b/.github/workflows/systems.yml
@@ -45,6 +45,10 @@ jobs:
     runs-on: macos-13
     steps:
       - uses: bytedeco/javacpp-presets/.github/actions/deploy-macosx@actions
+  windows-arm64:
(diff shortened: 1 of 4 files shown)


saudet:
It wouldn't be a bad idea to update to SDK 10.0 for x86 as well since you've already got it working

devjeonghwan:
OK, I'll update it and reopen the PR

PR #1660, opened 2025-07-31
Add deploy action to install WMIC for Windows on ARM

devjeonghwan:
Windows on ARM runner image has been updated(`20250716.19.1` to `20250728.24.1`) and the WMIC, has been excluded.
Therefore, install it directly through DISM.


saudet:
Why is WMIC needed at all?

saudet:
Is it just for querying info about the pagefile? Let's use a different command for that instead of installing WMIC

devjeonghwan:
We can use "Get-CimInstance Win32_PageFileUsage | Select-Object * | Format-List" powershell command instead of WMIC. 
Should I apply this?

saudet:
If it works, sure

devjeonghwan:
Applied, thanks

saudet (approved):

PR #1689, opened 2025-09-16
Update cppbuild.sh, enable "--use_dnnl" on macosx-arm64

blueberry:
Hi @saudet 

ONNX runtime works all right now on macosx-arm64. However, it has been built without the "--use_dnnl" flag, so it only supports the CPU engine provider. 

Since the latest snapshots, DNNL is available on  macosx, so it've updated the build to use DNNL in MacOS. It passes, so I guess it's safe to merge, and then I'll be able to check if it brings any speedups on the macos.


blueberry:
Thanks!

PR #1703, opened 2025-10-13
Fix newer C++ compiler errors by including <cstdint> in optimizer_api.h

blueberry:
The current build (1.22.2) of onnxruntime fails with newer compilers. This is a problem with upstream too, not just javacpp). In upstream, this is solved by adding #include <cstdint> in optimizer.h. However, that fix only made it to the 1.23 release, and is not present in 1.22.2.


saudet:
The fix is in 1.23.x so I assume we don't need this.

PR #1714, opened 2025-10-26
Update badges

ben221199:
Like https://github.com/bytedeco/javacv/pull/2363, this pull request is updating badges to show correct versions, showing both javacpp-presets as javacpp-presets-platform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,12 @@
 JavaCPP Presets
 ===============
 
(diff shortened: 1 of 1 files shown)


saudet:
There are more submodules

saudet:
We need to update the README.md file of each submodule like this one:
https://github.com/bytedeco/javacpp-presets/tree/master/pytorch

ben221199:
I see.

Issue #1721, opened 2025-12-05
[PyTorch] Native memory leak in AbstractTensor: createBuffer() and createIndexer() methods


iwalsh:
In `org.bytedeco.pytorch.AbstractTensor`, the methods [createBuffer(long index)](https://github.com/bytedeco/javacpp-presets/blob/master/pytorch/src/main/java/org/bytedeco/pytorch/AbstractTensor.java#L88-L119) and [createIndexer(boolean direct)](https://github.com/bytedeco/javacpp-presets/blob/master/pytorch/src/main/java/org/bytedeco/pytorch/AbstractTensor.java#L125-L170) are each creating objects of type `org.bytedeco.pytorch.TensorOptions` and `org.bytedeco.pytorch.Device` and never closing them. These types extend from `org.bytedeco.javacpp.Pointer`, so they allocate native memory and without them being closed this native memory is not freed in a timely / deterministic fashion, causing a leak.

---

Here is a partial heap dump of our production system showing the memory consumed by the leaks`TensorOptions` and `Device`:
```
1249940736  23.4%  23.4% 1249940736  23.4% Java_org_bytedeco_pytorch_TensorOptions_device__
1225691676  22.9%  46.3% 1225691676  22.9% Java_org_bytedeco_pytorch_TensorBase_options
1205899200  22.5%  68.8% 1205899200  22.5% Java_org_bytedeco_pytorch_TensorOptions_allocate__Lorg_bytedeco_pytorch_global_torch_00024ScalarType_2
655977372  12.3%  81.1% 655977372  12.3% os::malloc@ccb2e0
504454750   9.4%  90.5% 504762888   9.4% os::malloc@ccb520
155669273   2.9%  93.4% 155669273   2.9% inflate
111853568   2.1%  95.5% 111853568   2.1% c10::alloc_cpu
49284576   0.9%  96.4% 49284576   0.9% Java_org_bytedeco_pytorch_TensorBase_dtype
```

---

As a temporary fix, we have wrapped the call sites in `org.bytedeco.javacpp.PointerScope` so that the intermediate objects are closed when the methods complete:
```java
// Patching the native memory leak from createBuffer()
final org.bytedeco.pytorch.Tensor tensor = ...;
final ByteBuffer dst = ByteBuffer.allocate((int) tensor.numel());
try (PointerScope scope = new PointerScope(TensorOptions.class, Device.class)) {
[...]

PR #1722, opened 2025-12-09
[PyTorch] Uses PointerScope to fix native memory leaks in AbstractTensor

iwalsh:
In the `createBuffer()` and `createIndexer()` methods of `AbstractTensor`, objects of type `TensorOptions` and`Device` were being created but never closed. Each allocates a small amount of native memory, so this was causing a memory leak when `tensor.createBuffer()` or `AbstractTensor.create()`.

The fix is simply to wrap the methods in an appropriate `PointerScope` so that the offending objects are automatically closed when the method returns.

This resolves issue https://github.com/bytedeco/javacpp-presets/issues/1721


saudet:
Thanks! Please do the same for Tensor.create()

saudet (approved):

PR #1733, opened 2026-01-14
Add setter methods for GELUOptions.approximate() - Fixes #1730

### LINKED ISSUE #1730: Can't change GELU approx type
Image
GELUOptions have `GELU model(GELUOptions().approximate("none"));` as an example but there's no such method. Doing `GELUOptions().approximate().putString("tanh")` does nothing because if I pass it to `GELUImpl(options)` and then look at approximate from there - it will say "none"

```kotlin
val options = GELUOptions()
options.approximate().putString("tanh")
gelu = GELUImpl(options)

println(gelu.options().approximate().string)
```

Output is: "None"

Harshit057:
## Summary

This PR fixes issue #1730 by adding setter methods to `GELUOptions.approximate()`, allowing users to configure the approximation type as described in the documentation.

## Problem

The `GELUOptions` class previously exposed only a getter method for `approximate()`.  
Because of this, the documented API usage could not be used in practice.

## Solution

The `torch.java` preset has been updated to include `javaText` configuration so that setter methods are generated alongside the existing getter.

This enables the intended fluent API usage.

## Implementation Details

- Modified `torch.java`
- Follows the same pattern used by other options classes (e.g., `DistributedBackendOptions.group_id`)
- Setter methods return `GELUOptions` to support method chaining
- Backward compatibility is preserved by keeping the existing getter

## Testing

After rebuilding the PyTorch module using:

## MVN clean install
 
the regenerated `GELUOptions.java` contains the required setter and getter methods, enabling the documented API usage.

## Related Issue

Fixes #1730

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 
- * Add setter methods for `GELUOptions.approximate()` in presets for PyTorch ([pull #1730](https://github.com/bytedeco/javacpp-presets/pull/1730))
+ * Add setter methods for `GELUOptions.approximate()` in presets for PyTorch ([pull #1733](https://github.com/bytedeco/javacpp-presets/pull/1733))
  * Build FFmpeg with HarfBuzz to reenable the `drawtext` filter ([pull #1731](https://github.com/bytedeco/javacpp-presets/pull/1731))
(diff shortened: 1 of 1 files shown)


saudet:
Thanks! Looks good, but could you also update the generated code in src/gen ?

Harshit057:
@saudet sure i will

saudet (approved):

Harshit057:
Thanks for the review and approval @saudet! Since all checks have passed, could this be merged?

PR #1750, opened 2026-04-13
Upgrade ONNX Runtime to 1.24.4.

blueberry:
Just a change from 1.24.1 to the latest 1.24.4.


saudet:
Thanks! Mind doing it for ONNX as well?

blueberry:
> Thanks! Mind doing it for ONNX as well?

Sure! I wasn't aware that it got an update too. Done.

BTW, is it possible to add tensorrt ep to the onnxruntime build without much hassle? I hope so, since it comes as a binary, and a similar thing has already been done with the DNNL ep, but I can't test that on my machine, since I'm having problems with some cmake error when I build onnxruntime even as it is (on my machine, github actions is all right)?
OpenVINO support is also crucial to onnxruntime on CPU, since without it the vanilla and DNNL eps are poorly optimized for anything related to LLMs. Does that require a separate javacpp-preset if we only want to use it through onnxruntime, and not as a standalone preset?

saudet:
We can add TensoRT, OpenVINO, whatever supports in the current presets for ONNX Runtime, that's fine I think it's just a hassle, yes. Someone needs to do it...

blueberry:
@saudet Can you please restart the build actions? The failures seem to be due to network errors.

barrypitman:
> We can add TensoRT, OpenVINO, whatever supports in the current presets for ONNX Runtime, that's fine I think it's just a hassle, yes. Someone needs to do it...

I'm looking at building the OpenVINO EP for ONNX runtime. @saudet would you prefer that it gets linked as a separate maven module / classifier, or part of the main onnxruntime build? I'm thinking a separate module, i.e. -openvino, similar to -gpu, would be cleaner?

blueberry:
@barrypitman if it gets linked as a separate maven module, then it can't be mixed with ~CUDA/TensorRT EP or DNNL EP. 

I built onnxruntime with TensorRT (waiting for this PR to get merged to contribute another) and the build passed smoothly. I didn't test whether it actually works, but it shows in the list of available providers when queried through C api.

I think there should be only two builds since then it's only about available hardware. The users that only have CPU would use vanilla onnxruntime with DNNL and OpenVINO included, and the users that have GPU would use -gpu and get everything. If they don't try to call a provider that is not supported by their hardware or available redist binaries (say, OpenVINO), it's no harm to have EP included as it is simply ignored until requested...

Maybe OpenVINO would benefit from its separate preset, even a dummy one, as a binary redist provider.

barrypitman:
that's a good point @blueberry. Are you saying that you plan on tackling the OpenVINO EP after TensorRT? If that's the case I'll wait on your changes. Would you look at Linux and Windows support? Happy to test.

Also, agreed that having a "binary redist provider" for OpenVINO would be handy, to avoid installing OpenVINO separately

blueberry:
@barrypitman I didn't plan on it, as there's currently no preset for it, and I am not a C++ expert  :)
TensorRT was already there, so that was a lower hanging fruit. It would be great if you'd do OpenVINO (and when you're there, you can include TensorRT; you don't have to wait for my changes, TensorRT was fairly straightforward).

barrypitman:
OK, understood. Have you submitted your PR for the TensorRT EP? Oh wait, you are saying that you are waiting for this [current PR](https://github.com/bytedeco/javacpp-presets/pull/1750) to be merged before submitting the TensorRT EP one that you are working on?

blueberry:
Yes. I had to butcher my local cppbuild.sh to make it work with CMake on my machine (some conflicting packages on arch vs what onnxruntime internally pulls), that's why it is not a straightforward job to contribute easily. That's why it would be great if you could do it along with OpenVINO...

But if onnxruntime preset builds on your machine without changes, tensorrt is easy to add.

saudet (approved):

PR #1753, opened 2026-04-21
Ugrade to onnxruntime 1.25.0.

blueberry:
Today they released the 1.25.0 version, so I did a quick update. Hope it builds well :)


blueberry:
I've be trying to make this work on GPU for 2 days without success. It seems that onnxruntime got some new cccl/thrust headers in the last release that make the compiler trip, and my C++ fu is insufficient enough to fix it. Can you please try to build it @barrypitman ?

saudet:
@blueberry It looks like ORT Training is deprecated and no longer supported by anyone. Let's disable it?

blueberry:
if they block the build, then sure. But I couldn't fond the answer whether they are deprecated or it's just a matter of the ort build being brittle. Anyway, I'd prefer a working build :)
Siince you re at it now, can you please enable the TensiorRT ep? I tried it with 1.24.4 locally and it builds without additional hurdles. I wanted to get the plain version upgrade or merged first, and then enable it, but then got blocked by the gpu build of 1.15.0.
Also if we can enable OpenVINO (even the last years version until the new one is available in ort) that would be fantastic!

saudet (approved):

PR #1752, opened 2026-04-15
new OpenVINO preset

barrypitman:
This preset packages already-built OpenVINO v2025.4.1 libraries for CPU, Intel GPU and NPU for Windows and Linux. I have tested the sample on Linux and it works:

OpenVINO JavaCPP JNI library loaded from: /root/.javacpp/cache/openvino-2025.4.1-1.5.14-SNAPSHOT-linux-x86_64.jar/org/bytedeco/openvino/linux-x86_64/libjniopenvino.so
Allocated ov_core_t wrapper successfully: org.bytedeco.openvino.ov_core_t[address=0x0,position=0,limit=0,capacity=0,deallocator=null]
OpenVINO preset load verification complete.

I think that having this preset pre-packaged and easily includable will be useful for the OnnxRuntime OpenVINO EP (separate pull request).

PR #1769, opened 2026-05-14
Update onnxruntime to version 1.26.0, add OpenVINO build flags

barrypitman:
Once the OpenVINO changes are merged, we can declare an optional dependency on the OpenVINO preset (I think).


blueberry:
I built the CPU version on my Arch linux. It succeeded (with manual changes for the abseil problem specific for arch, not in general) and now I'm trying to build CUDA.

blueberry:
I am not sure that we need openvino on macos-arm64. It might work, but if it's anything like the dnnl build for macos, it's much slower than coreml that is already supproted, and surely slower than MLX, which we can add as a new preset not unlike openvino. So if ovenvino does not build on that platform, I think we should stick to linux and windows x86_64, and support macos with its own optimized backends.

barrypitman:
@m1ngyuan requested the MacOS support for OpenVINO on [this PR](https://github.com/bytedeco/javacpp-presets/pull/1752) so would be interesting to hear what they think

barrypitman:
I have updated the PR so that it references the new OpenVINO preset. 

I found that ONNX Runtime’s native build.py invokes CMake which calls `find_package(OpenVINO REQUIRED)` before JavaCPP has a chance to unpack OpenVINO resources from the Maven dependency jars. 

That is why I added this line[ `(cd "$TOP_PATH" && bash cppbuild.sh install openvino -platform "$PLATFORM")`](https://github.com/bytedeco/javacpp-presets/pull/1769/changes#diff-b2a5a87a80ce6a89ea62172e17c08534b0e9c8e330628ec51983f07160514157R79). The alternative is that we could simply install openvino using pip, I'm not sure what is preferred. 

I'd like to test the OpenVINOExecutionProviderSample locally, @saudet maybe you could advise - how/when would we push snapshot builds to Sonatype?

saudet:
The build already gets the OpenVINO binaries from Maven. Use those

barrypitman:
I think that we'd need to use the maven dependency plugin to unpack the relevant openvino JAR file in the maven validate goal, is that what you had in mind? 

When running the cppbuild.sh script (in the maven `initialize` lifecycle phase), it is looking for the OpenVINOConfig.cmake file, but the maven dependencies are only extracted / unpacked later in the build (during `generate-sources`).

saudet:
If it needs OpenVINOConfig.cmake add it to the Maven artifact, that's what it's for

saudet:
Anything added to buildResources like this will get extracted in time for the cppbuild.sh script:
https://github.com/bytedeco/javacpp-presets/blob/master/opencv/pom.xml#L105-L111

onnxruntime/src/main/java/org/bytedeco/onnxruntime/presets/onnxruntime.java
```diff
@@ -70,6 +70,10 @@
                        "onnxruntime/core/session/onnxruntime_ep_c_api.h"},
             link = {"onnxruntime_providers_shared", "onnxruntime@.1", "onnxruntime_providers_dnnl"}
         ),
+        @Platform(
+            value = {"linux-x86_64", "macosx-arm64", "windows-x86_64"},
+            link = {"onnxruntime_providers_shared", "onnxruntime@.1", "onnxruntime_providers_dnnl", "onnxruntime_providers_openvino"}
+        ),
```
saudet, line 76: These settings don't add new values, they override them. Modify the existing lines instead

onnxruntime/cppbuild.sh
```diff
@@ -98,6 +121,9 @@ sedinplace 's/MLAS_CPUIDINFO::GetCPUIDInfo().HasArmNeon_I8MM()/false/g' onnxrunt
 
 # work around toolchain issues on Mac and Windows
 patch -p1 < ../../../onnxruntime.patch
+git apply --recount ../../../onnxruntime-openvino-no-format.patch
+git apply --recount ../../../onnxruntime-openvino-gcc11.patch
+git apply --recount ../../../onnxruntime-openvino-gcc13.patch
```
saudet: Not sure why we need those patches, but let's just concat everything into onnxruntme.patch

onnxruntime/cppbuild.sh
```diff
@@ -172,7 +202,15 @@ sedinplace 's/Throw(javaException)/Throw((jthrowable)javaException)/g' java/src/
...
-ortApiHandle = initialiseAPIBase(ORT_API_VERSION_1);\
+ortApiHandle = initialiseAPIBase(ORT_API_VERSION_23);\
+if (ortApiHandle == 0L) {\
+  throw new IllegalStateException("There is a mismatch between the ORT class files and the ORT native library, and the native library could not be loaded");\
+}\
+ortTrainingApiHandle = initialiseTrainingAPIBase(ortApiHandle, ORT_API_VERSION_23);\
+ortCompileApiHandle = initialiseCompileAPIBase(ortApiHandle);\
+trainingEnabled = ortTrainingApiHandle != 0L;\
+providers = initialiseProviders(ortApiHandle);\
+version = initialiseVersion();\
```
saudet, line 211: What happens if we don't add all those lines?

onnxruntime/pom.xml
```diff
@@ -24,6 +25,11 @@
       <artifactId>dnnl</artifactId>
       <version>3.11-${project.parent.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.bytedeco</groupId>
+      <artifactId>openvino</artifactId>
+      <version>2026.1.0-${project.parent.version}</version>
+    </dependency>
```
saudet, line 33: Let's make this optional, and DNNL too I guess while we're at it

onnxruntime/pom.xml
```diff
@@ -161,4 +181,31 @@
...
+    <profile>
+      <id>windows-x86_64</id>
+      <activation>
+        <property>
+          <name>javacpp.platform</name>
+          <value>windows-x86_64</value>
+        </property>
+      </activation>
+      <properties>
+        <openvino.link.path>${basedir}/../openvino/cppbuild/${javacpp.platform}/runtime/lib/intel64/Release/</openvino.link.path>
```
saudet: Instead of using profiles, just put all the paths up there for all platforms

onnxruntime/src/main/java/org/bytedeco/onnxruntime/presets/onnxruntime.java
saudet: Actually, something clean with only a single `@Platform` like this may work. Could you try that?
```java
        @Platform(
...
            link = {"onnxruntime@.1"},
            preload = {"onnxruntime_providers_shared", "onnxruntime_providers_cuda", "onnxruntime_providers_dnnl", "onnxruntime_providers_openvino"}
...
```

saudet:
And we'll need this patch to get the build working for CUDA on Windows: 
- https://github.com/microsoft/onnxruntime/pull/28309

onnxruntime/src/main/java/org/bytedeco/onnxruntime/presets/onnxruntime.java
barrypitman: I tried running the sample with the providers in preload, but I got this error:
```
SIGSEGV
Problematic frame:
C  [libonnxruntime_providers_dnnl.so+0x1f825]  _GLOBAL__sub_I_provider_bridge_provider.cc+0x45
```
However if I put them all in "link" with a single @Platform, it works.

onnxruntime/cppbuild.sh
barrypitman: I've combined them into a single onnxruntime-openvino.patch. I was testing locally using Ubuntu 24.04 and it was passing, but failing on GH Actions, the changes seem to relate to usages of C++20 <format>. To be honest I'm not sure about them either, which is why I think maybe a separate patch file would be better. Happy to combine if you prefer though.

onnxruntime/cppbuild.sh
barrypitman: It still works without those files, I've removed them.

saudet: Let's still leave the old init() there...

saudet: Ah, I see what's it's doing. The new init() looks better

onnxruntime/src/main/java/org/bytedeco/onnxruntime/presets/onnxruntime.java
barrypitman: I've restored the separate -gpu @Platform annotation

onnxruntime/cppbuild.sh
barrypitman: i've also restored the new init() method

onnxruntime/src/main/java/org/bytedeco/onnxruntime/presets/onnxruntime.java
saudet: OpenVINO should work on mac though, why doesn't it work?

barrypitman: it might do, i don't have a mac to test on. We are currently only building openvino for windows-x86_64, linux-x86_64 and macosx-arm64

blueberry: Given that Apple don't ship Intel-based Macs for the last six years, and that OpenVINO targets novel HPC tasks, AND that there'r Apple dedicated CoreML engine anyway, I doubt that there would be many users who would need that feature, if at all.

blueberry: What I wanted to say, almost all potential users who are on Macos, would have arm64 macs anyway, and this build supports that.

onnxruntime/cppbuild.sh
```diff
@@ -109,6 +133,8 @@ sedinplace 's/-fvisibility=hidden//g' cmake/CMakeLists.txt cmake/adjust_global_c
 sedinplace 's:/Yucuda_pch.h /FIcuda_pch.h::g' cmake/onnxruntime_providers_cuda.cmake cmake/onnxruntime_providers.cmake
 sedinplace 's/${PROJECT_SOURCE_DIR}\/external\/cub//g' cmake/onnxruntime_providers_cuda.cmake cmake/onnxruntime_providers.cmake
 sedinplace 's/-Xcompiler \/Zc:__cplusplus/-Xcompiler \/Zc:__cplusplus -Xcompiler \/Zc:preprocessor/g' cmake/onnxruntime_providers_cuda.cmake cmake/onnxruntime_providers_cuda_plugin.cmake
+# CUDA 13 on Windows needs the ORT CUB wrapper to undefine MSVC SAL macros: https://github.com/microsoft/onnxruntime/pull/28309
+sedinplace 's|#include <cub/cub.cuh>|#include "core/providers/cuda/cu_inc/cub.cuh"|g' onnxruntime/contrib_ops/cuda/bert/gqa_unfused_attention.cu
```
saudet: Doesn't look like this is getting applied for some reason

barrypitman: It looks like [this pull request](https://github.com/microsoft/onnxruntime/pull/28309) was merged before v1.26.0 was released, so it is likely already patched. So I'm not too sure what the issue is...

saudet: Since it works with 1.25.1 maybe what we need to do is to revert this patch?

barrypitman: I will try to debug it over the weekend

barrypitman: OK, it is passing now. Is there a way that we can push those build artifacts as snapshots so that I can try them locally?

saudet: You can use your own account credentials for CI_DEPLOY_USERNAME and CI_DEPLOY_PASSWORD...

onnxruntime/cppbuild.sh
```diff
@@ -133,15 +133,19 @@ sedinplace 's/-fvisibility=hidden//g' cmake/CMakeLists.txt cmake/adjust_global_c
...
+CUDA_HOME_UNIX="$CUDA_HOME"
+if command -v cygpath >/dev/null 2>&1; then
+    CUDA_HOME_UNIX="$(cygpath -u "$CUDA_HOME")"
+fi
+CUDA_TCGEN05_LD_H="$CUDA_HOME_UNIX/include/cccl/cuda/__ptx/instructions/generated/tcgen05_ld.h"
+if [[ -f "$CUDA_TCGEN05_LD_H" ]]; then
+    echo "Patching CUDA CCCL tcgen05_ld.h from $CUDA_TCGEN05_LD_H"
+    mkdir -p onnxruntime/cuda/__ptx/instructions/generated
+    cp "$CUDA_TCGEN05_LD_H" onnxruntime/cuda/__ptx/instructions/generated/tcgen05_ld.h
+    sedinplace 's/__out/__cccl_out/g' onnxruntime/cuda/__ptx/instructions/generated/tcgen05_ld.h
```
saudet: Let's patch CUDA in the actions deploy script instead

onnxruntime/cppbuild.sh
blueberry: > OK, it is passing now. Is there a way that we can push those build artifacts as snapshots so that I can try them locally?

That's great! Thank you for making this work!

onnxruntime/cppbuild.sh
barrypitman: I'm guessing that should be a separate PR targeting the "actions" branch?

saudet: No, you can do it here, just redirect the actions scripts in your workflows like you did before. BTW, it seems possible to use relative paths for actions scripts these days...

saudet: Ah, I knew there was a catch. Let's try to do that kind of thing in another pull request 

> Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under 'D:\a\javacpp-presets\javacpp-presets\.github\actions\deploy-windows\action.yml'. Did you forget to run actions/checkout before running your local action?

saudet (approved):

blueberry:
@saudet It seems that the github action that should have deployed this stalled https://github.com/bytedeco/javacpp-presets/actions/runs/26425638965/job/77857915571

PR #1771, opened 2026-05-25
Patch windows workflow to patch CUDA CCCL __out macro collision

barrypitman:


saudet (approved):

saudet:
@devjeonghwan BTW, I think the training API will build with this patch applied

devjeonghwan:
@saudet I'm trying to build from the current HEAD. xD

devjeonghwan:
> @devjeonghwan BTW, I think the training API will build with this patch applied

image

It works.

PR #1793, opened 2026-07-28
Fix ONNX Runtime OpenVINO model compilation

### LINKED ISSUE #1788: [onnxruntime] OpenVINO EP broken in 1.27.0-SNAPSHOT
Hi @saudet 
I tried to test the onnxruntime build 1.27.0-SNAPSHOT from today. The default CPU provider, and the DNNL provider work, but OpenVINO fails even with the simplest models that work with 1.26.0-SNAPSHOT. I tried mnist-12.onnx, SmolLM, SmolLM2, GPT, from HF, but all fail in the same way. They all work with the previous version. Even a basic logit model fails.
This is the typical error:
ONNX runtime error: Exception during initialization:
   /home/runner/work/javacpp-presets/javacpp-presets/onnxruntime/cppbuild/linux-x86_64-gpu/onnxruntime/onnxruntime/core/providers/openvino/backends/basic_backend.cc:136
   onnxruntime::openvino_ep::BasicBackend::BasicBackend(std::unique_ptr<onnx::ModelProto>&,
   onnxruntime::openvino_ep::SessionContext&, const
   onnxruntime::openvino_ep::SubGraphContext&,
   onnxruntime::openvino_ep::SharedContext&,
   onnxruntime::openvino_ep::ptr_stream_t&)
   /home/runner/work/javacpp-presets/javacpp-presets/onnxruntime/cppbuild/linux-x86_64-gpu/onnxruntime/onnxruntime/core/providers/openvino/ov_interface.cc:58
   auto onnxruntime::openvino_ep::OvExceptionBoundary(Func&&,
   std::string_view, Args&& ...) [with bool typed = false; Func =
   onnxruntime::openvino_ep::OVCore::ReadModel(std::string&&, const
   string&)::<lambda()>; Args = {}; std::string_view =
   std::basic_string_view<char>] [OpenVINO-EP] Exception while Reading
   network

barrypitman:
### Motivation
OpenVINO 2026.2 can fail when compiling ONNX models directly from their serialized string, causing the OpenVINO execution provider to throw during initialization; the frontend read/convert path (used by the AUTO device) avoids this failure.

### Description
Force auto_unified_compile = false in the OpenVINO BasicBackend (via the OpenVINO integration patch) so all device types use the frontend read/convert path instead of the failing unified serialized-ONNX compile path.

### Testing
Reproduced the reported CPU failure using published 1.28.0-*-SNAPSHOT artifacts with OpenVINOExecutionProviderSample, which failed as described.
Verified the same sample succeeds with AUTO (frontend read/convert path) and that it succeeds after applying the change.
Verified the patch applies cleanly to the ONNX Runtime v1.28.0 source with git apply --recount --check, and ran bash -n onnxruntime/cppbuild.sh and git diff --check, all of which completed without errors.

### Note
This ideally needs to be addressed by an upstream fix, but this should work in the interim. Fixes https://github.com/bytedeco/javacpp-presets/issues/1788

### DIFF
--- a/onnxruntime/onnxruntime-openvino.patch
+++ b/onnxruntime/onnxruntime-openvino.patch
@@ -194,3 +194,19 @@ index dd43f856e0..6a34827b1a 100644
 +#elif __GNUC__
 +#pragma GCC diagnostic pop
 +#endif
+diff --git a/onnxruntime/core/providers/openvino/backends/basic_backend.cc b/onnxruntime/core/providers/openvino/backends/basic_backend.cc
+index b39db2be12..8cbded29ce 100644
+--- a/onnxruntime/core/providers/openvino/backends/basic_backend.cc
++++ b/onnxruntime/core/providers/openvino/backends/basic_backend.cc
+@@ -89,7 +89,10 @@ BasicBackend::BasicBackend(std::unique_ptr<onnx::ModelProto>& model_proto,
+ #if defined(OPENVINO_DISABLE_NPU_FALLBACK)
+     eligible_for_cpu_fallback = false;
+ #endif
+-    auto auto_unified_compile = (hw_target.find("AUTO") == std::string::npos);
++    // OpenVINO 2026.2 cannot compile an ONNX model directly from its serialized
++    // string. Use the frontend's read/convert path for all device types instead.
++    // This is already the path used by the AUTO device and other special cases.
++    auto auto_unified_compile = false;
+
+     // Unified compile is efficient with cahce_dir cached model loading that bypass Read Model
+     // Does not support model with exteral weights, dynamic input shape, Epctx onnx cached model,


blueberry (approved):
The details are beyond my knowledge of OpenVINO details and C++, but it looks all right on the surface.

saudet:
@blueberry Please test it locally

blueberry:
@saudet Locally, as in cloning the project to my machine and building it, or as in refreshing the snapshots from maven central and running the tests? (That didn't help, and this is the timestamp: Retrieving org/bytedeco/onnxruntime/1.28.0-1.5.14-SNAPSHOT/onnxruntime-1.28.0-1.5.14-20260728.185138-18.pom from maven-central-snapshots
As now is 4AM here, I can only do the cloning and the local build tomorrow.

saudet:
Bah, let's just merge this. It probably works I guess

Thanks @barrypitman !

saudet (approved):

blueberry:
Thanks @barrypitman ! Can you please test the official version from maven central, since it still fails with the official sample on my machine even with your changes applied (fetched the latest jars, etc.).

Issue #1794, opened 2026-07-29
[onnxruntime] OpenVINO EP broken in 1.28.0-SNAPSHOT


blueberry:
This is a new issue since I do not have permissions to re-open #1788

It seems that #1793 did NOT fix #1788. Here's what I get when I fetch the latest samples, update maven dependencies (with both -U, and erasing the existing jars from the local repository, just to be shure), and run the openvino sample
```
mvn compile exec:java -Dexec.mainClass="OpenVINOExecutionProviderSample"
mvn compile exec:java -Dexec.mainClass="OpenVINOExecutionProviderSample"
[INFO] Scanning for projects...
[INFO] 
[INFO] ---------------< org.bytedeco.onnxruntime:cxxapisample >----------------
[INFO] Building cxxapisample 1.5.14-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
Downloading from maven-snapshots: https://central.sonatype.com/repository/maven-snapshots/org/bytedeco/onnxruntime-platform/1.28.0-1.5.14-SNAPSHOT/maven-metadata.xml
Downloaded from maven-snapshots: https://central.sonatype.com/repository/maven-snapshots/org/bytedeco/onnxruntime-platform/1.28.0-1.5.14-SNAPSHOT/maven-metadata.xml (1.2 kB at 991 B/s)
Downloading from maven-snapshots: https://central.sonatype.com/repository/maven-snapshots/org/bytedeco/javacpp-presets/1.5.14-SNAPSHOT/maven-metadata.xml
Downloaded from maven-snapshots: https://central.sonatype.com/repository/maven-snapshots/org/bytedeco/javacpp-presets/1.5.14-SNAPSHOT/maven-metadata.xml (606 B at 1.1 kB/s)
Downloading from maven-snapshots: https://central.sonatype.com/repository/maven-snapshots/org/bytedeco/dnnl-platform/3.12.2-1.5.14-SNAPSHOT/maven-metadata.xml
Downloaded from maven-snapshots: https://central.sonatype.com/repository/maven-snapshots/org/bytedeco/dnnl-platform/3.12.2-1.5.14-SNAPSHOT/maven-metadata.xml (1.2 kB at 2.2 kB/s)
Downloading from maven-snapshots: https://central.sonatype.com/repository/maven-snapshots/org/bytedeco/opencl-platform/3.1-1.5.14-SNAPSHOT/maven-metadata.xml
[...]

PR #1795, opened 2026-07-29
Fix OpenVINOExecutionProviderSample.

blueberry:
Fix sample #1794


barrypitman:
I think hold off on this please, I'm working on a fix for the actual issue, which seems an issue with how we are currently packaging the openvino libraries in the preset.

barrypitman:
I've submitted https://github.com/bytedeco/javacpp-presets/pull/1796 instead, i think we can close this PR

Issue #1784, opened 2026-07-01
2 missing shared libraries in cuda-redist-cudnn


blueberry:
This might be an edge case, but I discovered it using ONNX runtime.

Sometimes ONNX runtime needs to create a convolution, and it tries to load some nvidia's shared libraries that seem to be parts of cuDNN.
When I don't specify the cuda-redist-cudnn dependency, everything works fine, as my Arch Linux has CUDA packages installed. However, if I include said dependency in my project, that particular part fails with: Unable to load any of {libcudnn_engines_tensor_ir.so.9.21.1, libcudnn_engines_tensor_ir.so.9.21, libcudnn_engines_tensor_ir.so.9, libcudnn_engines_tensor_ir.so}

These libraries are NOT inside ruda-redist-cudnn. It seems that they are distributed in some other (bigger?) Nvidia package thad is missed in JavaCPP either intentionally or because things changed during the years, and these simply failed through the cracks.

Thi is present in all recent versions as I have been encountering this for almost a year, but I didn't look into the cause as the easy workaround was to not use javacpp redist for cudnn. However, this is not a long-term solutions when other people need to use javacpp's cudnn calls...

saudet:
They do come with cuDNN, we'll just need to add them to these lists here to get them bundled and loaded:
https://github.com/bytedeco/javacpp-presets/blob/master/cuda/src/main/java/org/bytedeco/cuda/presets/cudnn.java#L38
https://github.com/bytedeco/javacpp-presets/blob/master/onnxruntime/src/main/java/org/bytedeco/onnxruntime/presets/onnxruntime.java#L106
https://github.com/bytedeco/javacpp-presets/blob/master/tensorrt/src/main/java/org/bytedeco/tensorrt/presets/nvinfer.java#L103
https://github.com/bytedeco/javacpp-presets/blob/master/opencv/src/main/java/org/bytedeco/opencv/presets/opencv_core.java#L99
https://github.com/bytedeco/javacpp-presets/blob/master/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java#L190

blueberry:
I tried, but unsuccessfully, to run this with GitHub actions. The trouble is that in Nvidia's distribution, these two new cudnn libraries that I added (cudnn_ext and libcudnn_engines_tensor_ir) are treated a bit differently that the long-existing .so s so the build can't find them. I've tried fixing this in the yaml, but unsuccessfuly:
[...]

PR #1796, opened 2026-07-30
Fix OpenVINO JavaCPP runtime layout for CPU execution-provider loading

barrypitman:
### Summary

Fixes OpenVINO execution-provider failures when ONNX Runtime is configured with
`device_type=CPU` through JavaCPP.

The failure was not caused by ONNX Runtime's CPU compilation path. The real
problem was that the OpenVINO preset flattened `libopenvino.so` and the frontend
libraries into JavaCPP's extracted `linux-x86_64/` cache directory. OpenVINO
discovers its frontend plugins relative to the location of the loaded OpenVINO
runtime library, so loading the flattened `libopenvino.so` meant the expected
`runtime/lib/intel64/` layout was no longer available.

As a result, the OpenVINO ONNX frontend could not be discovered correctly and
ONNX Runtime failed during session creation with:

    [OpenVINO-EP] Exception while Reading network

### Root cause

A native C++ reproducer using the same ONNX Runtime/OpenVINO artifacts succeeded
with both:

- `device_type=CPU`
- `device_type=AUTO`

The Java sample failed only with explicit `CPU`.

Comparing dynamic library loading showed the difference:

Successful native C++ process loaded OpenVINO from the original runtime tree:

    runtime/lib/intel64/libopenvino.so.2621
    runtime/lib/intel64/libopenvino_onnx_frontend.so.2621
    runtime/lib/intel64/libopenvino_intel_cpu_plugin.so

Failing JavaCPP process loaded flattened copies:

    linux-x86_64/libopenvino.so
    linux-x86_64/libopenvino_onnx_frontend.so
    linux-x86_64/libopenvino_intel_cpu_plugin.so

The binaries were byte-for-byte identical; the failure depended on where
`libopenvino.so` was loaded from.

Preloading the flattened `libopenvino.so` in the otherwise-working C++ reproducer
caused the same failure, confirming this as a JavaCPP runtime layout issue rather
than a JNI marshalling issue or a general ONNX Runtime/OpenVINO bug.

### Changes

- Update the Linux OpenVINO preset to preserve the original OpenVINO runtime
  directory layout for runtime loading.
- Keep normal versioned `link` entries for JavaCPP native compilation:

      link = {"openvino_c@.2621#", "openvino@.2621#"}

- Move runtime loading to explicit `preload` resource paths under
  `runtime/lib/intel64/`, including:
  - `libopenvino.so.2621`
  - `libopenvino_c.so.2621`
  - OpenVINO device plugins
  - OpenVINO frontend libraries, including the ONNX frontend
  - TBB from `runtime/3rdparty/tbb/lib/`
- Revert the previous ONNX Runtime OpenVINO patch change from PR #1793 that
  forced `auto_unified_compile = false`, since it was not the root cause.

### Validation

Built the OpenVINO preset with Maven using the real JavaCPP plugin configuration
and confirmed the generated Linux platform JAR preserves the versioned runtime
layout.

The original Java ONNX Runtime sample now succeeds with explicit CPU:

    OpenVINO device_type=CPU
    OpenVINO EP identity inference succeeded.

The AUTO control also remains successful.

### Why this matters

Users should be able to explicitly request:

    device_type=CPU

without relying on `AUTO` as a workaround. `AUTO` can select different devices on
systems with GPU/NPU support, so it is not equivalent to explicit CPU selection.


barrypitman:
@blueberry you should be able to test the snapshots on your end. There is an updated [openvino artifact](https://central.sonatype.com/repository/maven-snapshots/org/bytedeco/openvino/2026.2.1-1.5.14-SNAPSHOT/openvino-2026.2.1-1.5.14-20260730.094525-18-windows-x86_64.jar).

blueberry:
@barrypitman Thank you. The OpenVINO provider now works with "CPU" (and "AUTO" too).
@saudet  However, it seems that the buld change messed up something with the cuda distribution, as #1784 reappeared. If I use the cudnn redist, MNIST model fails with the Convolution28 etc., and if I comment out that dependency and instead use system's cudnn, the test works fine. (Should I report that in a separate issue or you have enough info to know what needs to be fixed?) Everything worked out fine last night.

blueberry:
@saudet Actually, it is not exactly the same as #1784, so I think that it is not related to this particular change. I've tried it with yesterday's snapshot, and now that particular cuda test fails with them too. It seem that that is unrelated to this particular PR, so it's good to go I think. I'll investigate the cuda problem, and if it turns to not be related to my system, I'll open a separate issue.

Thank you both @saudet and @barrypitman !

blueberry (approved):
This works on my system!

blueberry (approved):
The latest ORT (linux, from 16 minutes ago) works.

openvino/src/main/java/org/bytedeco/openvino/presets/openvino.java
```diff
@@ -34,99 +34,98 @@
...
-            },
-            resource = {"runtime"}
-        ),
-    },
-    target = "org.bytedeco.openvino",
-    global = "org.bytedeco.openvino.global.openvino"
+        inherit = OpenCL.class,
+        value = {
+                @Platform(
+                        value = {"linux-x86_64"},
```
saudet: Revert unnecessary formatting changes like using 8 spaces instead of 4...

openvino/src/main/java/org/bytedeco/openvino/presets/openvino.java
```diff
@@ -34,99 +34,98 @@
...
     @Override public void map(InfoMap infoMap) {
         infoMap.put(new org.bytedeco.javacpp.tools.Info("extern", "__cdecl").cppTypes().annotations())
-               .put(new org.bytedeco.javacpp.tools.Info("OPENVINO_C_API_EXTERN", "OPENVINO_C_API_CALLBACK").skip())
-               .put(new org.bytedeco.javacpp.tools.Info("OV_BOOLEAN", "BOOLEAN").skip())
-               .put(new org.bytedeco.javacpp.tools.Info("ov_rank_t").cast().valueTypes("long").pointerTypes("LongPointer", "LongBuffer", "long[]"))
-               .put(new org.bytedeco.javacpp.tools.Info("ov_dimension_t").cast().valueTypes("long").pointerTypes("LongPointer", "LongBuffer", "long[]"));
+                .put(new org.bytedeco.javacpp.tools.Info("OPENVINO_C_API_EXTERN", "OPENVINO_C_API_CALLBACK").skip())
+                .put(new org.bytedeco.javacpp.tools.Info("OV_BOOLEAN", "BOOLEAN").skip())
+                .put(new org.bytedeco.javacpp.tools.Info("ov_rank_t").cast().valueTypes("long").pointerTypes("LongPointer", "LongBuffer", "long[]"))
+                .put(new org.bytedeco.javacpp.tools.Info("ov_dimension_t").cast().valueTypes("long").pointerTypes("LongPointer", "LongBuffer", "long[]"));
```
saudet: And adding spaces here...

openvino/src/main/java/org/bytedeco/openvino/presets/openvino.java
```diff
@@ -34,99 +34,98 @@
...
     }
-}
+}
```
saudet: And removing a newline here

saudet (approved):

PR #1801, opened 2026-08-24
OpenVINO C API

barrypitman:
### Summary
Generate the public OpenVINO C API bindings instead of only opaque wrapper types. This allows the preset to be used directly instead of only via OnnxRuntime. 

- Expand the OpenVINO preset inputs to explicitly parse the public C headers, including GPU plugin property definitions.
- Map OPENVINO_C_API and OPENVINO_C_VAR so exported C API declarations are generated, while preserving Windows __cdecl handling.
- Generate callable core, model, tensor, preprocessing, infer-request, lifecycle, remote-context, and remote-tensor APIs in global.openvino.
- Add usable Java overloads for the variadic property APIs, including remote-context creation, remote tensor creation, and compilation with properties.
- Export Intel GPU property-key accessors, including VAAPI/OpenCL shared-memory context and tensor properties.
- Resolve the generated ov_dimension_t / ov_rank_t type collision by treating ov_dimension_t as its current struct representation.
- Update OpenVINOInfo to create/free a core, query available devices, and report API errors through generated bindings.


openvino/src/main/java/org/bytedeco/openvino/presets/openvino.java
```diff
@@ -27,19 +27,35 @@
...
 import org.bytedeco.javacpp.tools.InfoMap;
 import org.bytedeco.javacpp.tools.InfoMapper;
-import org.bytedeco.opencl.presets.OpenCL;
 
 /**
  *
  * @author Barry Pitman
  */
 @Properties(
-    inherit = OpenCL.class,
```
saudet, line 37: Please add this back

saudet (approved):

PR #1802, opened 2026-08-28
Make ONNX Runtime’s DNNL and OpenVINO dependencies optional

barrypitman:
## Summary

ONNX Runtime currently inherits the DNNL and OpenVINO presets, causing their native libraries to be loaded even when those execution providers are not required. This can prevent ONNX Runtime from loading on platforms without a corresponding OpenVINO preset, including ARM64.

This change:

- Marks the DNNL and OpenVINO Maven dependencies as optional.
- Removes their transitive Java module requirements and preset inheritance.
- Configures DNNL and OpenVINO preloads dynamically at load time, similar to the approach used for CUDA
- Adds platform-specific OpenVINO libraries only on supported platforms.
- Preserves the existing CUDA preload behavior for GPU builds.

This allows the base ONNX Runtime preset to load without requiring OpenVINO or DNNL while retaining support when their native resources are available.

Given that OpenVINO is quite a large dependency, I think that making if optional makes sense. 

Also - apologies for not picking up this issue earlier. Currently the onnxruntime preset is not functional on ARM.


blueberry (approved):
I tried the latest snapshots (onnxruntime-platform-1.28.0-1.5.15-20260828.110255-3) and it my tests pass (the CUDA problem from the other issue is still there, but it is unrelated to this PR). So, if this change is available in the snapshot, it works.

saudet (approved):

